PR #117, opened 2017-07-26
Rename set to depset

vladmos:


brendandouglas:
We don't yet accept pull requests here -- this is currently just a code drop for educational purposes (see the top level README).

Please feel free to open an internal CL though if you want, otherwise I will -- thanks!

Issue #148, opened 2017-10-19
Kotlin support


dlazerka:
Currently, Bazel seems to be working fine with Kotlin, thanks to rules by https://github.com/pubref/rules_kotlin, but whenever IDEA syncs project with BUILD file, it says "Kotlin not configured", and asks to re-enable Kotlin. Also, "Sources root" forgets that `./src/main/kotlin` is a source root.

brendandouglas:
There's no Kotlin support in this plugin, and we don't have any current plans to add support.

We'd welcome contributions adding it though (the scala support code provides a good example of exactly what would be necessary).

hsyed:
@brendandouglas simply stopping the plugin from wiping the kotlin configuration on a workspace sync would lead to a somewhat useable experience. Would this be doable ?

brendandouglas:
Definitely doable. To do that, a kotlin-specific BlazeSyncPlugin extension would need to be written (similar to the existing ones, such as for scala).

It's not something we have the bandwidth to do ourselves, but we'd certainly welcome contributions.

pcj:
Can you update the README then?  The readme says you're not accepting contributions.  Here at #bazelcon and a number of people are talking about this right now, reticent to even try this repo with such a tepid outlook on the status and support of this repo.

brendandouglas:
Good idea, that README definitely needs updating. To clarify, the plugins themselves (uploaded to the JetBrains plugin repository) are kept up to date, and supported.

This repository, on the other hand, is currently not. It is updated somewhat regularly though, and we'd have no trouble upstreaming any patch based off the current repo HEAD. I'll update the README to reflect this.

hsyed:
I've ported/copied most of the Scala plugin into a kotlin plugin.

@brendandouglas Can't figure out how to get the bazel intelij plugins launcher to work -- trying to use my installed IJ ultimate as the plugin JDK the config doesn't seem to be handling the space in the following flag `-Djb.vmOptionsFile=/Applications/IntelliJ IDEA.app/Contents/bin/idea.vmoptions`.

brendandouglas:
Sounds like the VM options field doesn't accept whitespace within arguments. Have you tried quoting it? Otherwise, is that vmoptions file necessary to test the patch?

hsyed:
@brendandouglas I did try quoting it, no form of quoting worked. I am now facing #165.

brendandouglas:
I'll close this one, as we now have Kotlin support thanks to @hsyed!

PR #157, opened 2017-11-13
wip: kotlin plugin copied from scala plugin.

hsyed:
[WIP PR for issue](https://github.com/bazelbuild/intellij/issues/148). 

Most of the work required to prevent the Kotlin settings from being wiped is done. I haven't figured out how to setup the plugin launcher so haven't been able to test it yet -- using my existing intelij and loading the plugin is too disruptive. 

@brendandouglas could you get take this from here and get it integrated ? I'm happy to continue work on the plugin after I get the launcher working properly.

### DIFF
--- a/kotlin/src/com/google/idea/blaze/kotlin/sync/AlwaysPresentKotlinSyncPlugin.java
+++ b/kotlin/src/com/google/idea/blaze/kotlin/sync/AlwaysPresentKotlinSyncPlugin.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2017 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.idea.blaze.kotlin.sync;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.idea.blaze.base.model.primitives.LanguageClass;
+import com.google.idea.blaze.base.model.primitives.WorkspaceType;
+import com.google.idea.blaze.base.plugin.PluginUtils;
+import com.google.idea.blaze.base.projectview.ProjectViewSet;
+import com.google.idea.blaze.base.scope.BlazeContext;
+import com.google.idea.blaze.base.scope.output.IssueOutput;
+import com.google.idea.blaze.base.sync.BlazeSyncPlugin;
+import com.google.idea.blaze.base.sync.projectview.WorkspaceLanguageSettings;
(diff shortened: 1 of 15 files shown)


bazel-io:

hsyed:
I signed it!

kotlin/src/com/google/idea/blaze/kotlin/sync/AlwaysPresentKotlinSyncPlugin.java
```diff
@@ -0,0 +1,64 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Set;
+
+/**
+ * Unlike most of the kotlin-specific code, will be run even if the JetBrains kotlin plugin isn't
+ * enabled.
+ */
+public class AlwaysPresentKotlinSyncPlugin extends BlazeSyncPlugin.Adapter {
+  private static final String KOTLIN_PLUGIN_ID = "org.jetbrains.kotlin";
```
hsyed: This should be correct.

kotlin/src/com/google/idea/blaze/kotlin/sync/BlazeKotlinSyncPlugin.java
```diff
@@ -0,0 +1,129 @@
...
+            Module workspaceModule,
+            ModifiableRootModel workspaceModifiableModel) {
+        if (!blazeProjectData.workspaceLanguageSettings.isLanguageActive(LanguageClass.KOTLIN)) {
+            return;
+        }
+        for (Library library : ProjectLibraryTable.getInstance(project).getLibraries()) {
+            // Convert the type of the SDK library to prevent the kotlin plugin from
+            // showing the missing SDK notification.
+            if (library.getName() != null && library.getName().startsWith("kotlin-stdlib")) {
+                ExistingLibraryEditor editor = new ExistingLibraryEditor(library, null);
```
hsyed: This should be the correct library type -- there isn't one for the JVM.

kotlin/src/com/google/idea/blaze/kotlin/sync/source/KotlinJavaLikeLanguage.java
```diff
@@ -0,0 +1,28 @@
...
+package com.google.idea.blaze.kotlin.sync.source;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.idea.blaze.java.sync.source.JavaLikeLanguage;
+import java.util.Set;
+
+/** Provides Java-like parts of Kotlin to the Java plugin. */
+public class KotlinJavaLikeLanguage implements JavaLikeLanguage {
+  @Override
+  public Set<String> getFileExtensions() {
```
hsyed: There are other valid extensions but the scope of this work is just the simple kotlin java.

brendandouglas:
This won't work without corresponding changes to the Skylark aspect, to recognize kotlin targets, and extract the required information during sync.

That would also require integration tests, to prevent this code and the scala rules diverging.

How about a smaller patch which only prevents the kotlin settings getting wiped during bazel sync? BlazeDartLibrarySource + BlazeDartSyncPlugin is an example -- it prevents dart libraries being removed during sync, but doesn't do much else.

hsyed:
@brendandouglas I'll have a look at the dart plugin approach in the short term and see how well that ends up working. I want to do this properly though if @pcj is up for it. 

I suppose we'd need to have Kotlin rules available under the bazelbuild org, right ?.  I'll have a closer look at the Scala plugin Skylark logic.

PR #162, opened 2017-11-21
Update rules_scala to the latest commit

vladmos:
The version of `rules_scala` that's used currently contains a reference to the deprecated `set` function, it will soon be not allowed to use such reference even in the parts of code that aren't being called.


brendandouglas:
Thanks! We've recently made this change internally (along with all the other changes needed for the plugin to work with a later version of rules_scala), so it will be present in the next repo sync.

PR #166, opened 2017-11-23
kotlin support modelled on dart plugin.

hsyed:


bazel-io:

hsyed:
@brendandouglas would love to get this simple patch fast tracked so I can start testing it in anger in one of our workspaces.

brendandouglas:
Thank you, it all looks good. I'll apply it upstream. It will take a while to appear here -- we're on a two-week release cycle, and generally sync from upstream once (or less) per release.

We've already cut the current release branch, so this will make it into the next one (scheduled to be uploaded on December 15th).

In the meantime, you can test this on your workspace by making a manual build of the plugin:

```
bazel build --define=ij_product=intellij-latest -- //ijwb:ijwb_bazel
```

hsyed:
@brendandouglas I'm trying the plugin out on some new features in our repo. It's working well enough to get started.

I am having to keep a dummy java target around for new external dependencies that aren't reachable by other targets -- otherwise they don't get indexed. So, I'm looking into adding the aspect support to address this (--i.e., the stuff that the `BlazeScalaWorkspaceImporter` does).

After this the run marker support.

Should I open new PR's for each bit of work ?

hsyed:
@brendandouglas Could you give me some feedback on what is happening with this PR. 

The last commit expanded the scope a bit -- it's an interim solution to let the plugin pick up some libraries. 

I've created an [issue](https://github.com/pubref/rules_kotlin/issues/44) to make the Kotlin rules compatible with the skylark aspect logic. Could you give some feedback on what the simplest approach would be on updating the rules.

brendandouglas:
@hsyed Only the first commit is upstreamed (I missed the notifications that there were further commits added).

As far as updating the kotlin rules so that the aspect recognizes them, the easiest way to do that would be to expose java_common.provider (https://docs.bazel.build/versions/master/skylark/lib/java_common.html).

Then our aspect will automatically recognize that these are 'java-like' targets, and will grab the necessary data (header/class/source jars, etc.).

While this Kotlin support is in flux, perhaps it's best for you to build and test the plugin locally, and once it's all worked out, I can upstream it all at once?

hsyed:
@brendandouglas Ok that makes sense but can we upstream this PR  -- It's quite a useable solution and I have activated Kotlin support in our mono repo, it would be great to not have to send a custom build of the plugin around. 

The next set of changes will happen after the Kotlin rules have been updated for aspect support -- this will probably take quite some time.

hsyed:
@brendandouglas I've added minimal aspect support after changes to upstream kotlin_rules, along with registering a `KotlinJavaLikeLanguage`. Intelij source directory recognition and refactoring is working as far as I can tell. 

I'd like to look into the compiler errors now I'm getting a generic message, instead of the full output from the kotlin compiler in a form suitable for intelij to analyse. 

```Error:(5, 1) Compiling 3 Kotlin source files to cloud/artifacts/butler/lib.jar failed (Exit 1): worker failed: error executing command```

@pcj  @brendandouglas Where do I start with this one, do the rules need to be tweaked, is the protobuf interaction not working correctly or is it something in the plugin that need to be updated ?

hsyed:
Closing this as a huge PR will be incoming soon.

PR #167, opened 2017-11-27
Removing a check for infix expression element from specs2 test configuration creation

### LINKED ISSUE #168: Unable to run specs2 individual test method after running the entire test class
Given a specs2 test class:
```scala
class MyTest extends SpecWithJUnit {
  "my test" should {
    "execute correctly" in {
      ..
    }
  }
}
```
When running `MyTest` first (by right-clicking and selecting `Run 'Bazel test MyTest'), it will run successfully, creating a configuration for `Bazel test MyTest`.
If, afterwards, trying to run "execute correctly" by right clicking, it will run the `MyTest` configuration instead (running the entire test class)

The problem seems to be related to #167, where the check is done on the wrong element (`PsiIdentifier` instead of `ScInfixExpr`).

hmemcpy:
Fixes #155 
Fixes #168  

This pull request removes the check for `ScInfixExpr` from the test configuration for scala2 (see #155). This fixes a bug that prevents right-clicking and running the individual tests.

This is caused due to the element which is right-clicked will usually be the `Literal` identifier (i.e. "the test name"). An additional lookup for the parent `ScInfixExpr` is needed to get the actual test case element. This PR adds this lookup.

In addition, some APIs were tweaked to reduce the dependance on the Scala types.

### DIFF
--- a/scala/src/com/google/idea/blaze/scala/run/Specs2Utils.java
+++ b/scala/src/com/google/idea/blaze/scala/run/Specs2Utils.java
@@ -16,16 +16,15 @@
 package com.google.idea.blaze.scala.run;
 
 import com.google.idea.blaze.base.run.smrunner.SmRunnerUtils;
-import com.intellij.execution.PsiLocation;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.util.PsiTreeUtil;
-import javax.annotation.Nullable;
 import org.jetbrains.plugins.scala.lang.psi.api.expr.ScInfixExpr;
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScTypeDefinition;
 import org.jetbrains.plugins.scala.testingSupport.test.TestConfigurationUtil;
 import org.jetbrains.plugins.scala.testingSupport.test.structureView.TestNodeProvider;
-import scala.Option;
-import scala.Tuple2;
+
+import javax.annotation.Nullable;
+import java.util.function.Predicate;
 
 /** Common functions for handling specs2 test scopes/cases. */
 public final class Specs2Utils {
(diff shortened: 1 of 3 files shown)


bazel-io:

hmemcpy:
Ok, this probably relates to the bug #168, relating to the finding of existing configurations. There's a check in `BlazeScalaJunitTestClassConfigurationProducer` which also checks for `ScInfixExpr`. I'll investigate more and update the bug with more information.

scala/src/com/google/idea/blaze/scala/run/Specs2Utils.java
```diff
@@ -98,4 +96,20 @@ public static String getTestFilter(ScTypeDefinition testClass, PsiElement testCa
...
+      element = PsiTreeUtil.getParentOfType(element, ScInfixExpr.class);
+    }
+    return (ScInfixExpr) element;
+  }
+
+  @Nullable
+  private static String getElementStaticText(PsiElement element) {
+    return TestConfigurationUtil.getStaticTestNameOrDefault(
+            element,
+            null,
```
chaoren: Ugh, they named the parameter "default". I think it's messing up the IDE.

chaoren: https://youtrack.jetbrains.com/issue/IDEA-182888

hmemcpy: Oh wow, so that's why the red squiggly :(
I upvoted the issue, let's hope it will be fixed before flying cars :)

chaoren:
Merged internally, now to export to github.

ittaiz:
Thanks!

PR #177, opened 2017-12-03
Sanitizing test names (escaping parentheses) per specs2 rules

### LINKED ISSUE #176: Test names with parentheses won't run in specs2
Seems that specs2 sanitizes `()` to `[]` due to some junit displaying limitation:
https://github.com/etorreborre/specs2/blob/5d14719c4438ea01cf0c7cc7c4a52c1d8fd6c29b/junit/shared/src/main/scala/org/specs2/reporter/JUnitDescriptions.scala#L106-L111

Test names containing parentheses should be sanitized to match specs2.

hmemcpy:
This fixes #176 by having Bazel correctly match sanitized names to the test filter string.


bazel-io:

hmemcpy:
@chaoren would appreciate you looking at this :)

chaoren:
Merged internally.

hmemcpy:
Many thanks!

PR #180, opened 2017-12-07
HotSwap support for Scala binaries

### LINKED ISSUE #50: Support HotSwap / Reloading Classes
IntelliJ supports "hotswapping" (reloading) classes while debugging a process. https://www.jetbrains.com/help/idea/2016.3/reloading-classes.html

For example, open this sample project https://github.com/dfabulich/bazel-intellij-hotswap

Import it as a Bazel project, open `X.java`, and set a breakpoint on line 14.

```java
System.out.println("hello");
```

Run `//:x` in Debug mode so you hit the breakpoint. Now replace the word "hello" with the word "goodbye"

Expected: Build and hotswap the class
Actual: Nothing happens at first, because you have to Make the project by hand. But, of course, you can't, because of issue #34. "Another command (pid=####) is running.  Waiting for it to complete..."

hmemcpy:
Fixes #50.

### DIFF
--- a/aspect/java_classpath.bzl
+++ b/aspect/java_classpath.bzl
@@ -23,9 +23,12 @@ def _runtime_classpath_impl(target, ctx):
   })
 
 def _get_runtime_jars(target):
-  if not hasattr(target, "java"):
-    return depset()
-  return target.java.compilation_info.runtime_classpath
+  if java_common.provider in target:
+      java_provider = target[java_common.provider]
+      return java_provider.transitive_runtime_jars
+  if hasattr(target, "java"):
+      return target.java.compilation_info.runtime_classpath
+  return depset()
 
 def _aspect_def(impl):
   return aspect(implementation=impl)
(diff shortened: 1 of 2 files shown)


bazel-io:

brendandouglas:
Thanks! All looks good, apart from using java_provider.transitive_runtime_jars by default.

That API is relatively new (so isn't in older versions of bazel), and was quite recently broken (throwing NPEs) -- fixed in https://github.com/bazelbuild/bazel/commit/49008a3c90e65bc4abf5292af823a931b8f4e096.

I noticed because when I applied your patch and tried to debug a java_binary target with this aspect attached, it failed to build.

For now, let's use JavaSkylarkApiProvider preferentially (it's the deprecated API which has worked well for a long time), and fall back to JavaInfo only if that isn't present.

Once JavaInfo stabilizes, we can default to using it.

hmemcpy:
Sounds good, I'll fix the API! Thanks!

brendandouglas:
Thanks -- I've applied the patch upstream. We'll be cutting another release branch next Monday (for release December 22nd).

hmemcpy:
That's great! Just to make sure, will you fix the JavaSkylarkApiProvider on your end? Or should I?

brendandouglas:
I've swapped the order of checks in _get_runtime_jars (so for any target exposing the 'java' provider, we get the runtime classpath from that provider).

Will that work for scala? -- I assume it doesn't expose 'java', so will fall back to using 'java_common.provider'.

hmemcpy:
Yeah, it doesn't expose 'java', but of course, changing the oder would definitely help!

PR #186, opened 2017-12-12
Preventing running individual tests in specs2 inside an abstract class

### LINKED ISSUE #183: Support abstract test classes in specs2
When trying to run an abstract test class (using specs2/junit), it should prompt you for the concrete instance to execute.

hmemcpy:
This disables the run marker for test methods in an abstract test class for Specs2 tests, similar to how it currently works in Java/JUnit test classes.
Closes #183 for now, will send another PR (bigger refactoring) for actually being able to run individual tests in abstract classes.

### DIFF
--- a/scala/src/com/google/idea/blaze/scala/run/producers/BlazeScalaTestRunLineMarkerContributor.java
+++ b/scala/src/com/google/idea/blaze/scala/run/producers/BlazeScalaTestRunLineMarkerContributor.java
@@ -27,6 +27,7 @@ import com.intellij.icons.AllIcons;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.project.Project;
 import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiModifier;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.testIntegration.TestFramework;
 import com.intellij.testIntegration.TestRunLineMarkerProvider;
@@ -125,6 +126,9 @@ public class BlazeScalaTestRunLineMarkerContributor extends ScalaTestRunLineMark
   private static String getSpecs2TestUrl(ScClass testClass, ScInfixExpr testCase) {
     String name = null;
     String protocol = null;
+    if (testClass.hasModifierProperty(PsiModifier.ABSTRACT)) {
+      return null;
+    }
     if (TestNodeProvider.isSpecs2ScopeExpr(testCase)) {
       protocol = SmRunnerUtils.GENERIC_SUITE_PROTOCOL;
       name = Specs2Utils.getSpecs2ScopeName(testCase);


bazel-io:

chaoren:
Same change in BlazeScalaSpecs2TestExprConfigurationProducer?

brendandouglas:
Our java configuration producer intentionally supports producing run configurations when clicking on test methods in abstract test classes. Are you proposing removing that functionality?

chaoren:
Oh wait, so the missing gutter run icon on methods is unintentional?

hmemcpy:
> Oh wait, so the missing gutter run icon on methods is unintentional?
if so, this changes things :)

hmemcpy:
From what I've seen, though, clicking on methods in abstract class indeed offers to create a run configuration, but does not offer to select a subclass, so it will subsequently fail to run in the runner itself.

brendandouglas:
I haven't been able to reproduce that in abstract java test classes -- for me, clicking on a method in an abstract class pops up the 'choose subclass' dialog.

hmemcpy:
Hmm, perhaps I made a mistake... I'll try it again later, let's see if I can repro (it would make sense that it would work for test methods too, since the `BlazeJavaAbstractTestCaseConfigurationProducer` handles those cases.

hmemcpy:
Ok, this is weird, I stopped at the `getTestMethod` handling to figure out what's going on, and I noticed that the psi element is of the wrong type:
image

Just to make sure, I'm clicking on the test method name (i.e. the `testJavaCode` part of `public void testJavaCode`), and it shows up as a PsiIdentifier instead. Am I holding it wrong?

brendandouglas:
That's expected. The next line (TestMethodSelectionUtil.getSelectedMethods) should recognize that you're at a child of a PsiMethod.

Are you clicking on a JUnit4 @Test-annotated method?

hmemcpy:
@brendandouglas just checked again... yeah, my test class looks like this:
```java
public abstract class ExampleJavaTest {
  @Test
  public void testJavaCode() {
    assertTrue(true);
  }
}
```
When clicking on the `testJavaCode` method, I get back `null` from the call to `getSelectedMethods`, because it subsequently does a call to `JUnitUtil.isTestMethod`, and it returns `false` for methods that are either abstract or in an abstract class themselves (see [PsiClassUtil.isRunnableClass](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-api/src/com/intellij/psi/util/PsiClassUtil.java#L41). The value for `mustNotBeAbstract` is true, passed all the way from `JUnitUtil.isTestMethod`)
All I get is the fallback to the subclass selection, which fails to ultimately build the correct filter string:

dec-14-2017 15-01-38

brendandouglas:
You're quiet right, it's not picking up the method, but reverting to running the entire chosen subclass.

hmemcpy:
(yeah, there was some issue with my bazel runner, fixed now)
Anyway, if that's the intended behavior (fallback to running the entire subclass), then individual test run markers don't really make sense, I suppose...
I would go as far as prevent the run configuration creation from right-clicking, only enabling it on the class-level.

PR #191, opened 2017-12-25
don't use deprecated Futures.get method from guava library

chashnikov:
Futures.get method was deprecated in Guava long time ago and it's even removed from modern Guava versions. This change is needed to avoid compatibility problems with IntelliJ IDEA 2018.1 where a new version of Guava is bundled.


bazel-io:

jvoung:
Thanks for the patch and bringing this to our attention! We've merged this internally and it will be on github after the next sync.

PR #206, opened 2018-01-07
Kotlin v2

hsyed:
@brendandouglas here is a more feature complete implementation of the rules.

I will continue work on the rules, so I am hoping the code won't get re-indented or that the files in the Kotlin packages won't get moved around it makes it very confusing to rebase after the code goes through a google round trip via capybara. 

Beyond the code management aspects, when will this repo become the official one ? Not having nightly builds to IJ and only getting work included after months is quite detrimental. I am having to upload jars to slack for my colleagues.


hsyed:
@brendandouglas 

The PR above picks up class and sources jars for the Kotlin distribution which should be primed by the Kotlin rules. This ensures the plugin and the rules are aligned.

The Kotlin plugin has multiple strategies for setting up Kotlin stdlibs one such way is by adding a composite library called KotlinJavaRuntime to the library table -- I am mimicking this. I am also adding some more source jars optional from the `kotlin_additional_stdlibs` section. I should  probably just infer the additional libs from context and attach the source jars as needed -- I will remove this. 

I have some specific problems for which I could use some suggestions.

### Compile error and compile warning Destructuring.

The rules build through a persistent worker now and when I updated the rules I ensured the worker follows the protocol, any output from the workers is ignored and if is an error I just get a generic warning. How do I destructure the warnings and compile errors and direct them to the console. Presumably I need to involve the  Kotlin plugin infrastructure explicitly ?  I found some code for parsing console lines, do I need to hook into these ? Or will there be some way of connecting the worker output to some extension in the kotlin plugin, if this is the case would this be setup by the build runners ?

### Reliable Kotlin source priming.

As mentioned above I am attaching class jars and source jars to the library table and providing them as `BlazeJarLibraries`.  The only source jar that is reliably loaded and synced is the `kotlin-stdlib` this library contains the bulk of the Kotlin stdlib which means it definitely has `PsiClasses` which are referenced.  I cannot get navigate to sources for `kotlin-stdlib-jdk7` and `kotlin-stdlib-jdk8`,  these mainly contain extension inline functions.

The test case I am using is the extension function `use` which is defined in `kotlin-stdlib7` on `AutoCloseable`.

From what I can gather I think the infrastructure responsible for lazily mounting source jars is somewhere in this area I think `AddLibraryTargetDirectoryToProjectViewAttachSourcesProvider.java` and 
`BlazeSourceJarNavigationPolicy.java`. I assume this is not being triggered for when I try to navigate to the Kotlin source jars. One potential aspect is that what I am trying to navigate to is not a `PsiClass` but rather some form of a Kotlin `PsiMethod`.  However, even though it is a PsiMethod it still needs to be in some synthetic PsiClass afaik. 

I have spent a day tinkering around with a debugger and commenting out parts of the code with no joy, I can't find a code path that is attempting to load a psi element belonging to `kotlin-stdlib-jdk8` or `kotlin-stdlib-jdk7`. 

some more points: 

* It didn't make any difference if I  associate the stdlibs to all kotlin targets via a `BlazeSyncAugmenter`.
* Intelij doesn't navigate to stub or decompiled  representations is this the root cause instead ?  
* Is there a way for me  to be able to just attach the sources explicitly ? As  far as I know I was already doing that by creating and attaching the `KotlinJavaRuntime`.

bazel-io:

hsyed:
@brendandouglas when will master be updated with the latest changes ?

hsyed:
creating a new PR.

PR #210, opened 2018-01-12
Updating docs to reference ijwb_bazel_zip

mikelikespie:
The docs currently refer to using the jar as the plugin, however, it appears that we need to use the zip now which contains both the jar and the aspect. Using the current instructions to build/run the plugin do not work, however, using the zip does.

Also, it seems that the output is in `bazel-bin`, not `bazel-genfiles`

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,6 @@ Detailed docs are available [here](http://ij.bazel.build).
 
 ## Building the plugin
 
-Install Bazel, then run `bazel build //ijwb:ijwb_bazel --define=ij_product=intellij-latest`
-from the project root. This will create a plugin jar in
-`bazel-genfiles/ijwb/ijwb_bazel.jar`.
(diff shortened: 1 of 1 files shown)


mikelikespie:
@googlebot We have a corporate CLA (I work for Square, Inc.)

README.md
```diff
@@ -24,6 +24,6 @@ Detailed docs are available [here](http://ij.bazel.build).
 
 ## Building the plugin
 
-Install Bazel, then run `bazel build //ijwb:ijwb_bazel --define=ij_product=intellij-latest`
+Install Bazel, then run `bazel build //ijwb:ijwb_bazel_zip --define=ij_product=intellij-latest`
 from the project root. This will create a plugin jar in
```
brendandouglas: Thanks for the patch! Change the 'jar' reference here to 'zip' as well?

mikelikespie: Good catch! Fixed.

PR #217, opened 2018-01-22
Kotlin v2

hsyed:
This is a much simplified implementation. I've added jdeps support and a `kotlin_stdlib` type to the Kotlin rules.

@brendandouglas any ETA on when the master branch will be synced ?  I am stuck on 2017.2 since the test runners don't work with the current code in master.


bazel-io:

brendandouglas:
Sorry for the delay replying. We haven't had a chance to review this yet, but we'll take a look at what's required to upstream it, and if any changes are needed on your end.

@chaoren could you take a look at this?

chaoren:
Just curious, what's the deal with https://github.com/pubref/rules_kotlin? Looks like we're using your fork of it?

hsyed:
@chaoren I completely reworked the rules for large scale usage and for reducing the amount of hacks I had to put in for developing this plugin. 

* The Pubref rules were macros -- this is the main problem. They created _kt kotlin target and then `java_import` those,  this was getting painful in our mono repo as there were propagation issues so we had to depend on the _kt targets directly .  From the plugins perspective I had to put in a hack in the aspect logic to work around this -- the hack didn't get me far enough.
*  Separate dep attributes for  java deps and kotlin deps, this is a big deal. 
* No proper persistent worker support (5-10 second latency on each bazel command involving kotlin, my toy laptop which I was using over the Christmas period). Now the UX is very similar to java test-compile cycles. It takes a significant amount of time off of a full build of our monorepo as well.
* No `exports`.
* No `resources`,`resource_jars`, `resource_strip_prefix` .
* No `Data`.
* No Jdeps.
* Wouldn't have been possible to add annotation processors easily. At least not in a way that would be compatible with the bazel intellij plugin -- it would have resulted in a ton of corner cases logic in the aspects.

The rules are under consideration for acceptance into bazebuild, to this end I've tried to be as minimalist as possible with the rework, so there is a good foundation for other build facets --e.g., JS, native, Android, kotlin modules etc.

chaoren:
Is pubref not receptive to your changes?

> The rules are under consideration for acceptance into bazebuild

Your fork or the original pubref repo?

Until the rules go under bazelbuild, I would prefer if we depend on a more canonical source (i.e., the pubref one) than your personal fork.

hsyed:
@chaoren I completely agree, I was just mailing @davidstanke about this.  

The plugin functionality will either be severely degraded or completely broken using the Pubref rules, I've removed the aspect hacks that were required for the pubref rules.

hsyed:
@chaoren Sorry missed one of your points. I've committed a few patches into pubref rules.Consider the following:

* It's more of a rewrite, There aren't many lines left that came from the pubref repo. I probably should have started from scratch to avoid confusion. 
* Also it's a time scale thing, I did the majority of the work over Christmas. The rules weren't going to work for large scale use in our repo where we have hundreds of modules that we are planning to gradually split up into packages and port to Kotlin. We needed Kotlin rules that have the feature set of the java rules and perform similarly.  pubref/rules_kotlin is mainly maintained by one person so getting all these changes in would have taken a very long time. 
* Disruptive to the pubref/rules_kotlin users, the switch from macros to rules and all of the things I have added is too disruptive to realistically integrate. I have also focused on simplicity and have an aim on using idiomatic bazel concepts (rules, providers, toolchains) for all the features. There aren't any real test cases in the pubref rules, so I would not have gotten very far trying to maintain backward compatibility.

So, It's not that they are not receptive. pubref/rules_kotlin is maintained by one person, and he isn't responsive enough and the rework that I embarked on is likely out of scope. I am just as happy to have the work upstreamed into pubref/rules_kotlin.

plumpy:
FWIW, I also looked at cleaning up the pubref rules and came to basically all the same conclusions: the changes are realistically so big it becomes a complete rewrite, and I didn't think he'd accept the patches since it would break all the builds. (I never bothered to ask, though.)

Could you just publish them on their own instead of as a fork? I've been using my own custom rules for my personal projects and I'd happily switch to yours. (And submit patches where appropriate.)

holmes:
@hsyed what's the best way to get this plugin to work with an android app?

I've got almost all my packages building, but it's failing when I try to put a `kotlin_library` in an `android_binary` with the error:

```
/work/catalog holmes/new-kotlin-rules * ./bazelw mobile-install --start_app //src/test/java/com/squareup/thyme:thyme
ERROR: /work/catalog/src/test/java/com/squareup/thyme/BUILD:53:15: in exports attribute of android_library rule //src/test/java/com/squareup/thyme:java: kotlin_library rule '//src/test/java/com/squareup/thyme:kt' is misplaced here (expected aar_import, android_library, cc_library, java_import, java_library or java_lite_proto_library)
ERROR: Analysis of target '//src/test/java/com/squareup/thyme:thyme' failed; build aborted: Analysis of target '//src/test/java/com/squareup/thyme:java' failed; build aborted
INFO: Elapsed time: 4.541s
```

Kotlin definition:
```
kotlin_library(
    name = "kt",
    srcs = glob(["MainActivityPresenter.kt"]),
    deps = [
        ":res",
        "@com_android_support_appcompat_v7_27_0_2//aar",
        "@com_android_support_animated_vector_drawable_27_0_2//aar",
        "@com_android_support_support_annotations_27_0_2//jar",
        "@com_android_support_support_vector_drawable_27_0_2//aar",
        "@com_android_support_support_core_utils_27_0_2//aar",
        "@com_android_support_support_core_ui_27_0_2//aar",
        "@com_android_support_support_fragment_27_0_2//aar",
        "@com_android_support_support_compat_27_0_2//aar",
        "@com_android_support_recyclerview_v7_27_0_2//aar",
        "//src/test/java/com/squareup/thyme/utilities:utilities",
        "//src/test/java/com/squareup/thyme/calendar:calendar",
        "@android_arch_lifecycle_common_1_0_3//jar",
        "//third_party:rxbinding",
        "//third_party:rxjava",
        "//src/main/java/com/squareup/shared/serum/event",
        "//src/main/java/com/squareup/shared/serum/store",
    ],
)
```

Android binary definition:
```
android_binary(
    name = "thyme",
    srcs = glob(["*.java"]),
    custom_package = PACKAGE,
    manifest = MANIFEST,
    manifest_values = {
        "minSdkVersion": "21",
        "targetSdkVersion": "27",
    },
    multidex = "native",
    deps = [
        ":java",
        ":kt",
        ":res",
        "//src/main/java/com/squareup/shared/appointments/calendar/layout",
        "//src/main/java/com/squareup/shared/appointments/calendar/model",
        "//src/main/java/com/squareup/shared/appointments/calendar/stores",
        "//src/main/java/com/squareup/shared/appointments/sync/model",
        "//src/main/java/com/squareup/shared/serum/event",
        "//src/main/java/com/squareup/shared/serum/store",
        "//third_party:dagger",
        "//third_party:rxandroid",
        "//third_party:rxbinding",
        "//third_party:rxjava",
        "@android_arch_core_runtime_1_0_0//aar",
        "@android_arch_lifecycle_common_1_0_3//jar",
        "@android_arch_lifecycle_runtime_1_0_3//aar",
        "@com_android_support_animated_vector_drawable_27_0_2//aar",
        "@com_android_support_appcompat_v7_27_0_2//aar",
        "@com_android_support_recyclerview_v7_27_0_2//aar",
        "@com_android_support_support_annotations_27_0_2//jar",
        "@com_android_support_support_compat_27_0_2//aar",
        "@com_android_support_support_core_ui_27_0_2//aar",
        "@com_android_support_support_core_utils_27_0_2//aar",
        "@com_android_support_support_fragment_27_0_2//aar",
        "@com_android_support_support_vector_drawable_27_0_2//aar",
    ],
)
```

chaoren:
@holmes You might want to wrap the output jar of the kotlin rule with `java_import`.

hsyed:
@holmes logically that should work, the rules are creating [java providers](https://github.com/hsyed/rules_kotlin/blob/f0aeb20a8c7b291984657a620c8865cad5ff4879/kotlin/rules/compile.bzl#L119). 

The android binary probably needs to be updated to use Java providers, or maybe a skylark created provider doesn't provide enough metadata (byte code level ?). 

Could you create an issue and I'll put in a test case.

holmes:
@hsyed Thanks, created issue at https://github.com/bazelbuild/intellij/issues/220

hsyed:
@holmes the problem is with the rules they are currently found here: https://github.com/hsyed/rules_kotlin. 

Just to double check you are using those rules right and not pubref/rules_kotlin ?

holmes:
@hsyed yep, I switched over to your rules as well. I'm seeing the same failure on the command line so I guess the issue above has more to do with the rules right now than with the intellij plugin.

hsyed:
@brendandouglas @chaoren any updates on when the master in this repo will get synced ? desperately need this to build a new version of the plugin for 2017.3 for our internal use.

brendandouglas:
I'm planning on syncing from upstream today. It won't include this patch, but will include 2017.3 compatibility

hsyed:
@brendandouglas thats cool, I'll rebase and push any needed fixes.

hsyed:
@brendandouglas @chaoren 

* I fixed a bug in the sync plugin -- `kotlin jars` for external dependencies are now attaching correctly.
* updated the reference to the new rules.

I added a experimental [`kotlin_import`](https://github.com/bazelbuild/rules_kotlin/blob/b846341e3e9c9ad71e5d50ecd839562608913c05/kotlin/kotlin.bzl#L293-L326) rule  that makes it somewhat easy to relay the class, src pairs to the plugin (via the `kt` provider attribute). 

I spent a lot of time trying to find a way for the machinery in the `libraries` package in the java plugin to carry out the attach and source navigation. I could never get this working apart for the sources for the main kotlin stdlib -- and this is done by the KotlinPluginConfigurator.

If I remember correctly The JarNavigationPolicy does not trigger on any Kotlin PsiElements.

base/src/com/google/idea/blaze/base/model/primitives/Kind.java
```diff
@@ -89,9 +89,11 @@
...
   DART_PROTO_LIBRARY("dart_proto_library", LanguageClass.DART, RuleType.LIBRARY),
   DART_LIBRARY("_dart_library", LanguageClass.DART, RuleType.LIBRARY),
-  KOTLIN_LIBRARY("kotlin_library", LanguageClass.KOTLIN, RuleType.LIBRARY),
-  KOTLIN_BINARY("kotlin_binary", LanguageClass.KOTLIN, RuleType.BINARY),
-  KOTLIN_TEST("kotlin_test", LanguageClass.KOTLIN, RuleType.TEST),
+  KT_JVM_LIBRARY("kt_jvm_library", LanguageClass.KOTLIN, RuleType.LIBRARY),
+  KT_JVM_BINARY("kt_jvm_binary", LanguageClass.KOTLIN, RuleType.BINARY),
+  KT_JVM_TEST("kt_jvm_test", LanguageClass.KOTLIN, RuleType.TEST),
+  KT_JVM_IMPORT("kt_jvm_import", LanguageClass.KOTLIN, RuleType.UNKNOWN),
+  KOTLIN_STDLIB("kotlin_stdlib", LanguageClass.KOTLIN, RuleType.UNKNOWN),
```
chaoren: Why not `RuleType.LIBRARY`?

java/src/com/google/idea/blaze/java/run/ExternalFilePositionManager.java
```diff
@@ -56,11 +56,17 @@ protected PsiFile getPsiFileByLocation(Project project, Location location) {
     return refType != null ? qualifiedNameToPsiFile.get(refType.name()) : null;
   }
 
+
   @Override
   public List<ClassPrepareRequest> createPrepareRequests(
       ClassPrepareRequestor requestor, SourcePosition position) throws NoDataException {
+    List<ClassPrepareRequest> prepareRequests = super.createPrepareRequests(requestor, position);
+    // PrepareRequests returns an empty list for Kotlin sources. Returning this blocks the KotlinManager from contributing.
```
chaoren: > Returning this blocks the KotlinManager from contributing.

The doc says:
>   * @return the prepare request, or null if the code is managed by this {@code PositionManager} but no class prepare notification is needed
>   * @throws NoDataException if the position is not in the code managed by this {@code PositionManager}

How does the built-in position manager avoid this conflict? Is it based on extension point ordering?

base/src/com/google/idea/blaze/base/model/primitives/Kind.java
chaoren: Oh, because you probably can't edit the standard library. Nevermind.

kotlin/src/com/google/idea/blaze/kotlin/run/NonBlazeProducerSuppressor.java
```diff
@@ -0,0 +1,48 @@
...
+package com.google.idea.blaze.kotlin.run;
+
+import com.intellij.openapi.components.AbstractProjectComponent;
+import com.intellij.openapi.project.Project;
+
+
+/**
+ * Suppresses certain non-Blaze configuration producers in Blaze projects.
+ */
+public class NonBlazeProducerSuppressor extends AbstractProjectComponent {
```
chaoren: What's going on here?

kotlin/src/com/google/idea/blaze/kotlin/run/producers/KotlinTestConfigBuilder.java
```diff
@@ -0,0 +1,192 @@
...
+
+import javax.annotation.Nullable;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Private utility class for
```
chaoren: for what?

third_party/kotlin/get_kotlin_ws_entries.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
chaoren: See above. We don't need this.

WORKSPACE
```diff
@@ -162,6 +162,53 @@ new_http_archive(
...
+      'kotlin/lib/allopen-ide-plugin.jar',
+      'kotlin/lib/noarg-compiler-plugin.jar',
+      'kotlin/lib/noarg-ide-plugin.jar',
+  ],
+  visibility = ['//visibility:public'],
+)
+"""
+
+# Kotlin plugin for IntelliJ CE 2017.2. Required at compile-time for kotlin-specific features.
+new_http_archive(
```
chaoren: The intellij-community itself contains the kotlin plugin under plugins/Kotlin/lib/*.jar, so we don't need these.

kotlin/src/com/google/idea/blaze/kotlin/sync/BlazeKotlinSyncPlugin.java
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * Update the compiler settings of the project if needed. The language setting applies to both the api version and the language version. Blanket setting
+     * this project wide is fine. The rules should catch incorrect usage.
+     */
+    private static void maybeUpdateCompilerConfig(Project project, ProjectViewSet projectViewSet) {
+        LanguageVersion languageLevel = BlazeKotlinSections.getLanguageLevel(projectViewSet);
+        String languageLevelVersionString = languageLevel.getVersionString();
+        CommonCompilerArguments settings = (CommonCompilerArguments) KotlinCommonCompilerArgumentsHolder.Companion.getInstance(project).getSettings().unfrozen();
+        boolean updated = false;
+        if (settings.getApiVersion() == null || !settings.getApiVersion().equals(languageLevelVersionString)) {
```
chaoren: This seems to be `settings.apiVersion` in 2017.2, but `settings.getApiVersion()` in 2017.3, we'll need an sdkcompat for this.

WORKSPACE
hsyed: @chaoren Ok I got it working with IJ. Currently I have created a `intellij_platform_sdk:kotlin` / `kotlin_for_test`. 

 I assume the Kotlin plugin is available in all variants (Clion, Android Studio and IJ), I'll add the jars to `intellij_platform_sdk:sdk` ?.

chaoren: It's not in CLion.

hsyed: I'm not going to make any assumptions and only include Kotlin in IJ as it extends to the sdk compat as well.

base/tests/utils/unit/com/google/idea/blaze/base/scope/ErrorCollector.java
```diff
@@ -35,6 +36,14 @@ public void assertNoIssues() {
     assertThat(issues).isEmpty();
   }
 
+  public List<IssueOutput> issuesOf(IssueOutput.Category category) {
+    return issues.stream().filter(i -> i.getCategory().equals(category)).collect(ImmutableList.toImmutableList());
```
chaoren: `ImmutableList.toImmutableList()` doesn't exist until guava 21, so it won't be available in Android Studio. Use `GuavaHelper.toImmutableList()`.

chaoren: `Collectors.toList()` to avoid the extra dependency.

java/src/com/google/idea/blaze/java/run/ExternalFilePositionManager.java
hsyed: Can't say much on this one without fully getting into the underlying code. 

I spend a lot of time to track down this problem, and I have been debugging Kotlin and Java code in our mono repo since without any issues - I even break pointed on Scala jars. 

We could be more explicit in this method (and / or the other ones) by testing the language type: --i.e., `position.getElementAt().getLanguage().is(KotlinLanguage.INSTANCE)`. Java would have a link dependency on the Kotlin but not a problem now since it's included in the platform.

hsyed:
@chaoren thanks for the review, I have made all of the changes. 

* The sdk compat the compiler settings for 2017.2 is not fully tested, the api is completely different and I am not using 2017.2 anymore -- there is a test case for it though and it passes.
* I only added the Kotlin plugin to the intellij config. 

Is the debugger going to be an issue ?

kotlin/BUILD
```diff
@@ -1,19 +1,106 @@
...
+    name = "kotlin_integration_test_plugin",
+    testonly = 1,
+    optional_plugin_xmls = OPTIONAL_PLUGIN_XMLS,
+    plugin_xml = ":kotlin_plugin_xml",
+    deps = [":kotlin"],
+)
+
+intellij_unit_test_suite(
+    name = "unit_tests",
+#    test_class = "com.google.idea.blaze.kotlin.sync.KotlinSdkUtilsTest",
```
chaoren: Why is this commented out?

kotlin/src/META-INF/kotlin-contents.xml
```diff
@@ -0,0 +1,15 @@
+<idea-plugin>
+    <extensions defaultExtensionNs="com.google.idea.blaze">
+        <SyncPlugin implementation="com.google.idea.blaze.kotlin.sync.BlazeKotlinSyncPlugin"/>
+        <JavaLikeLanguage implementation="com.google.idea.blaze.kotlin.sync.source.BlazeKotlinJavaLikeLanguage"/>
+    </extensions>
+
+    <extensions defaultExtensionNs="com.intellij">
+        <runConfigurationProducer
+                implementation="com.google.idea.blaze.kotlin.run.producers.BlazeKotlinTestClassConfigurationProducer"
+                order="first"/>
```
chaoren: Is there a specific producer that this needs to override? The scala one is also wrong. They can't all be first. You can use `order="before OtherProducer"`

chaoren: Also specify language here?

chaoren: The scala one was for RunLineMarker.

hsyed: Makes sense, I just copied Scala to start with. 

I think it should go before the Java Producer, that should do the trick ? 

The original Kotlin run producers are already suppressed by the java plugin I think.

chaoren: Oh, then it won't be necessary. You can just get rid of order.

hsyed: There isn't a specific producer to override at the moment, this is wrong.  I wrote this producer before I had any clue what I was doing , so it's just about doing the right thing -- in a cause and effect sort of a way. 

Once this changeset goes in and after a bit of a break I will revisit testing. We are already on junit Jupiter (which we are running via the bazel runner) and we want to use a BDD framework for acceptance and some integration testing. 

When I revisit it I hope to address the mapping of test events to tree like test events as represented in a BDD framework. Kotlin Spek and KotlinTest are both run via Jupiter now. So revisiting  will require the Bazel test runner to do the right thing for Jupiter.

@brendandouglas any comments on what would be required to integrate with Jupiter properly -- and to support tree based test models ?

hsyed: @chaoren The Scala and Kotlin producers don't really collide as they are both performing language specific psi introspections and will naturally bail out. I don't think we need to make the ordering more precise for now.

hsyed: removed the order entries and validated.

kotlin/src/com/google/idea/blaze/kotlin/sync/AlwaysPresentKotlinSyncPlugin.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.idea.blaze.base.model.primitives.LanguageClass;
+import com.google.idea.blaze.base.plugin.PluginUtils;
+import com.google.idea.blaze.base.projectview.ProjectViewSet;
+import com.google.idea.blaze.base.scope.BlazeContext;
+import com.google.idea.blaze.base.scope.output.IssueOutput;
+import com.google.idea.blaze.base.sync.projectview.WorkspaceLanguageSettings;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.Nullable;
+
+import static com.google.idea.blaze.kotlin.BlazeKotlin.PLUGIN_ID;
```
chaoren: This `PLUGIN_ID` comes from the Kotlin plugin itself. We can't have this in the always present sync plugin. Since the purpose of this is to handle the case where the kotlin plugin is absent.
I would recommend getting rid of `BlazeKotlinBaseSyncPlugin` entirely, since there are many other things which implicitly require the Kotlin plugin to be present.

chaoren:
So I've been playing around with a sample kotlin project, and found some issues. They're probably combinations between this change and rules_kotlin.

1. Kotlin not configured. Kotlin needs a KotlinJavaRuntime library to be added (and it does so once you click the notification), but our sync removes it each time. It should be done as part of the sync. Scala does this too with the scala-library jar. This requires the KotlinJavaRuntime be provided through the aspect as a dep or a special attribute, recognized by the sync plugin, and passed over to the kotlin plugin to be registered.
2. The kotlin-test library is not added to kt_jvm_test targets by default? Likewise with kotlin-reflect to, I guess, kt targets in general. Not sure about the kotlin-reflect, but I think kotlin-test definitely needs to be added by default. This is rules_kotlin only.
3. The gutter run marker on binaries appears broken, and shows "Nothing here". A Kotlin JUnit test doesn't show run markers at all.

hsyed:
I suspect your sync configuration is missing some things:

1. The root bazelproject file should look like this:
```
java_language_level: 8

kotlin_language_version: 1.2

additional_languages:
  scala
  kotlin
```

2. A kotlin target should be discoverable as part of the sync, the stdlibs enter the sync model via an active kotlin rule.

3. The rules from the bazebuild/rules_kotlin should be used (I suppose you have this one covered). 

---

1. A library named` KotlinJavaRuntime` isn't strictly necessary, The Kotlin plugin activates if it sees appropriately named libraries in the library table -- it activates just by detecting Kotlin-stdlib.  The blaze plugin is designed to load the libraries from the compiler repo in the workspace this is so that the libraries from the workspace are used. This does require the bazel build rules. The rules also set the kotlin stdlib as deps for each target.  Eventually the stdlib setup will be provided via a toolchain.

The last three functions in the sync plugin set up the library table.  Earlier versions of the plugin populated a `KotlinJavaRuntime` this design is simpler.  here is a screenshot from our mono repo,  you should see something similar: 
screen shot 2018-02-07 at 22 30 47

2. Kotlin test isn't needed by default, it's just a few utilities. If the user want's it he can add it. It's available at "com_github_jetbrains_kotlin//:kotlin-test" (I have it declared in our mono repo as //third-party")

3. The run markers require the a proper sync to have occurred to line up. I think it's to do with the package name detected. Until the sync system knows about the packages the Run Providers will not be able to create the run configuration.

4. `kt_jvm_import` is a stop gap rule. I have an issue open in bazel core to make the ijar process leave inlining metadata alone.  Once this is done we can use `java_import` to load kotlin artifacts. `kt_jvm_library` needs to have some srcs, it can't be used to `declare` libraries (I can add this if we really need it).

chaoren:
Ah, yes. We have a separate plugin target internally, and it was missing the optional xml, so the entire sync plugin was missing. It's working fine now.

Why not just always add this flag? 
```
INFO: From Compiling 1 Kotlin source files to experimental/users/chaorenl/kotlin/minimal.jar:
HINT: KotlinCompile will compile faster if you run: echo "build --strategy=KotlinCompile=worker" >>~/.bazelrc
```

hsyed:
@chaoren Could you find out what we should be doing by asking internally ?  (we need a slack group)

1. The worker documentation suggests that a persistent worker should be opt-in (for Kotlin opting in is a no-brainer imo).
2.  If  there is a way for me to hard code this flag how do I do It in the rules -- [some setting here ?](https://github.com/bazelbuild/rules_kotlin/blob/7585b9e70044129c788768f76a97b2d429d80880/kotlin/rules/compile.bzl#L87).
3. I could disable the warning, the hint generating code was part of the worker wrapper I copied from bazel core. 
4. We could pass it from the plugin -- but that would mean a potential mismatch when used from the command line.

chaoren:
> Could you find out what we should be doing by asking internally ? (we need a slack group)

Don't worry about it. I've handled it.

The worker part is unrelated to this plugin, but it seems like the warning is the way intended by the bazel team.

kotlin/src/com/google/idea/blaze/kotlin/sync/importer/BlazeKotlinWorkspaceImporter.java
```diff
@@ -0,0 +1,98 @@
...
+            HashMap<LibraryKey, BlazeJarLibrary> libraries,
+            HashMap<TargetIdeInfo, ImmutableList<BlazeJarLibrary>> targetLibraryMap
+    ) {
+        List<BlazeJarLibrary> scratchLibSet = new ArrayList<>();
+
+        targetMap
+                .targets().stream()
+                .filter(target -> target.kind.languageClass.equals(LanguageClass.KOTLIN))
+                .filter(importFilter::isSourceTarget)
+                .forEach(ideInfo -> withTransitiveTargets(ideInfo).forEach(depIdeInfo -> {
```
chaoren: Instead of `.forEach(... .forEach` you can `.flatMap(this::withTransitiveTargets).forEach(`

kotlin/src/com/google/idea/blaze/kotlin/sync/importer/BlazeKotlinWorkspaceImporter.java
```diff
@@ -0,0 +1,98 @@
...
+                    scratchLibSet.clear();
+                    depIdeInfo.javaIdeInfo.jars.stream().map(BlazeJarLibrary::new).forEach(depJar -> {
+                        libraries.putIfAbsent(depJar.key, depJar);
+                        scratchLibSet.add(depJar);
+                    });
+                    if (depIdeInfo.kind.languageClass == LanguageClass.KOTLIN) {
```
chaoren: Just do a `.filter(depIdeInfo.kind.languageClass == LanguageClass.KOTLIN)` before `.forEach`?

kotlin/src/com/google/idea/blaze/kotlin/sync/importer/BlazeKotlinWorkspaceImporter.java
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                }));
+    }
+
+    @Nonnull
+    private Stream<TargetIdeInfo> withTransitiveTargets(TargetIdeInfo target) {
+        return Stream.concat(
+                Stream.of(target),
+                TransitiveDependencyMap.getTransitiveDependencies(target.key, targetMap).stream()
+                        .map(depTarget -> {
```
chaoren:
```
            .map(targetMap::get)
            .filter(Objects::nonNull)
            .filter(info -> info.javaIdeInfo != null)
            .filter(Objects::nonNull));
```

kotlin/tests/integrationtests/com/google/idea/blaze/kotlin/KotlinSyncTest.java
```diff
@@ -0,0 +1,214 @@
...
+        assertThat(BlazeKotlinSections.getLanguageLevel(projectViewSet)).isEqualTo(languageVersion);
+
+        // test the compiler reflect the project view.
+        BlazeKotlinCompilerArgumentsUpdaterCompat updater = BlazeKotlinCompilerArgumentsUpdaterCompat.build(getProject());
+        assertThat(updater.getApiVersion()).isEqualTo(languageVersion.getVersionString());
+        assertThat(updater.getLanguageVersion()).isEqualTo(languageVersion.getVersionString());
+    }
+
+
+    private void assertNoSyncErrors() {
```
chaoren: `errorCollector.assertIssues(BlazeKotlin.Issues.RULES_ABSENT_FROM_WORKSPACE)` seems to be identical to this?

kotlin/src/com/google/idea/blaze/kotlin/sync/importer/BlazeKotlinWorkspaceImporter.java
hsyed: This would change the semantics. The process does two things:

1. Collect all external transitive libraries reachable from a Kotlin target.
2. Create a lookup cache to be used in setting up the library table for Kotlin.

if we pull the filter up we won't be doing (1) --i.e.,  plain java libraries won't be reachable from Kotlin targets. Is there something in the Java plugin that will take care of this for JavaLikeLanguage ?

kotlin/src/com/google/idea/blaze/kotlin/sync/importer/BlazeKotlinWorkspaceImporter.java
hsyed: last filter shouldn't be necessary I think. 
```
                 .map(targetMap::get)
                        .filter(Objects::nonNull)
                        .filter(info -> info.javaIdeInfo != null)
        );
```

kotlin/tests/integrationtests/com/google/idea/blaze/kotlin/KotlinSyncTest.java
hsyed: The rules being absent from the workspace is a warning `IssueOutput.warn(BlazeKotlin.Issues.RULES_ABSENT_FROM_WORKSPACE).submit(context);`. so both assertions are needed here.

kotlin/src/com/google/idea/blaze/kotlin/sync/importer/BlazeKotlinWorkspaceImporter.java
hsyed: Never mind my reasoning is wrong and double injects the java libraries (the Scala sync plugin has a comment on this being ok).. So prefiltering is correct.

 I've updated the method body to:

```
    private void collectTransitiveLibsFromKotlinSourceTargets(
            HashMap<LibraryKey, BlazeJarLibrary> libraries,
            HashMap<TargetIdeInfo, ImmutableList<BlazeJarLibrary>> targetLibraryMap
    ) {
        Stream<TargetIdeInfo> transitiveClosureOfKotlinTargets = targetMap
                .targets().stream()
                .filter(target -> target.kind.languageClass.equals(LanguageClass.KOTLIN) && importFilter.isSourceTarget(target))
                .flatMap(this::withTransitiveTargets)
                .filter(depIdeInfo -> depIdeInfo.kind.languageClass == LanguageClass.KOTLIN);

        List<BlazeJarLibrary> scratchLibSet = new ArrayList<>();

        transitiveClosureOfKotlinTargets.forEach(depIdeInfo -> {
            scratchLibSet.clear();
            // noinspection ConstantConditions javaIdeInfo has been asserted in ::withTransitiveTargets
            depIdeInfo.javaIdeInfo.jars.stream().map(BlazeJarLibrary::new)
                    .forEach(depJar -> {
                libraries.putIfAbsent(depJar.key, depJar);
                scratchLibSet.add(depJar);
            });
            targetLibraryMap.put(depIdeInfo, ImmutableList.copyOf(scratchLibSet));
        });
    }
```

chaoren: I'm not a fan of the `scratchLibSet`. They way it's cleared and built seems prone to bugs. Does this work?
```
    targetMap
        .targets()
        .stream()
        .filter(target -> target.kind.languageClass.equals(LanguageClass.KOTLIN))
        .filter(importFilter::isSourceTarget)
        .map(target -> TransitiveDependencyMap.getTransitiveDependencies(target.key, targetMap))
        .flatMap(Collection::stream)
        .map(targetMap::get)
        .filter(Objects::nonNull)
        .filter(info -> info.kind.languageClass == LanguageClass.KOTLIN)
        .filter(info -> info.javaIdeInfo != null)
        .forEach(
            info -> {
              List<BlazeJarLibrary> targetLibraries =
                  info.javaIdeInfo
                      .jars
                      .stream()
                      .map(BlazeJarLibrary::new)
                      .peek(jar -> libraries.putIfAbsent(jar.key, jar))
                      .collect(Collectors.toList());
              targetLibraryMap.put(info, ImmutableList.copyOf(targetLibraries));
```

hsyed: Interesting, I've never used peek ! The huge process chain will be confusing to interpret later on, but I have removed the tally.

kotlin/tests/integrationtests/com/google/idea/blaze/kotlin/KotlinSyncTest.java
chaoren: ErrorCollector doesn't make a distinction. `errorCollector.assertIssues(BlazeKotlin.Issues.RULES_ABSENT_FROM_WORKSPACE)` works just fine.

hsyed: aah I see what you mean, the underlying assertion is based on `containsExactly` `assertThat(messages).containsExactly((Object[]) requiredMessages);` I'll make the change. The method names are a misleading.

hsyed:
@chaoren just added annotation processing to the rules, the repo contains the dagger example which you can try out against the plugin.

hsyed:
@chaoren just added initial toolchain for the jvm aspect of the rules, it currently holds flags relevant to the kotlin plugin settings. 

Once this pr goes in I can work add support for this.

PR #248, opened 2018-03-03
navigate directly to loaded symbol for load aliases.

hsyed:
Hopefully this is the correct way to do it.

### DIFF
--- a/base/src/com/google/idea/blaze/base/lang/buildfile/psi/TargetExpression.java
+++ b/base/src/com/google/idea/blaze/base/lang/buildfile/psi/TargetExpression.java
@@ -15,10 +15,14 @@
  */
 package com.google.idea.blaze.base.lang.buildfile.psi;
 
+import com.google.idea.blaze.base.lang.buildfile.psi.util.PsiUtils;
 import com.google.idea.blaze.base.lang.buildfile.references.TargetReference;
 import com.intellij.lang.ASTNode;
+import com.intellij.psi.PsiElement;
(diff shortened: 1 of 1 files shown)


hsyed:
Ok so I am now working on struct member completion that works for local and imported symbols.

I can see when working on member completion that what I put in for resolving navigation elements for anonymous symbols in the last commit is not consistent with the general strategy for symbol resolution. I'll know more as I continue the work. Any comments welcome.

@brendandouglas I can't use rules-kotlin as my test workspace  as master doesn't have Kotlin changes. Also if I spend a lot of time on this and other isolated feature patches it would mean I won't get to benefit from any of them for months without managing my own releases. Releasing just Kotlin is fine but multiple patch streams would become crazy.

base/src/com/google/idea/blaze/base/lang/buildfile/psi/TargetExpression.java
```diff
@@ -34,6 +38,28 @@ protected void acceptVisitor(BuildElementVisitor visitor) {
...
+    PsiElement referencedElement = getReferencedElement();
+    if(referencedElement != null) {
+      return referencedElement.getNavigationElement();
+    }
+    return super.getNavigationElement();
+  }
+
+  @Nullable
+  @Override
+  public PsiElement getReferencedElement() {
```
hsyed: this is likely wrong -- it should break rename refactoring and reference references I think -- will fix after feedback on #249

PR #249, opened 2018-03-05
completion support for arbitrary structs

hsyed:
Complete struct paths. The work is very useful for modularising growing rulesets. 

Path construction will complete a struct embedding structs from other files. 

The static helpers in this class could be isolated and then used to expand navigation support implemented in  #248 -- that PR only works for the root of a path. 

Updating the Psi tree to support reference resolution correctly is probably the best long term solution but that would be quite a sprawl of work whereas this approach is nice and contained.

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -282,6 +282,7 @@
     <completion.contributor language="BUILD" implementationClass="com.google.idea.blaze.base.lang.buildfile.completion.BuiltInSymbolCompletionContributor"/>
     <completion.contributor language="BUILD" implementationClass="com.google.idea.blaze.base.lang.buildfile.completion.BuiltInFunctionAttributeCompletionContributor"/>
     <completion.contributor language="BUILD" implementationClass="com.google.idea.blaze.base.lang.buildfile.completion.ArgumentCompletionContributor"/>
+    <completion.contributor language="BUILD" implementationClass="com.google.idea.blaze.base.lang.buildfile.completion.StructAttributeCompletionContributor"/>
     <langCodeStyleSettingsProvider implementation="com.google.idea.blaze.base.lang.buildfile.formatting.BuildLanguageCodeStyleSettingsProvider"/>
     <codeStyleSettingsProvider implementation="com.google.idea.blaze.base.lang.buildfile.formatting.BuildCodeStyleSettingsProvider"/>
     <editor.backspaceModeOverride language="BUILD" implementationClass="com.intellij.codeInsight.editorActions.SmartBackspaceDisabler"/>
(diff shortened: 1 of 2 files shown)


brendandouglas:
Could you give an example of the problem this PR is solving? In what context is autocomplete not currently working for BUILD/Skylark files?

hsyed:
the PR above lets you code complete through a struct path --i.e., `utils.actions.build_resourcejar`. This [modularisation](https://github.com/bazelbuild/bazel-skylib#writing-a-new-module) pattern is becoming quite common for skylark. 

Here are two snippets from the Kotlin rules. They are in separate files.

file `//kotlin/internal/utils.bzl`
```
utils = struct(
    actions = struct(
        build_resourcejar = _build_resourcejar_action,
        fold_jars = _fold_jars_action,
        maybe_make_srcsjar = _maybe_make_srcsjar_action,
        write_launcher = _write_launcher_action,
    ),
    collect_all_jars = _collect_all_jars,
    collect_jars_for_compile = _collect_jars_for_compile,
    restore_label = _restore_label,
)
```
file `//kotlin/internal/kt.bzl`
```
_defs = struct(
    DEFAULT_TOOLCHAIN = "@io_bazel_rules_kotlin//kotlin:default_toolchain",
    # The name of the Kotlin compiler workspace.
    KT_COMPILER_REPO = "com_github_jetbrains_kotlin",
    # The name of the rules repo. Centralised so it's easy to change.
    REPO_ROOT = "io_bazel_rules_kotlin",
    TOOLCHAIN_TYPE = "@io_bazel_rules_kotlin//kotlin:kt_toolchain_type",
)

_KtInfo = provider(
    fields = {
        "src": "the source files. [intelij-aspect]",
        "outputs": "output jars produced by this rule. [intelij-aspect]",
    },
)

_KtPluginInfo = provider(
    fields = {
        "processors": "a serializeable list of structs containing an annotation processor definitions",
    },
)

kt = struct(
    defs = _defs,
    info = struct(
        KtInfo = _KtInfo,
        KtPluginInfo = _KtPluginInfo,
    ),
)
```

hsyed:
it will be a while before I have time to work on this again -- I could write some tests then. 

The scope of this task might be a bit bigger.  I don't think rename refactoring currently works for struct paths and symbol navigation only works for the first element from task 248. If we want to address those then we might want to move out the common functions somewhere.

brendandouglas:
Thanks, that makes sense. I'd prefer to implement this via reference resolution, which will give us more reliable autocomplete behavior (CompletionContributor can be quite fiddly and difficult to debug), as well as navigation, refactoring support, etc.

I'm happy to take a look at this, though might not get to it for a month or two.

hsyed:
@brendandouglas I'll do some more work on this, I'll close #248 and implement that here. I'll isolate the functions in this PR into `PsiReferenceUtils` and then begin updating the Psi tree to reference correctly.

PR #262, opened 2018-03-16
GitHub Sync

bazelbuild-intellij:
This is an automatically generated pull request.


chaoren:
Actually, could we leave the next one open a bit to see what actually happens.

PR #288, opened 2018-04-16
Run/debug arbitrary main methods in Scala projects, without an explicit scala_binary rule.

### LINKED ISSUE #247: Feature Request: generate a temp binary target for main classes
We have a flow where many developers who are working on a piece of code will write an improvised `main` method and will try to run it to see some results.
This is in code which is essentially a `library` rule and they just want to "play" with it.
In maven this works without a problem and what I wanted to suggest is to somehow generate a *_binary target (rule name by config?) when the developer clicks a main method (or Scala `app`) which isn't part of a binary target.
The binary target should have a main_class pointing to the element the developer clicked and should have a runtime dependency on the library target that exists

hmemcpy:
Fixes #247 

After considerable effort, we decided on a different approach: build a deployable jar for a given `scala_library` target, and running it with a JarApplication run configuration, and a main method.

This supports all scenarios (run, debug), without introducing any temporary files in the workspace.

It works by having a before-build task that executes `bazel build //:foo_deployable.jar` (similarly to the `BuildPluginBeforeRunTaskProvider`).


ittaiz:
@hmemcpy awesome!
Is this still WIP?

hmemcpy:
@ittaiz *sigh*, unfortunately, yeah... found an issue where "Debug" entry does not appear for an existing run target. So if you create it the first time, then try to run it again, no way to debug. Still investigating...

hmemcpy:
About the missing debug: the problem stems from a missing `targetKind` value in the selected configuration, after loading a solution. The reason is, during loading, all configurations are updated (from `BlazeRunConfigurationSyncListener#updateExistingRunConfigurations`), and it attempts to update the target kind, by looking up existing targets.

Our newly created temporary target does not appear in the `targetsMap`, even after re-synching, and returns `null` from the query in `updateTargetKindAsync`.

Need to understand how to add this target to all targets.

hmemcpy:
Regarding the prompt to add the file: adding the directory `ijwb_tmp` to `.gitignore` would solve this, however we can't assume such file exists. Alternatively, we could put it in IntelliJ's own "ignored files" for Version Control, but I don't know how to do it programmatically. Also, this would not prevent these files from appearing in git changed files.

ittaiz:
@chaoren @brendandouglas any thoughts?

hmemcpy:
An update on the status (and proposed operation) of this PR:

Given a new `object Foo extends App` anywhere inside a Scala library, if no corresponding `scala_binary` target exists, will do the following:

 1. create a run configuration for the target, executing a target located in `//ijwb_tmp`, with a custom before-launch command, that, when executed, will: 
 2. create a file in `<workspace_root>/ijwb_tmp/BUILD`, adding a `load` statement for rules_scala, and a `scala_binary` target, containing the reference to the library where the `main()` was defined.
 3. the target will be executed with `--nocheck_visibility`
 3. to prevent sync issues (due to visibility), an additional `--delete_packages=ijwb_tmp` should be passed to all commands, skipping the temp target.

Open issues:

1. When creating the temp `ijwb_tmp/BUILD`, IntelliJ will prompt to add it to git. The directory `ijwb_tmp/` should be explicitly ignored by the user in `.gitignore`
2. Due to `--delete_packages`, the newly created target is not processed during sync. This causes, among other things, the Debug operation on that target to stop working. Without this flag, however, sync will fail due to missing visibility rule from the actual `scala_library` target to our temporary one.
3. The temp directory is not currently being ignored by the project view, has to be manually added to `.bazelproject` with a `-` prefix.

Would appreciate any input/ideas on how to solve these issues (#2 is critical)

hmemcpy:
After giving it some further thought, and taking into account the suggestions by @chaoren in #247, there are 4 possible courses of action we can take:

1. Use the current proposal as (with limitations, but they can be addressed)
2. Build a `//:foo_deploy.jar` with an additional `main_class` entry (passed in via `--define`, requires a small change in rules_scala), then execute it with `java -jar bazel-bin/foo_deploy.jar` (via JAR Application run configuration)
3. Conditionally delegate to a custom rule within rules_scala (from a custom implicit output target)
4. Modify `scala_library` to conditionally generate an executable output

Definitely welcome any other ideas that minimize modifications!

hmemcpy:
To make it extra clear, we think there's no way currently to do 3 and 4 in bazel, but perhaps you have some ideas if/how that's indeed possible.

brendandouglas:
How about changing 4 to 'always generate an executable output, but put it in a custom output group'?

Then the action won't be run unless that output group is requested (which it wouldn't be, except in this one particular situation).

hmemcpy:
Alright, after a lot of back and forth, this PR is ready for review.
Thanks!

scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
```diff
@@ -0,0 +1,140 @@
...
+    Project project = config.getProject();
+    RunManagerEx runManager = RunManagerEx.getInstanceEx(project);
+    runManager.setBeforeRunTasks(config,
+      ImmutableList.of(new GenerateExecutableDeployableJarProviderTaskProvider.Task()));
+  }
+
+  public File getDeployJarFile(Label target, Project project) {
+    WorkspaceRoot root = WorkspaceRoot.fromProject(project);
+    return root.fileForPath(WorkspacePath.createIfValid(
+      String.format("bazel-bin/%s_deploy.jar", target.targetName())));
```
brendandouglas: This relies on some bazel implementation details (and won't work internally, for example). Can you get this from the BuildResultHelper in the before-run task instead, and fill in the additional run configuration details then?

scala/src/com/google/idea/blaze/scala/run/producers/GenerateExecutableDeployableJarProviderTaskProvider.java
```diff
@@ -0,0 +1,181 @@
...
+    return "Generate executable deployable JAR for custom target";
+  }
+
+  @Override
+  public final boolean canExecuteTask(RunConfiguration configuration, GenerateExecutableDeployableJarProviderTaskProvider.Task task) {
+    return isValidConfiguration(configuration);
+  }
+
+  private boolean isValidConfiguration(RunConfiguration runConfiguration) {
+    return runConfiguration instanceof JarApplicationConfiguration;
```
brendandouglas: What determines whether this is a JarApplicationConfiguration? Should DeployableJarRunConfigurationProducer specify this explicitly, rather than supporting any ApplicationConfiguration type?

scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
hmemcpy: Hmm, I think I have a bit of a chicken-and-egg problem here, because I need to provide the path to .jar in the configuration details prior to my before-task is allowed to run, I'm not quite sure if/how to do this after it had finished (can I even update a configuration object outside of its producer?)

Alternatively, is there a better way to "guestimate" the output jar location? Any helpers to give me the projected output directory?

scala/src/com/google/idea/blaze/scala/run/producers/GenerateExecutableDeployableJarProviderTaskProvider.java
hmemcpy: @brendandouglas oops, this was a leftover from a previous attempt, where I tried using the JarApplication dialog. I'll update it, it should be `ApplicationConfiguration` instead! Thanks!

scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
hmemcpy: @brendandouglas Additional issue I thought of is the `doIsConfigFromContext` - I would need the jar path to compare with the existing value. So any way of retrieving the output path for the deploy jar before actual execution would be very helpful.
Do you have any suggestions?

ittaiz:
@brendandouglas can we advance this?

scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
brendandouglas: If you replace 'bazel-bin' by BlazeInfo#blazeBinKey, that will be an improvement.

Otherwise you could also replace the use of ApplicationConfiguration with a custom run configuration type which doesn't require the deploy jar location to be known up front.

This is fine though -- if you like I'll take a look at these things when upstreaming the PR.

hmemcpy: @brendandouglas if you could look at that upstream, it would be great. I'd really love to get this merged so we can at least start using it.
I'll send a change for the `bazel-bin` string.

ittaiz:
Brendan,
Can we get this forward? Part of the problem is that after this lands it still takes ~ a month before we can use it

brendandouglas:
I've applied the patch upstream -- it will be in the next release (scheduled for June 1st).

ittaiz:
Thanks!

PR #291, opened 2018-04-17
intellij_ide_info.proto update for Kotlin (please patch proto_deps.jar)

hsyed:
@brendandouglas I could really do with an update to the `proto_deps.jar` I have been blocked on this for a long time.

I need to update `IdeInfoFromProtobuf` and I can't find a way to include the updated generated proto classes. 

I tried adding `//proto:intellij_ide_info_java_proto` above `//proto:proto_deps` but this causes other problems.  This was I think what you suggested (placing the updated package jar higher up in classpath)?

### DIFF
--- a/proto/intellij_ide_info.proto
+++ b/proto/intellij_ide_info.proto
@@ -172,7 +172,12 @@ message Dependency {
   DependencyType dependency_type = 2;
 }
 
-// Next Available: 30
+// Contains the location of a Json encoded file containing Kotlin Toolchain Info..
+message KtToolchainIdeInfo {
+  ArtifactLocation location = 1;
+}
+
+// Next Available: 31
 message TargetIdeInfo {
   string label = 1 [deprecated = true];
   repeated string dependencies = 4 [deprecated = true];
@@ -215,4 +220,5 @@ message TargetIdeInfo {
   DartIdeInfo dart_ide_info = 28;
 
   AndroidAarIdeInfo android_aar_ide_info = 29;
+  KtToolchainIdeInfo kt_toolchain_ide_info = 30;
 }


brendandouglas:
What problems do you see when rebuilding 'proto:intellij_ide_info_java_proto' from source?

hsyed:
The proto library builds fine on it's own and I am already using it in the aspect tests. I need access to the updated proto messages in `base` and base picks up the proto generated package from `proto_deps.jar`.

I tried updating the rule to 

```
java_library(
    name = "base",
    srcs = glob(["src/**/*.java"]),
    javacopts = ["-Xep:FutureReturnValueIgnored:OFF"],
    resources = glob(["resources/**/*"]),
    visibility = ["//visibility:public"],
    deps = [
        "//common/actionhelper",
        "//common/binaryhelper",
        "//common/concurrency",
        "//common/experiments",
        "//common/formatter",
        "//common/guava",
        "//common/transactions",
        "//intellij_platform_sdk:jsr305",
        "//intellij_platform_sdk:plugin_api",
        "//proto:intellij_ide_info_java_proto", # <--- was hoping this would work if it's earlier in CP 
        "//proto:proto_deps",
        "//sdkcompat",
        "//third_party/auto_value",
    ],
)

trying to build base leads to:

```
Error:(3, 1) Building base/libbase-class.jar (598 source files) and running annotation processors (AutoValueProcessor) failed (Exit 1).
Error:(111, -1) no suitable method found for merge(InputStreamReader,com.google.devtools.intellij.ideinfo.IntellijIdeInfo.TargetIdeInfo.Builder)
```

**Comment by hsyed** [comment]
@plumpy I tried working around it by doing what you suggested, it wasn't that simple. I then tried patching the `deps.jar` with the generated java proto, but it seems the original classes were not generated with the standard proto-java compiler. The only way is to patch the jar with whatever process you guys have or to update the code so that it doesn't use a prepackaged jar.  

So I have a bit of a rant now. I have been trying to update the Kotlin plugin for 3 months now and I am blocked on this. I have asked to get the `deps.jar` patched with a single new proto field on numerous occasions.  I have seen multiple syncs with Github that have added new entries to the proto file the patch could have been bundled with that. I have spend a considerable amount of time trying to find a work around to this extremely simple problem.

**Comment by brendandouglas** [comment]
Fair enough -- I'll take a look at submitting this today.

File: proto/intellij_ide_info.proto [hunk 1]
```diff
@@ -172,7 +172,12 @@ message Dependency {
   DependencyType dependency_type = 2;
 }
 
-// Next Available: 30
+// Contains the location of a Json encoded file containing Kotlin Toolchain Info..
+message KtToolchainIdeInfo {
```
brendandouglas: Any objections to calling this KotlinToolchainIdeInfo?

proto/intellij_ide_info.proto
```diff
@@ -172,7 +172,12 @@ message Dependency {
...
+  ArtifactLocation location = 1;
```
brendandouglas: If there a more descriptive name we can use for this? (location is implied by the type, so maybe 'json_info_file' or similar would give more information?)

hsyed: In reply to both naming comments, Yes thats fine.

When I look at the other aspects `jvm`, `js` etc I plan to just piggy back over the same `json_info_file` -- one concern might be that native might not be well suited to IJWB. So I wonder if there is value in discriminating this in the model.

PR #307, opened 2018-05-06
Extracting the left operand name of an infix expression using an unapply

### LINKED ISSUE #303: Exception when running a specs2 text in IntelliJ 2018.1
```
org.jetbrains.plugins.scala.lang.psi.api.expr.ScInfixExpr.lOp()Lorg/jetbrains/plugins/scala/lang/psi/api/expr/ScExpression;
java.lang.NoSuchMethodError: org.jetbrains.plugins.scala.lang.psi.api.expr.ScInfixExpr.lOp()Lorg/jetbrains/plugins/scala/lang/psi/api/expr/ScExpression;
	at com.google.idea.blaze.scala.run.Specs2Utils.getSpecs2TestName(Specs2Utils.java:69)
	at com.google.idea.blaze.scala.run.Specs2Utils.getSpecs2ScopedTestName(Specs2Utils.java:74)
	at com.google.idea.blaze.scala.run.producers.BlazeScalaTestRunLineMarkerContributor.getSpecs2TestUrl(BlazeScalaTestRunLineMarkerContributor.java:134)
	at com.google.idea.blaze.scala.run.producers.BlazeScalaTestRunLineMarkerContributor.getTestUrl(BlazeScalaTestRunLineMarkerContributor.java:97)
	at com.google.idea.blaze.scala.run.producers.BlazeScalaTestRunLineMarkerContributor.getInfo(BlazeScalaTestRunLineMarkerContributor.java:84)
	at com.google.idea.blaze.scala.run.producers.BlazeScalaTestRunLineMarkerContributor.getInfo(BlazeScalaTestRunLineMarkerContributor.java:71)
   ...
```

Seems that `ScInfixExpr.lOp()` was removed in 2018.x, and replaced with [`ScInfixExpr.left`](https://github.com/JetBrains/intellij-scala/blob/idea181.x/scala/scala-impl/src/org/jetbrains/plugins/scala/lang/psi/api/expr/ScInfixExpr.scala#L21).

How are API changes handled between plugin versions?

hmemcpy:
Fixes #303 

This works around a renamed API between SDK versions.

### DIFF
--- a/scala/src/com/google/idea/blaze/scala/run/Specs2Utils.java
+++ b/scala/src/com/google/idea/blaze/scala/run/Specs2Utils.java
@@ -18,14 +18,10 @@ package com.google.idea.blaze.scala.run;
 import com.google.idea.blaze.base.run.smrunner.SmRunnerUtils;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.util.PsiTreeUtil;
-import org.jetbrains.plugins.scala.lang.psi.api.expr.ScExpression;
 import org.jetbrains.plugins.scala.lang.psi.api.expr.ScInfixExpr;
-import org.jetbrains.plugins.scala.lang.psi.api.expr.ScInfixExpr$;
-import org.jetbrains.plugins.scala.lang.psi.api.expr.ScReferenceExpression;
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScTypeDefinition;
 import org.jetbrains.plugins.scala.testingSupport.test.TestConfigurationUtil;
 import org.jetbrains.plugins.scala.testingSupport.test.structureView.TestNodeProvider;
-import scala.Tuple3;
 
 import javax.annotation.Nullable;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


chaoren:
Would get first child work?

hmemcpy:
Hmm, I haven't checked, to be honest.
I used the `lOp()` API from [your suggestion](https://github.com/bazelbuild/intellij/pull/155#discussion_r150332749), back when I implemented specs2 support.
If I recall correctly, my original implementation had looking up a child `REFERENCE_EXPRESSION`... see that commit [here](https://github.com/hmemcpy/intellij/blob/88f4debcf49bd5f4e9f50d0605f12468ef9a4036/scala/src/com/google/idea/blaze/scala/run/producers/BlazeSpecs2ConfigurationProducer.java#L135-L148)

Perhaps I cold just really look up the first child and it would work. Certainly would be cleaner than calling Scala-specific constructs from Java.

hmemcpy:
After taking an additional look, seems that `firstChild` would work here, as it gives me the Literal element.

PR #316, opened 2018-05-14
kotlin v3 changes

hsyed:
* KtToolchainIdeInfo from Kotlin rules to relay the Kotlin plugin configuration from the workspace.
* Aspect processing for Kotlin rules (currently the KtToolchainIdeInfo is coming in via this but I will update this to inject the stdlibs from here as well, currently these are hard coded and discovered by the plugin).
* Attach srcjar jars to targets for code generated code.
* configure plugin elements found in KtToolchainIdeInfo (this deprecates the config section -- I will remove it outright in v4).
* a start to some aspect integration tests.

The code has been in use in our production repo for quite a while, so it's quite well tested.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -11,3 +11,4 @@
 
 Brendan Douglas <brendandouglas@google.com>
 Tom Lundell <tomlu@google.com>
+Hassan Syed <h.a.syed@gmail.com>
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -277,10 +277,10 @@ scala_repositories()
 # LICENSE: The Apache Software License, Version 2.0
 git_repository(
     name = "io_bazel_rules_kotlin",
-    commit = "6d8dcd4d6000d0cf3321eb8580d8fc67f8731f8e",
+    commit = "2ab126808329700c6c23df5d571762e77fbd9efc",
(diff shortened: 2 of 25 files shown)


brendandouglas:
Hi Hassan,

Thanks for the patch. I've just started working on adding kotlin support internally, which in the long term should make the process for patches such as this much smoother, but in the short term will delay review of this patch, since I'll likely make some changes to the core kotlin sync plugin to support both blaze and bazel rules.

I'll keep you updated.

hsyed:
The patch is ready now. I have spent a lot of time to get this patch ready for submission. I can't do this all over again after things have changed. The further work that I had planned to the rules and for the Kotlin aspect of the plugin depends on this patch going in. 

If you guys are going to support Kotlin then I'll hand over the work to you. There is a close interaction between Kotlin rules and the plugin so it's best you take over the overall direction of the rules as well (@davidstanke).

brendandouglas:
I won't ask you to redo any of this work, just meant that I want to delay adding any major comments to this patch until after the internal work is done (likely 1-2 days), when I'll have a better understanding how it all fits together, and can find any potential longer-term issues with internal/external compatibility.

We'll need to apply this manually upstream anyway, so I'll handle any merge conflicts then.

Issue #387, opened 2018-08-07
Unable to Run/Debug when using java_binary and Kotlin main class


zsiegel:
Hello

I have a mixed java/kotlin project that works quite well with Intellij. However the current `kt_jvm_binary` rule is not a full replacement for the `java_binary` rule so in order to use it you end up with something like the following.

```
java_binary(
    name = "server",
    main_class = "com.zsiegel.example.ApiServer",
    runtime_deps = [":api"]
)

kt_jvm_library(
    name = "api",
    srcs = glob(["src/main/kotlin/**/*.kt"]),
    deps = [
        "//config:config",
    ]
)
```

This allows for building a typical `deploy_jar` using the standard java binary and using a kotlin main function. However since the main file to be run is Kotlin the run/debug tooltip on the left does not have any suggestions for adding targets. I can manually add the target and run/debug just fine.

<img width="389" alt="screenshot 2018-08-07 12 32 59" src="https://user-images.githubusercontent.com/304495/43792416-3110d6ac-9a3e-11e8-86b6-842950c9d025.png">

I am not sure if this support should be added in the kotlin module or java. I suspect it should go in the Java module since it is I am guessing recognizing the `java_binary`

I think the current support in the Java run provider package does not seem to recognize Kotlin main files as being available to run. Could this be added or adjusted? Id be happy to open a PR to address this if someone could point me in the right direction within the `run` package in the `java` module

hsyed:
@zsiegel are you missing a Kt at the end of the class name ?

hsyed:
@zsiegel are you fully synced up ? 

@brendandouglas this used to work if I remember correctly I connected `java_binary` to kt run markers -- could this have regressed ?

hsyed:
This is also showing up in the coffee app example in the rules repo. kotlinx.coroutines dependency is not being synced either.

hsyed:
on the kotlinx.coroutines dependency not being added to the library table:

I have the jar cache enabled.

1. I did a clean expunge and then a non-incremental sync and the dep did not attach.
2. I worked through some previous commits to see if I regressed something -- the first old commit I went to and synced against attached kotlinx.coroutines. 
3. I worked my way to HEAD. 
4. then when I got back to head the dep is still attached. 
5. I then did step 1 again and this time the jar did attach. 

Could this be a  problem with the jar cache ?

zsiegel:
@hsyed yes I am fully synced up! I tried on two difference machines and experience the same issue.

[...]

PR #390, opened 2018-08-08
java_binary should line up with kotlin run markers.

hsyed:
Simple fix for #387 -- this regressed or it was never included from my kotlin-v3 work.

### DIFF
--- a/kotlin/src/com/google/idea/blaze/kotlin/run/producers/BlazeKotlinRunConfigurationProducer.java
+++ b/kotlin/src/com/google/idea/blaze/kotlin/run/producers/BlazeKotlinRunConfigurationProducer.java
@@ -152,6 +152,6 @@ public class BlazeKotlinRunConfigurationProducer
         project,
         projectData.artifactLocationDecoder,
         projectData.targetMap,
-        (target) -> target.kind == Kind.KT_JVM_BINARY && target.isPlainTarget());
+        (target) -> (target.kind == Kind.KT_JVM_BINARY || target.kind == Kind.JAVA_BINARY) && target.isPlainTarget());
   }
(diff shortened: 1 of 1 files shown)


hsyed:
@chaoren @brendandouglas could you have a look and upstream this simple upstream this simple PR please.

kotlin/src/com/google/idea/blaze/kotlin/run/producers/BlazeKotlinRunConfigurationProducer.java
```diff
@@ -152,6 +152,6 @@ private static FilteredTargetMap computeTargetMap(Project project, BlazeProjectD
         project,
         projectData.artifactLocationDecoder,
         projectData.targetMap,
-        (target) -> target.kind == Kind.KT_JVM_BINARY && target.isPlainTarget());
+        (target) -> (target.kind == Kind.KT_JVM_BINARY || target.kind == Kind.JAVA_BINARY) && target.isPlainTarget());
```
chaoren:
```
KotlinRunConfigurationProducer.Companion.getEntryPointContainer(location.getPsiElement());
```
above should return `null` for `java_binary`s without a kotlin main method, correct?

Just making sure this wouldn't show up for regular `java_binary`s.

chaoren:
Merged upstream.

kotlin/src/com/google/idea/blaze/kotlin/run/producers/BlazeKotlinRunConfigurationProducer.java
hsyed: correct, and thanks !

PR #404, opened 2018-09-03
Run main with deployable jars

hmemcpy:
This pr fixes #288, the original goal was to create a deployable jar based on the current `scala_library`, and run it with a specified classpath main.
Seems the original PR got changed when it was merged upstream, breaking the pr (it looked up binary targets instead)
This PR restores the functionality with a few tweaks.

Please review

### DIFF
--- a/scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
+++ b/scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
@@ -15,29 +15,33 @@
  */
 package com.google.idea.blaze.scala.run.producers;
 
-import static com.google.idea.blaze.scala.run.producers.BlazeScalaMainClassRunConfigurationProducer.getMainObject;
-
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import com.google.idea.blaze.base.dependencies.TargetInfo;
 import com.google.idea.blaze.base.ideinfo.TargetIdeInfo;
-import com.google.idea.blaze.base.model.BlazeProjectData;
 import com.google.idea.blaze.base.model.primitives.Label;
+import com.google.idea.blaze.base.model.primitives.RuleType;
+import com.google.idea.blaze.base.run.SourceToTargetFinder;
 import com.google.idea.blaze.base.run.producers.BlazeRunConfigurationProducer;
-import com.google.idea.blaze.base.sync.data.BlazeProjectDataManager;
 import com.google.idea.blaze.java.run.RunUtil;
 import com.intellij.execution.RunManagerEx;
 import com.intellij.execution.actions.ConfigurationContext;
 import com.intellij.execution.application.ApplicationConfiguration;
 import com.intellij.execution.application.ApplicationConfigurationType;
+import com.intellij.execution.configurations.RunConfigurationBase;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.util.Ref;
 import com.intellij.psi.PsiClass;
 import com.intellij.psi.PsiElement;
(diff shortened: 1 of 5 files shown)


hmemcpy:
Now with a test :)

scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
```diff
@@ -95,46 +90,34 @@ protected boolean doIsConfigFromContext(
...
+  private static TargetInfo findTarget(Project project, ScObject mainObject) {
     File mainObjectFile = RunUtil.getFileForClass(mainObject);
     if (mainObjectFile == null) {
       return null;
     }
 
-    Collection<TargetIdeInfo> targets =
-        BlazeScalaMainClassRunConfigurationProducer.findScalaBinaryTargets(project, mainObjectFile);
+    Collection<TargetInfo> targets =
+            SourceToTargetFinder.findTargetsForSourceFile(
```
brendandouglas: This is going to find all library rules reachable from the current source file (in the transitive closure). Is that actually what you want here?

SourceToTargetProvider (the almost duplicate APIs are confusing and need to be redone) will return only targets directly building the given source file -- is that enough?

brendandouglas:
Before I forget -- I'm in the process of upstreaming this, and the only major change I've made is to the bazel build task (and output artifact handling), to make it more consistent with similar code in the project (e.g. BlazePyRunConfigurationRunner#executeBeforeRunTask).

The test still passes, but that doesn't exercise the bazel build parts -- I'll let you know when it lands, then could you check it's still working? (submitting it with the unverified changes is going to be quicker than porting changes to and from github a few more times).

scala/src/com/google/idea/blaze/scala/run/producers/DeployableJarRunConfigurationProducer.java
hmemcpy: Hmm, I need to check, thanks for the heads up. Will report back shortly.

Tradunsky:
Hi guys,

Any updates on this? Really need this to start using Bazel with full power of IJ

ittaiz:
Hi,
Where does this stand? People really need this on our side

brendandouglas:
I held off applying it upstream given the concern about searching the transitive closure of targets reachable from a source file (the pending question above).

This could cause performance problems, and doesn't seem generally advisable. Perhaps you'd be fine limiting the search to direct deps?

ittaiz:
You’re right on both counts (not merging and the fact that we should use
only the direct targets).
I’ll update tomorrow with final word.
Thank you and sorry for the delay

hmemcpy:
Sorry for the (very) late reply, Brendan.
I wasn't able to, unfortunately, use the `SourceToTargetProvider` API - it always returned an empty result for me.

Perhaps using the bigger API is fine in this case, because this feature is very much an "on-demand" usage, and shouldn't happen very often.

Maybe you could recommend the correct usage of the `SourceToTargetProvider` API?

brendandouglas:
Ah sorry, SourceToTargetProvider is only ever used for finding targets outside the project (and as you say, doesn't have an implementation in the bazel plugin). The API definitely needs a cleanup.

In this case, perhaps using the BlazeProjectData target map directly is better, so you can limit to direct deps? If that's not a problem for your use case, I'm happy enough to leave this as is, since it's scala specific.

brendandouglas:
I ended up using SourceToTargetMap instead of SourceToTargetProvider, so it's only looking for direct deps. It should land soon -- let me know if it's not working as expected.

ittaiz:
Thanks!
By soon do you mean the next release to appear?

brendandouglas:
I meant it would land in the source code (it's just been merged). It will be in the release after next (we cut the release branch last Monday). It's scheduled to be uploaded to the JetBrains' plugin repo Friday December 28th, but it will probably be pushed back a week due to the holidays.

PR #417, opened 2018-09-25
Add fallback runfiles path for setting debug CWD.

mtsgrd:
There is a default workspace name `__main__` that is used when none is
declared in a WORKSPACE file. This commit adds this as a fallback path
since the expected path derives the workspace name from the root
directory.

See Bazel change log at: https://github.com/bazelbuild/bazel/blob/master/CHANGELOG.md#release-023-2016-05-10


mtsgrd:
Note that `__main__` is printed as **main** in markdown formatting in the Bazel changes.

sgowroji:
Hello @mtsgrd,  Are you still looking to submit this PR? Could you please fix the buildkite checks, so that i can share it for review. Thanks!

PR #428, opened 2018-10-21
Wrong method was used to get the first character

hmemcpy:
The `indexOf(int ch)` method was not the right thing to use here. Replaced with `charAt`.


brendandouglas:
Thanks, I'll submit this upstream.

PR #430, opened 2018-10-29
Add java language level 11.

uri-canva:
Based on the information in https://github.com/JetBrains/intellij-community/blob/3a4aa66df493c87d41135cb2d72195274c031827/java/java-psi-api/src/com/intellij/pom/java/LanguageLevel.java which is included in 182.4892.20, tested it successfully in IntelliJ Idea UE 182.4892.20.


brendandouglas:
Thanks, I'll add support for this upstream (we'll probably just delegate parsing to LanguageLevel rather than hard-coding a list of language levels here).

PR #433, opened 2018-11-02
Silence warnings about illegal reflective access operations

philwo:
See https://github.com/bazelbuild/bazel/issues/6151#issuecomment-432774820 for an explanation.

cc @cushon


wcurrie:
Looks like half of this made it's way to master https://github.com/bazelbuild/intellij/pull/432/files#diff-12325344c329e32d88387038d7b3ecdbR36

wcurrie:
IIUC these flags need to be in a different file in order to be bundled with the plugin .zip. Eg https://github.com/wcurrie/intellij/pull/1

(Didn't mean to open the PR against my fork 😊. PRs don't get merged here anyhow it seems..)

PR #452, opened 2018-11-21
Make Python scripts in build_defs compatible with Python 3.

philwo:
Some distributions like Arch Linux no longer ship Python 2 by default
and /usr/bin/python on these systems is a Python 3.x version, which
causes Bazel to try to run them with Python 3.x as part of the build.

With these changes, I was able to successfully build the plug-in on my
Arch Linux machine. I made sure that the changes are backwards
compatible with Python 2.x.

### DIFF
--- a/build_defs/stamp_plugin_xml.py
+++ b/build_defs/stamp_plugin_xml.py
@@ -203,7 +203,7 @@ def main():
   for new_element in new_elements:
     idea_plugin.appendChild(new_element)
 
-  print dom.toxml(encoding="utf-8")
+  print(dom.toxml())
 
 
 if __name__ == "__main__":
--- a/build_defs/stamp_plugin_xml.py
+++ b/build_defs/stamp_plugin_xml.py
@@ -203,7 +203,7 @@ def main():
   for new_element in new_elements:
     idea_plugin.appendChild(new_element)
 
-  print(dom.toxml())
+  print(dom.toxml(encoding="utf-8").decode("utf-8"))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


build_defs/stamp_plugin_xml.py
```diff
@@ -203,7 +203,7 @@ def main():
   for new_element in new_elements:
     idea_plugin.appendChild(new_element)
 
-  print dom.toxml(encoding="utf-8")
+  print(dom.toxml())
```
plumpy: FYI the encoding was there for a reason, please don't remove it. It caused errors when we didn't have it, I believe because we have non-ASCII characters in our changelogs.

The docs for 2.7 specifically say you should ALWAYS include it: https://docs.python.org/2.7/library/xml.dom.minidom.html#xml.dom.minidom.Node.toxml

philwo: No problem. I had to remove this, because `dom.toxml(encoding="utf-8")` returns a byte-string, but `print` only accepts text-strings.

I'll add a commit that should restore the previous behavior.

philwo: FYI, in `merge_xml.py` you already had a plain `print(dom.toxml())` without a specified encoding - shall we keep this as is, or should I add a drive-by fix for that file as well? :)

drichelson:
It looks like this should resolve #453

philwo:
@drichelson Yes, I can confirm that was one of the errors I also saw on my machine. :)

@brendandouglas @plumpy I've imported the change internally in our code review system for review and merging.

philwo:
Closing, because this has been merged and pushed as part of https://github.com/bazelbuild/intellij/commit/f054292b07806a288eb205d13f336badbbe1464b :)

PR #474, opened 2018-12-18
add default flags section for single test class run configuration

berman7:
Adding a section in .bazelproject files `single_test_flags`, listing all default flags needed for a single test run.

### DIFF
--- a/base/src/com/google/idea/blaze/base/projectview/section/sections/Sections.java
+++ b/base/src/com/google/idea/blaze/base/projectview/section/sections/Sections.java
@@ -41,7 +41,8 @@ public class Sections {
           RunConfigurationsSection.PARSER,
           ShardBlazeBuildsSection.PARSER,
           TargetShardSizeSection.PARSER,
-          BazelBinarySection.PARSER);
+          BazelBinarySection.PARSER,
+          SingleTestFlagsSection.PARSER);
 
   public static List<SectionParser> getParsers() {
     List<SectionParser> parsers = Lists.newArrayList(PARSERS);
--- a/base/src/com/google/idea/blaze/base/projectview/section/sections/SingleTestFlagsSection.java
+++ b/base/src/com/google/idea/blaze/base/projectview/section/sections/SingleTestFlagsSection.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2016 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


berman7:
I signed it! :)

or-shachar:
## motivation
At Wix - we would like to add a flag `--test_output=streamed` to each generated test configuration for single class. It makes sense to us because we run test locally and it allows us to follow logs as the test is running. 
We don't want this flag to apply when running multiple tests since we lose the parallelism.

ittaiz:
Also we have fine grain targets and we don’t use sharping. OTOH not having
the logs visible as the tests are running is a big pain.

ittaiz:
@brendandouglas wdyt?

or-shachar:
@brendandouglas - seems like this PR became stale since your last copybara sync 😭

brendandouglas:
Is the 'single test class' distinction going to be broadly useful? There are a lot of test classes around with expensive test cases, or large numbers of test cases, for which you wouldn't generally want to run the tests serially.

I'm not sure the extra complexity added to the already complex area of test flags is worth the benefit of seeing test logs as the test is running (how long do these tests take to run?).

or-shachar:
> Is the 'single test class' distinction going to be broadly useful? There are a lot of test classes around with expensive test cases, or large numbers of test cases, for which you wouldn't generally want to run the tests serially.

We understand that - and that's why we added this as a configuration so users can opt in to this if they want. 

> I'm not sure the extra complexity added to the already complex area of test flags is worth the benefit of seeing test logs as the test is running (how long do these tests take to run?).

In our case, for instance, we have some e2e tests that tries to boot the test env for up to 10 minutes (with retries), if there's a critical problem that was printed to the logs - developer would want to see the log as soon as possible rather than wait for the test to fail... 

In any case - since it's a very generic feature and do not require anyone to make special changes to their configuration unless they want to - I don't understand why not to put it in.

hmemcpy:
Apologies for the delays! I rebased this PR on top of master. Had to re-introduce @berman7's changes to the newly refactored `TestContext` strategies. It's a much nicer abstraction, btw!
Tests should be passing again!

sha1n:
@brendandouglas if you agree to accept that feature, I want to close this PR and open a new one that implements it like `test_flags` is implemented - in a way that does not effect run configurations, but only the actual bazel command generated for the tests (related to [this](https://github.com/bazelbuild/intellij/issues/562) issue).

I'm ready to create the PR from [this](https://github.com/wix-playground/intellij/commit/831d8b3c775d15f1f175926dd8e037abe46463af) commit.

What do you think?

@or-shachar, @hmemcpy - FYI

brendandouglas:
I'm not keen to add more options to the project view file, without a clear benefit. In this case, I don't see why you'd want flags which act only when running a single test case.

I understand there are long-running tests in your code base, and they are also single test cases, but not why that would be a common-enough case to justify it's own project view section.

sha1n:
Thanks @brendandouglas. I understand your standpoint and after pulling my head out of the code and looking at the bigger picture, I actually agree. I will have a discussion about this with my colleagues, to decide whether we want this feature implemented as an extension of the plugin or drop it. If we will decide to do that, we will need to find a way to do that using extension points, which is very similar to what I requested on #537. I think that with the right API we can easily solve both use-cases.

**You can close this PR** and I will add this use-case to #537.

@or-shachar - FYI

PR #480, opened 2019-01-02
avoid false targets if it contains `build` directory instead of build file

mskonovalov:
Currently to form the bazel command plugin gets workspace data about added files and modified files. After that it checks if these targets are real bazel targets (have corresponding build files).
The problem occurs when there is no build file but instead directory with the `build` name.
It leads to further errors from bazel command
So to avoid this it is possible to add check `isFile`


mskonovalov:
I signed it!

brendandouglas:
Thanks, I've applied this upstream. It will be in the next plugin release.

mskonovalov:
great, thanx

Issue #482, opened 2019-01-04
Bazel with incompatible_disallow_data_transition


laurentlb:
Based on the CI, the plugin doesn't work with `bazel --incompatible_disallow_data_transition`: https://buildkite.com/bazel/bazel-at-release-plus-incompatible-flags/builds/48#de3c42b3-4841-4c2f-9c0c-272f51568af7

Since the error message is in scala.bzl, can you please update the dependency on Scala? It has been fixed there.

brendandouglas:
Thanks, this should be fixed at head now.

PR #485, opened 2019-01-07
Update the versions of rules_scala and rules_kotlin to the latest commits

vladmos:
The currently used versions contain the deprecated `cfg = "data"` configuration which will not be compatible with the next version of Bazel.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -260,7 +260,7 @@ http_archive(
 # LICENSE: The Apache Software License, Version 2.0
 git_repository(
     name = "io_bazel_rules_scala",
-    commit = "8359fc6781cf3102e918c84cb1638a1b1e050ce0",
+    commit = "326b4ce252c36aeff2232e241ff4bfd8d6f6e071",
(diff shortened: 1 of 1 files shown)


laurentlb:
This should fix https://github.com/bazelbuild/intellij/issues/482

meteorcloudy:
I heard @buchgr has write access to everything under bazelbuild ;)

laurentlb:
This repository is synced with Google depot. I'd prefer to hear from an owner (cc @brendandouglas) before escalating to a bazelbuild admin.

meteorcloudy:
Ah, make sense!

brendandouglas:
Thanks for the PR, I've applied it upstream (we don't make changes to this repo directly).

I'll merge the pending [PR from upstream](https://github.com/bazelbuild/intellij/pull/470) when the CI is passing -- looks like it's currently failing due to an old version of the protobuf library using REPOSITORY_NAME.

I'll update that library, and close #482 once the CI is passing.

Issue #441, opened 2018-11-11
Can't open project in IntelliJ


solomatov:
I created a .bazelrc in the root with the following content:

    build --define=ij_product=intellij-latest

But when I import it in IntelliJ I get this:

    ERROR: /home/kostik/Projects/intellij/aswb/BUILD:112:1: Configurable attribute "srcs" doesn't match this configuration: define an intellij product version, e.g. --define=ij_product=intellij-latest.
    ERROR: /home/kostik/Projects/intellij/aswb/BUILD:87:12: Configurable attribute "srcs" doesn't match this configuration: define an intellij product version, e.g. --define=ij_product=intellij-latest.
    ERROR: /home/kostik/Projects/intellij/aswb/BUILD:28:1: Configurable attribute "srcs" doesn't match this configuration: define an intellij product version, e.g. --define=ij_product=intellij-latest.
    ERROR: /home/kostik/Projects/intellij/aswb/BUILD:112:1: Configurable attribute "srcs" doesn't match this configuration: define an intellij product version, e.g. --define=ij_product=intellij-latest.
    ERROR: /home/kostik/Projects/intellij/aswb/BUILD:142:1: Configurable attribute "srcs" doesn't match this configuration: define an intellij product version, e.g. --define=ij_product=intellij-latest.
    ERROR: /home/kostik/Projects/intellij/aswb/BUILD:142:1: Configurable attribute "srcs" doesn't match this configuration: define an intellij product version, e.g. --define=ij_product=intellij-latest.

chaoren:
You can't build aswb with `intellij-latest`. You want `android-studio-latest`.

chaoren:
When you're importing, are you using the catch-all `//...` target expression? That doesn't work for this project, you have to choose one IDE target, and define the appropriate `ij_product` for that target.

solomatov:
But which target should I choose? I want to be able to run tests from the IDE.

chaoren:
You can import `ijwb/ijwb.bazelproject` that includes IJwB and all relevant tests, also sets the `--define` flag so you don't need it in your .bazelrc.

solomatov:
Everything worked. Thank you!

hmemcpy:
I have the same issue, after pulling the latest changes (copybara #460), I am unable to sync the project.
Importing regularly, `ijwb/ijwb.bazelproject`, it has `intellij-latest` defined, getting the same errors as above.
The previous commit (#451) is working fine.
[...]

PR #498, opened 2019-01-17
Skip project data subdir

mskonovalov:
Current behavior:
When you import the plugin IDE creates project data subdir and a bunch of files in it (e.g. `.ijwb`). If `addWorkingSet` is `true` it leads to the situation when such files are determined as edited and the target for these files is determined `//:all`
So the sync command looks like
```
bazel build --tool_tag=ijwb:IDEA:ultimate ... --output_groups=intellij-info-generic,intellij-info-java -- //ijwb:ijwb_bazel_dev //ijwb:ijwb_lib //:ijwb_tests //:ijwb_ue_tests //:all
```
which leads to building some `awsb` stuff (as it is referenced from BUILD file in the root of the project)

Expected behavior:
target `//:all` should not be added to the build command

This change solves the issue https://github.com/bazelbuild/intellij/issues/441

This PR is based on top of https://github.com/bazelbuild/intellij/pull/497 (it includes the changes from it) so that adding the project subdir to the exclude list leads to the target not being added any more.

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/projectview/ImportRoots.java
+++ b/base/src/com/google/idea/blaze/base/sync/projectview/ImportRoots.java
@@ -29,6 +29,7 @@ import com.google.idea.blaze.base.projectview.section.sections.DirectoryEntry;
 import com.google.idea.blaze.base.projectview.section.sections.DirectorySection;
 import com.google.idea.blaze.base.settings.Blaze;
 import com.google.idea.blaze.base.settings.BuildSystem;
+import com.google.idea.blaze.base.sync.data.BlazeDataStorage;
 import com.google.idea.blaze.base.util.WorkspacePathUtil;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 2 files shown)


base/src/com/google/idea/blaze/base/sync/projectview/ImportRoots.java
```diff
@@ -91,6 +92,7 @@ public ImportRoots build() {
       if (buildSystem == BuildSystem.Bazel && hasWorkspaceRoot(rootDirectories)) {
         excludeBuildSystemArtifacts();
       }
+      excludeProjectDataSubDirectory();
```
brendandouglas: Thanks for the patch, I'll apply this and #497 upstream with some minor modifications (e.g. it's only applicable to bazel projects).

mskonovalov: 👍 great, thank you

Issue #472, opened 2018-12-13
Running tests on a broken package looks successful (UI)


sha1n:
When running tests from a broken package, the console view shows the Bazel build failure, but the tests panel shows an empty test result labeled "Test Results" and marked with a green V icon, which is very similar to successful test results when "Show Passed" is turned off, and can be very confusing.

I'm trying to solve that and I'm seeking some good advice. Here is where I got so far:
I was trying to intercept that condition in [BuildEventProtocolOutputReader](https://github.com/bazelbuild/intellij/blob/b55f77a6d13fa4c03e7be21a1ea599f13551d72d/base/src/com/google/idea/blaze/base/command/buildresult/BuildEventProtocolOutputReader.java#L67), by adding a case for `BUILD_FINISHED` events, checking whether the build was successful and trying to figure out what was the reason for that failure from the value of `event.getFinished().getExitCode()`. My problem is that the information is there, but it's a string and I prefer to avoid relying on hard-coded strings in the plugin. I saw that there is a corresponding representation of `ExitCode` [in the Bazel project](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/util/ExitCode.java), but depending on Bazel code is not really an option here, so I'm looking for another way to extract that information safely and nicely or maybe another approach to solve the whole problem.

Thanks!

or-shachar:
Note: this behavior was observed on latest released plugin `2018.11.12.0.4` with intellij 2018.3.1 
cc: @brendandouglas @chaoren  - @sha1n would like to contribute a fix for it. 

Can you pls help @sha1n with his exploration or mention the right person here?

chaoren:
> My problem is that the information is there, but it's a string and I prefer to avoid relying on hard-coded strings in the plugin.

Haha. Hardcoded strings are everywhere in this plugin.
You basically want to translate a bazel `ExitCode` into a `TestStatus` enum, right?
I think it's probably best to just add a `fromExitCode` method to `TestStatus`, and hardcode all the strings there in a giant switch.

chaoren:
Wouldn't they all just be `FAILED_TO_BUILD`? So there's no need to parse the string.

brendandouglas:
This should already be fixed at HEAD -- BlazeXmlToTestEventsConverter#flushBufferOnProcessTermination now checks the exit code (BlazeTestExitStatus#forExitCode).

sha1n:
[...]

PR #507, opened 2019-02-11
Fix compare for empty test results on prematurely-terminated test run

hmemcpy:
This is a complement to #472. Seems that if test run terminated prematurely, BUT there exists a bep file without any tests, it will create an unpopulated (empty) `BlazeTestResults` object. It will fail comparison with `BlazeTestResults.NO_RESULTS`, and report the wrong status:
image

This PR adds an `.equals` implementation, which makes the comparison check for actual contents. This fixes the


brendandouglas:
This is actually already fixed in the latest plugin release up on JetBrains plugin repo (it was fixed slightly differently, by returning NO_RESULTS for empty lists, and continuing to rely on object equality).

The fix went in upstream on Jan 9th, but we haven't merged the copybara branch here since then. Looks like that's currently blocked by some Android Studio specific failures.

hmemcpy:
That's good to know, @brendandouglas. I'll close this, then!

PR #524, opened 2019-03-04
Custom buildifier and buildozer for build file generation / modification

ignasl:
Buildozer and buildifier paths are made configurable using Settings dialog box. Defaults are set to simply `buildozer` and `buildifier` so that system would pick up the one that is found anywhere in the PATH, it it is available.

Fallback scheme for `buildozer`:
- Attempt to invoke the configured one
- If configured one fails and it is not the same as default (not found, etc.) invoke the default one in path (`buildozer`)
- If default one fails, use string concatenation to generate build file
The advantage of `buildozer` over string manipulation is that `buildozer` has already implemented this and in a more thorough way, also it provides way more features thus allowing plugin to become easily extendable.

At the moment there is no fallback scheme for buildifier yet. Fallback scheme for buildifier would require design change (rewrite of `getBuildifierBinary`) at least and how it is invoked.

Note: this PR comes a bit late and branch has been created from slightly older commit, because most recent code dump made this whole feature implementation obsolete. However, it might still be a useful contribution to consider should build file modification would be deemed desirable.

### DIFF
--- a/base/src/com/google/idea/blaze/base/buildmodifier/BazelBuildifierBinaryProvider.java
+++ b/base/src/com/google/idea/blaze/base/buildmodifier/BazelBuildifierBinaryProvider.java
@@ -15,6 +15,7 @@
  */
 package com.google.idea.blaze.base.buildmodifier;
 
+import com.google.idea.blaze.base.settings.BlazeUserSettings;
 import com.google.idea.common.binaryhelper.HelperBinaryUtil;
 import java.io.File;
 import javax.annotation.Nullable;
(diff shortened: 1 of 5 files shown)


or-shachar:
@brendandouglas  - this PR is starving ... :-(  
Someone dedicated a week of work to contribute this to the project.

Was is merged by any chance to master in mysterious way? 

cc: @ittaiz @sha1n

brendandouglas:
I haven't had a chance to review yet. Some quick comments:

- we removed BuildFileModifier upstream (along with the 'new rule' action) at the start of February. I don't have any plans to add support for buildozer at the moment.
- supporting a custom buildifier binary is something we could do, but it does add complexity, so I'd want to have a good reason for it.

ignasl:
Is it fair to assume that "Add rule" feature is retired for good and it would not be replaced with anything similar?

brendandouglas:
We won't be adding back that 'add rule' action, but I'd certainly like to one day add support for automatically fixing up BUILD targets (e.g. when adding imports).

It's very difficult to do that properly though, and we have a lot of higher priorities, so I don't think it's likely any time soon.

hmemcpy:
Thanks anyway Brendan! We'll explore other options for this.

PR #528, opened 2019-03-06
Fixing Scala Test Events Handler to actually process Scala files

hmemcpy:
There is a mistake in the handler, which makes it check for incorrect language kind (Kotlin, instead of Scala)

This one fix solves several issues, most notable of which are restoring the ability to navigate to tests (and other context menu actions), and fixes trimming the `::` character from test names (which explains why this started happening again in the first place.)


brendandouglas:
Thanks for catching this, I'll submit it upstream.

PR #530, opened 2019-03-06
Ensure windows paths are converted to the correct format for blaze

tommy-yardley:
I was having issues using this on Windows. It seems to simply append the relative target path to the workspace - which works fine on Linux and MacOS but on Windows due to the different file separators means the target paths are invalid.

This takes any windows file path separators in the targetPattern and replaces it with the blaze separator.

The error I was seeing without this fix:
```
ERROR: Skipping '//main\java\example\example:all': Invalid package name 'main\java\example\example:all': package names may contain A-Z, a-z, 0-9, or any of ' !"#$%&'()*+,-./;<=>?[]^_`{|}~' (most 127-bit ascii characters except 0-31, 127, ':', or '\').
WARNING: Target pattern parsing failed.
```

### DIFF
--- a/base/src/com/google/idea/blaze/base/model/primitives/WorkspacePath.java
+++ b/base/src/com/google/idea/blaze/base/model/primitives/WorkspacePath.java
@@ -48,6 +48,7 @@ public class WorkspacePath implements ProtoWrapper<String>, Serializable {
    * @throws IllegalArgumentException if the path is invalid
    */
   public WorkspacePath(String relativePath) {
+    relativePath = replaceFileSeperator(relativePath);
     String error = validate(relativePath);
     if (error != null) {
       throw new IllegalArgumentException(
@@ -138,5 +139,10 @@ public class WorkspacePath implements ProtoWrapper<String>, Serializable {
   @Override
   public String toProto() {
     return relativePath;
+  }
+
+  /** Parse targetPattern to use the correct fileSeperator expected by Blaze - needed for windows */
+  private String replaceFileSeperator(String targetPattern) {
+    return targetPattern.replace('\\', BLAZE_COMPONENT_SEPARATOR);
   }
 }


base/src/com/google/idea/blaze/base/model/primitives/WorkspacePath.java
```diff
@@ -48,6 +48,7 @@ public static WorkspacePath createIfValid(String relativePath) {
    * @throws IllegalArgumentException if the path is invalid
    */
   public WorkspacePath(String relativePath) {
```
improbable-nickkrempel: Drive by: the comment above says:

> relative path that must use the Blaze specific separator char to separate

So is this not a case of the user/caller of this function using it incorrectly?

tommy-yardley: I was just importing a bazel project using the BUILD files - that led to this error.

This creates the following `.bazelproject`

```
directories:
  main/java/example

targets:
  //main/java/example/...:all

additional_languages:
  # Uncomment any additional languages you want supported
  # android
  # dart
  # go
  # javascript
  # kotlin
  # scala
  # typescript
```

tommy-yardley: It could be a case that the caller is using it incorrectly - I'll see if I can investigate the root at which the conversion takes place resulting in the windows line separators being introduced.

mattnathan: When I get this (on windows) it's related to `Expand sync to working set`

brendandouglas: There are a lot of places where we don't correctly convert to canonical path formats -- we really don't support Windows at all yet.

We can't make this change as is in WorkspacePath, because it's used for both bazel target labels and files/directories. The proper fix is to audit every call site (e.g. GitWorkingSetProvider is one example which you'll hit with git projects and 'Expand sync to working set' active).

I would be open to making this change for Windows only though (or more generally, whenever File.separatorChar == '\\'), since it's an already-broken use case. Eventually we'd want to remove it though, and fix up the callers instead.

keertk:
Hi there! Thank you for contributing to the IntelliJ repository. We appreciate your time and effort. We're doing a clean up of old PRs and will be closing this one since it seems to have stalled. Please feel free to reopen/let us know if you’re still interested in pursuing this or if you'd like to discuss anything further. We’ll respond as soon as we have the bandwidth/resources to do so.

PR #531, opened 2019-03-06
Added a toggle for filtering out warnings from the blaze problems view

sha1n:
This PR adds the ability to filter in/out messages with category != ERROR from the Blaze Problems View.

The `BlazeProblemsView` class now keeps messages in the `BlazeProblemsSink` in a way that allows looking into the data later and repopulate the panel.

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -99,6 +99,11 @@
       icon="BlazeIcons.Blaze">
       <add-to-group group-id="FileOpenGroup" relative-to-action="OpenFile" anchor="after"/>
     </action>
+    <action id="Blaze.ShowOnlyErrorsAction"
+      class="com.google.idea.blaze.base.ui.problems.ShowOnlyErrorsAction"
+      text="Show Errors Only"
+      icon="BlazeIcons.BlazeFailed">
+    </action>
 
     <group id="Blaze.MainMenuActionGroup" class="com.google.idea.blaze.base.actions.BlazeMenuGroup">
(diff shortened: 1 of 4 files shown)


sha1n:
@hmemcpy FYI

brendandouglas:
Thanks for the patch. I've upstreamed it, though I implemented it differently.

There's no need to make any changes to BlazeProblemsView (other than clearing the tree more thoroughly between runs) -- I've just added a copy of NewErrorTreeViewPanel$ShowWarningsAction to BlazeProblemsViewPanel, storing the setting in ErrorTreeViewConfiguration.

hmemcpy:
Thank you for taking the patch, Brendan! I don't want to keep commenting on a now-closed issue, so I'd appreciate you input on #533 (since it is closely related to this area). Thanks!

sha1n:
Thanks @brendandouglas!
From your description of the implementation, I understand that a sync is required in order to update the view after the toggle state changes, is that the case? If so, then I wanted to give you some background regarding why I implemented it the way that I did. So, we had that kind of implementation in our fork for a few months and we got many requests from users to change the implementation, so that changing the state of the toggle will update the UI immediately, without re-syncing the project. In order to implement that, I needed to maintain an in memory model of the raw data, that's detached from the UI itself, so I can feed the UI with whatever I want without rebuilding the raw model.

Thanks again for your time and efforts!

brendandouglas:
No sync is required to update the view after toggling -- it just reuses the built-in functionality of ErrorViewStructure, which keeps all issues in memory but selectively displays them.

PR #532, opened 2019-03-07
Replace use of replaceAll with replace to fix Pattern.compile error

mattnathan:
String.replace is like String.replaceAll except it doesn't use regex, it
still replaces all the occurrences but this doesn't throw when
File.separator is `\` (an invalid regex pattern).

### DIFF
--- a/java/src/com/google/idea/blaze/java/sync/projectstructure/JavaSourceFolderProvider.java
+++ b/java/src/com/google/idea/blaze/java/sync/projectstructure/JavaSourceFolderProvider.java
@@ -105,7 +105,7 @@ public class JavaSourceFolderProvider implements SourceFolderProvider {
     if (Strings.isNullOrEmpty(relativePath)) {
       return parentPackagePrefix;
     }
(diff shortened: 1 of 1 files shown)


mattnathan:
I signed it!

mattnathan:
I signed it!

java/src/com/google/idea/blaze/java/sync/projectstructure/JavaSourceFolderProvider.java
```diff
@@ -105,7 +105,7 @@ private static String derivePackagePrefix(File file, SourceFolder parentFolder)
     if (Strings.isNullOrEmpty(relativePath)) {
       return parentPackagePrefix;
     }
-    relativePath = relativePath.replaceAll(File.separator, ".");
+    relativePath = relativePath.replace(File.separator, ".");
```
brendandouglas: Thanks! We're already replacing File.separatorChar with '/' in the call to FileUtil#toCanonicalPath above, so the correct fix here should be:
`relativePath.replace('/', '.')`

I'll make the change upstream.

PR #548, opened 2019-03-17
Respect user's package view settings

ceason:
This PR preserves the user's package settings in the project view (namely, the 'Flatten Packages' and 'Abbreviate Qualified Package Names' options).

### DIFF
--- a/base/src/com/google/idea/blaze/base/treeview/BlazeTreeStructureProvider.java
+++ b/base/src/com/google/idea/blaze/base/treeview/BlazeTreeStructureProvider.java
@@ -122,11 +122,17 @@ public class BlazeTreeStructureProvider implements TreeStructureProvider, DumbAw
 
       @Override
       public boolean isFlattenPackages() {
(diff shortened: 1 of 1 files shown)


brendandouglas:
This doesn't work properly as is (at least in some of our internal use cases) because it interacts poorly with our project view customization (we briefly did exactly this 1-2 years ago, and quickly reverted it).

Unfortunately, this feature is high risk and low value for most of our workflows, so isn't the kind of change we currently have the bandwidth to look into.

ceason:
@brendandouglas I'd consider poking around for a solution if I could reproduce, but I'm assuming the 'project view customization' you're referencing is another plugin and it's probably not public?

brendandouglas:
There's no additional customization, just this BlazeTreeStructureProvider. However we've found IntelliJ's handling of these things to be very fragile, so this customization has never worked particularly well.

Given that, and the history of breakages here, I'd prefer to err on the side of having a limited set of features which work, rather than trying to get all the options working.

PR #570, opened 2019-03-27
Update version of rules_scala

cushon:
See
bazelbuild/rules_scala@7bc18d0#diff-d68dd15f790bd2a815ab04d3d0a320d8
bazelbuild/bazel#7186


brendandouglas:
Could you make this change internally? This isn't the source of truth for this file, and we'd want to keep this version consistent with the rules_scala mirrored internally.

cushon:
This was merged in https://github.com/bazelbuild/intellij/commit/0605734cb08b80288c024475620c8bfc07e10874.

PR #575, opened 2019-03-27
Fix location of flag_hack 

### LINKED ISSUE #574: Project sync fails because @intellij_aspect//aspect:flag_hack is not found
I'm using the latest master with Bazel 0.24. Project sync fails with error:

```every rule of type java_library implicitly depends upon the target '@intellij_aspect//aspect:flag_hack', but this target could not be found because of: no such package '@intellij_aspect//aspect': BUILD file not found on package path.```

The correct location of the `flag_hack` rule is `@intellij_aspect//:flag_hack`

The following patch fixed the issue in my case:
```
diff --git a/aspect/intellij_info_impl.bzl b/aspect/intellij_info_impl.bzl
index 2837dba5..3fee2764 100644
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -905,7 +905,7 @@ def make_intellij_info_aspect(aspect_impl, semantics):
                 allow_files = True,
             ),
             "_flag_hack": attr.label(
-                default = "//aspect:flag_hack",
+                default = "//:flag_hack",
             ),
         },
         attr_aspects = attr_aspects,
```

ccascone:
Fixes #574

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -905,7 +905,7 @@ def make_intellij_info_aspect(aspect_impl, semantics):
                 allow_files = True,
             ),
             "_flag_hack": attr.label(
-                default = "//aspect:flag_hack",
+                default = "//:flag_hack",
             ),
         },
         attr_aspects = attr_aspects,


ccascone:
I signed it!

brendandouglas:
This has been fixed upstream.

PR #610, opened 2019-04-03
Update OLDEST_SUPPORTED_VERSION to 0.23.0+

promiseofcake:
Given the swap to using [PyInfo](https://github.com/bazelbuild/bazel/blob/ddfc430e54ac2b2d9eebb5adf672cbf1318175f2/src/main/java/com/google/devtools/build/lib/rules/python/PyInfo.java) in place of the legacy python provider (see: https://github.com/bazelbuild/bazel/issues/7298), this plugin no longer supports Bazel Versions < 0.23.

This updates the OLDEST_SUPPORTED_VERSION to `0.23.0`


brendandouglas:
Thanks for the patch, I'll apply this upstream.

PR #623, opened 2019-04-07
Clion 2019.1 fixes

miketzian:
Didn't build for me as-is, seems like a few things in the sdk have changed.

Changes here allowed me to build the plugin and load into CLion 2019.1. 

I "fixed" the IntelliJ urls as the ones there aren't in the repo anymore, though I'm not using bazel there to know if that works ok.

I saw the blurb about contribution difficulties, feel free to take if this is useful or close if not.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -20,8 +20,8 @@ http_archive(
 http_archive(
     name = "intellij_ce_2019_1",
     build_file = "@//intellij_platform_sdk:BUILD.idea",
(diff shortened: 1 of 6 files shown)


brendandouglas:
Thanks for the patch. We'll be adding support for CLion 2019.1 over the next month or so. There are several plugin API differences between 2018.3 and 2019.1 which need to be handled for both 2018.3 and 2019.1 (via our sdkcompat package) -- this along with testing is the vast majority of the work required.

PR #655, opened 2019-04-11
Move the pull request template up one level

jin:
The current location doesn't seem to be working, even though it's what the docs suggest. Let's try moving it to .github/, which follows Angular's repo (https://github.com/angular/angular/tree/master/.github)


brendandouglas:
The [docs](https://help.github.com/en/articles/creating-a-pull-request-template-for-your-repository) seem to suggest capitalization isn't necessary -- maybe make the file name lowercase?

I doubt it will make a difference, if the angular file works, but it looks nicer.

jin:
Made it lowercase.

PR #707, opened 2019-04-23
Don't test on Ubuntu 14.04

philwo:
Ubuntu 14.04 is about to be end-of-life and Bazel CI will stop supporting it shortly afterwards.

Context: https://groups.google.com/d/msg/bazel-dev/_D6XzfNkQQE/8TNKiNmsCAAJ

### DIFF
--- a/.bazelci/android-studio.yml
+++ b/.bazelci/android-studio.yml
@@ -1,6 +1,6 @@
 ---
 platforms:
-  ubuntu1404:
+  ubuntu1604:
     build_flags:
     - --define=ij_product=android-studio-beta
     build_targets:
@@ -10,7 +10,7 @@ platforms:
     - --test_output=errors
     test_targets:
     - //:aswb_tests
-  ubuntu1604:
+  ubuntu1804:
     build_flags:
     - --define=ij_product=android-studio-beta
     build_targets:
--- a/.bazelci/aspect.yml
+++ b/.bazelci/aspect.yml
@@ -1,13 +1,13 @@
 ---
 platforms:
-  ubuntu1404:
+  ubuntu1604:
     build_targets:
     - //aspect:aspect_files
     test_flags:
     - --test_output=errors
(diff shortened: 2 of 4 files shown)


brendandouglas:
Thanks Phil!

philwo:
@brendandouglas Thanks for the quick review!

I noticed that the tests for intellij are failing with a checksum mismatch:

```
ERROR: /workdir/intellij_platform_sdk/BUILD:374:1: no such package '@intellij_ce_2019_1//': java.io.IOException: Error downloading [https://www.jetbrains.com/intellij-repository/releases/com/jetbrains/intellij/idea/ideaIC/2019.1.1/ideaIC-2019.1.1.zip] to /home/bazel/.cache/bazel/_bazel_bazel/ec321eb2cc2d0f8f91b676b6d4c66c29/external/intellij_ce_2019_1/ideaIC-2019.1.1.zip: Checksum was d05abcfd50673c542558bcbd2c4853386e1e45b72759b8ed9bd374b87dc3a391 but wanted 600bfa4e334d7a03162a04deb9849be210cd34318efd76d83473171f221a8d27 and referenced by '//intellij_platform_sdk:application_info_jar'
 ```

Should we update the checksum?

brendandouglas:
I'm not sure why it's failing... I just changed the checksum from d05abcfd5 to 600bfa4e (cr/244664722), which matches what's currently available at that url. I also can't reproduce the error at head.

Are the CI tests accessing an old version of that zip rather than downloading it from scratch?

philwo:
The Linux workers don't have a persistent repository cache, so an old version of that zip shouldn't be the cause.

I reran the tasks on CI and now they worked. It's strange that both machines downloaded the file and got to the identical wrong checksum, which should rule out a network issue. Maybe a temporary issue on the server, serving a wrong file?

Anyway, it's fine now, so we should be good. :)

PR #722, opened 2019-04-25
Disable tests on ubuntu1804 that don't work with JDK11

philwo:
I have migrated ubuntu1804 to use OpenJDK 11, because that's the default JDK for that distribution.

During the migration I noticed that you have tests that fail with OpenJDK 11, so I recommend to
disable them until they're fixed.

For your reference, here's the log: https://buildkite.com/bazel/intellij-plugin/builds/2057#45f59142-dd60-491c-94eb-155b965fb0a4

### DIFF
--- a/.bazelci/android-studio.yml
+++ b/.bazelci/android-studio.yml
@@ -2,21 +2,11 @@
 platforms:
   ubuntu1604:
     build_flags:
-    - --define=ij_product=android-studio-beta
+      - --define=ij_product=android-studio-beta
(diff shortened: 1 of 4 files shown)


brendandouglas:
I can't reproduce these exceptions running on java11 internally, and it seems there's a single root cause for all those failures.

Either way, no objections to turning off the ubuntu1804 tests. I doubt we'll have to bandwidth to investigate this, but it can wait until we properly support java11 internally.

Issue #802, opened 2019-05-14
User-facing error message refers to "Blaze" rather than "Bazel".


strican:
Build failures show error messaging that says "Blaze failure building debug binary."

<img width="369" alt="Screen Shot 2019-05-14 at 12 34 19 PM" src="https://user-images.githubusercontent.com/104850/57716056-24e0a080-7646-11e9-9b68-01dd1b89188c.png">

Seems like this is one that was missed in #30.

jin:
This should probably use https://github.com/bazelbuild/intellij/blob/87777f152fe8d1ece8a98fb7ef2578ca9f6bb5c9/base/src/com/google/idea/blaze/base/bazel/BuildSystemProvider.java?

strican:
@jin - are you suggesting that that error shouldn't be called at all? Is your suggestion to get rid of that file then? Otherwise it sounds like a separate issue potentially.

sgowroji:
Hi there! We're doing a clean up of old issues and will be closing this one. Please reopen if you’d like to discuss anything further. We’ll respond as soon as we have the bandwidth/resources to do so.

PR #803, opened 2019-05-14
Update error message to use "Bazel"

strican:
Changing error message to use "Bazel" rather than "Blaze" (which is not the external name).

# Checklist

details.

Issue number: #802 

# Description of this change

Changes the error message from "Blaze" to "Bazel".


jin:
This will cause the internal name to be "Bazel", which is wrong. The correct approach here is to fix our import/export transformation, or derive the product name at compile time.

Issue #529, opened 2019-03-06
Bazel incompatible flags


laurentlb:
As mentioned in https://github.com/bazelbuild/intellij/issues/517, we run test the repository on Bazel CI. To avoid any breakage, we'd like to make sure issues are fixed in advance. Based on our latest CI run (https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/28#0ce981a8-efbe-481c-8105-72ee6098a934), migration is needed for the following flags:
  * --incompatible_depset_is_not_iterable
  * --incompatible_depset_union
  * --incompatible_disable_deprecated_attr_params
  * --incompatible_disable_legacy_cc_provider
  * --incompatible_disallow_legacy_py_provider
  * --incompatible_merge_genfiles_directory
  * --incompatible_new_actions_api
  * --incompatible_use_toolchain_providers_in_java_common

In particular, I plan to flip `--incompatible_merge_genfiles_directory` very soon.
Let me know if you need any help.

jvoung:
There's some update in the latest commits for `--incompatible_merge_genfiles_directory`:
https://github.com/bazelbuild/intellij/commit/9251dcac889928902cbdd6b3317ab02a01141819#diff-8825cb09855ccccac6d20911755d4284L45

`--incompatible_disable_legacy_cc_provider` should be https://github.com/bazelbuild/intellij/issues/527

When would `--incompatible_disallow_legacy_py_provider` be scheduled for a flip? It sounds like it was introduced in 0.23 (https://github.com/bazelbuild/bazel/issues/7298)

Some updates for `--incompatible_depset_union`, `--incompatible_new_actions_api` as well:
https://github.com/bazelbuild/intellij/commit/9251dcac889928902cbdd6b3317ab02a01141819#diff-ce84cc6cc21958f1b5bc7f65e7555c00L34

brandjon:
I sent #538 for `--incompatible_disallow_legacy_py_provider`. It was introduced in Bazel 0.23 and is scheduled to be flipped in 0.25.

laurentlb:
Thanks.

There's still some work to do. Migration is needed for `--incompatible_depset_is_not_iterable` and `--incompatible_disable_deprecated_attr_params`.

https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/88#c2fb4d3f-73ab-4881-ab43-85eaf8af2e7d

laurentlb:
[...]

PR #816, opened 2019-05-17
Add explicit depset to list conversions

laurentlb:
Implicit conversions will be forbidden with upcoming
`incompatible_depset_is_not_iterable` change in Bazel.

#529


laurentlb:
cc @jin

PR #977, opened 2019-07-11
 Bazel CI: Build all recursive targets instead of just the jar for higher coverage 

jin:
https://github.com/bazelbuild/bazel/issues/8571#issuecomment-510366705


jin:
Blocked https://github.com/bazelbuild/bazel/pull/8860 making into a patch release for 0.28. Once the patch release is out, CI should be green.

PR #1016, opened 2019-07-23
Expand test coverage on Bazel CI to also include the latest version of each product

jin:
This increases the coverage of the tests to both `-beta` and `-latest` variations of IJ products.

### DIFF
--- a/.bazelci/android-studio.yml
+++ b/.bazelci/android-studio.yml
@@ -1,12 +1,24 @@
 ---
-platforms:
-  ubuntu1604:
+tasks:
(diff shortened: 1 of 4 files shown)


brendandouglas:
We only ever build the bazel plugin against the beta version -- what's the use case for testing against the older 'stable' version?

('beta' and 'stable' and really internal terms not fully relevant to bazel users)

jin:
Discussed offline: there's currently no need to test against the stable version since we don't build against it. We can revisit this PR if we hit an use case that tests against stable would have helped.

PR #1053, opened 2019-08-05
Bazel fixes for --incompatible_disallow_empty_glob

laurentlb:
Globs that don't match anything may be removed, or made explicit with an
attribute. This is needed for compatiblity with Bazel 1.0.


laurentlb:
Should I send it as a Google CL instead? (cc @jin)

jin:
Importing.

PR #1096, opened 2019-08-17
Add bazel-invoking integration test for the bundled aspect

jin:
This test covers the end to end behavior of calling Bazel with the bundled aspect in a workspace. This should help us capture future incompatible changes to bzl files.


achew22:
@jin, it's really exciting to see this used in other repos. Thanks for reaching out and in there is anything I can do to help feel free to ping me internally or externally

PR #1126, opened 2019-08-27
Fix a bit of the language in the readme, to highlight -beta versions

cgruber:
The current readme doesn't really show how `-beta` and `-latest` defines work, and don't even note that `<IDE>-beta` exists as a possible `ij_product` definition.

Pure documentary change.


brendandouglas:
LGTM

Issue #1183, opened 2019-09-20
Scala Import fails to add Scala SDK


liucijus:
IntelliJ IDEA 2019.2.2 (Community Edition)
Bazel Plugin: source build

Sync fails attaching Scala SDK to module. Can be reproduced since 5f03bde0f02f5d1693a4567dc129230420a30372

Repo to reproduce: https://github.com/liucijus/bazel-scala-example

jin:
cc @iirina this change caused all Scala workspaces to stop resolving code. Do you know what may be the root cause here?

jin:
cc @brendandouglas this is pretty serious as it causes all Scala workspaces to not work. In the mean time while we find a mitigation (either the plugin or rules_scala), shall we revert the 5f03bde0f02f5d1693a4567dc129230420a30372 and cherry pick the revert into the next release?

jin:
Narrowed it down to this diff:

```
 def get_java_provider(target):
     """Find a provider exposing java compilation/outputs data."""
-    if hasattr(target, "java"):
-        return target.java
+    if JavaInfo in target:
+        return target[JavaInfo]
```

Background of this change: https://github.com/bazelbuild/bazel/issues/7598

iirina:
Thanks @jin for finding a fix!

ittaiz:
@jin @iirina was this due to a bug in `JavaInfo`? If so any idea what's the bug?  Ideally we'd like scala to use the regular JavaInfo and not a custom attribute

iirina:
My hunch is that the issue was caused by https://github.com/bazelbuild/bazel/issues/8916 which is now fixed but not released yet.

ittaiz:
@liucijus I tried reproducing the issue by manually editing the `intellij_info_impl.bzl` file to the order pre the fix but couldn't reproduce it. Can you reproduce locally? I was trying to see if the fixed Bazel version indeed fixes it but after 2-3 hours of failing to reproduce I stopped.

liucijus:
I can't run this fix without upgrading rules_scala version. But if I upgrade problem does not reproduce with older versions of bazel as well. For example, with `rules_scala_version = "b2273e7a90eac81132c9cdb8b2ca05fdbba74e46"` problem does not reproduce with Bazel 0.29.0.
If I don't upgrade Scala rules, and run with Bazel 1.0.0:
```
[...]

PR #1202, opened 2019-09-25
Prioritzie scala and kt providers before JavaInfo

jin:
While `scala_library` / other Scala rules have `JavaInfo` as well, their Scala-specific data lives in the `scala` provider. Proritizing `JavaInfo` over `scala` causes the plugin to assume that there are no outputs, hence causing all Scala workspaces to fail to resolve ("No Scala SDK added") https://github.com/bazelbuild/intellij/issues/1183

Instead, prioritize `kt` and `scala` providers over `JavaInfo`. 

This wasn't an issue previously because we c[hecked for the legacy "java" provider first](https://github.com/bazelbuild/intellij/commit/5f03bde0f02f5d1693a4567dc129230420a30372), which Scala targets do not have, and the conditional successfully fell through.

This fixes https://github.com/bazelbuild/intellij/issues/1183.

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -498,6 +498,10 @@ def collect_c_toolchain_info(target, ctx, semantics, ide_info, ide_info_file, ou
 
 def get_java_provider(target):
     """Find a provider exposing java compilation/outputs data."""
+
+    # Check for scala and kt providers before JavaInfo. e.g. scala targets have
+    # JavaInfo, but their data lives in the "scala" provider and not JavaInfo.
+    # See https://github.com/bazelbuild/intellij/pull/1202
     if hasattr(target, "scala"):
         return target.scala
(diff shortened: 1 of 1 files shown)


ittaiz:
Thanks! What I don’t understand is what is missing from the JavaInfo rules_scala generates? The scala attribute is more brittle than the canonical JavaInfo provider

ittaiz:
Or is it just about having some sort of “signal” to the plugin that this is scala core?

brendandouglas:
It's an actual difference between target[JavaInfo] and target.scala. We treat them identically otherwise.

The aspect otherwise doesn't distinguish between java and scala targets, and the plugin only uses the rule kind to distinguish between them.

ittaiz:
Wow. That’s bad. Sounds like on rules_scala side. I’ll run the aspect and try to print debug the differences between the two. Thanks!

Issue #1203, opened 2019-09-26
Two minor Windows issues related to file separator


gmishkin:
When importing a project, the contents of the `.bazelproject` file are stuffed into an editor control that can't handle carriage returns. This prevents importing projects on Windows.

When reading jdeps file it uses `ArtifactLocationDecoderImpl.outputArtifactFromExecRoot` to process the path to the file, there are forward slashes hardcoded in the method. This prevents external libraries from being loaded into the workspace

jin:
:+1: thanks for applying the fix.

sgowroji:
Hi there! We're doing a clean up of old issues and will be closing this one. Please reopen if you’d like to discuss anything further. We’ll respond as soon as we have the bandwidth/resources to do so.

PR #1204, opened 2019-09-26
Windows separator

gmishkin:
# Checklist

the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more details.

Issue number: #1203

# Description of this change

For the first thing, I just added a gitattributes file so that project view files keep only `\n` when you check out on Windows. The alternative would be to change the code to do the conversion before passing the file contents to the editor. Not doing this requires any repo with a project view file to do the same thing, but I didn't want to just start making changes to the contents of the file after it gets read in without talking to someone.

Edit: I think the above also used to work but I'm not as sure. I only found a very minor change in the way files are loaded and I don't have a good enough understanding to know whether that would make a difference in converting the line endings ahead of time.

The file separator thing is a relatively recent regression, this all used to work. I just did the smallest change I could think of, without doing anything like changing from `String` to `Path` everywhere (or some internal concept of a path), which might be better.


jin:
Can you please look at the failing tests?

gmishkin:
It looks like those tests don't run on Windows. Surprising cause it should resolve to `/` on the Ubuntu 16 image, and all the example paths in the tests have forward slashes too. I'll take a look when I'm back on my Mac tomorrow, obviously still want to make sure it works there too.

It seems like it might be going down a slightly different code path before it gets to the code I changed. The examples in the tests have `blaze-out` but all mine are `bazel-out` obviously. Is that a red herring? See: https://github.com/bazelbuild/intellij/blob/master/base/src/com/google/idea/blaze/base/model/RemoteOutputArtifacts.java#L158

brendandouglas:
Two issues here:
- the tests are failing because 'pathSeparatorChar' is ':' on linux (separatorChar is what's intended here)
- more importantly, ArtifactLocations are produced from a variety of contexts. In some situations, they're produced from normalized paths output by bazel, so will have '/' separators for both windows and linux.

To properly support Windows, we need to audit all sources of path strings, and treat raw / normalized sources differently.

gmishkin:
Ah I totally just made a typo on that between when I tested the build as a local modification (which worked) and when I actually forked the repo before preparing the PR, sorry.

Understood on your second point.

gmishkin:
Would it be feasible / desirable to use `Path` instead of `String` anywhere something that's logically a path is passed around? Would it be an appropriate class to use for both "physical" paths and "normalized" paths?

gmishkin:
Can you help find/write a test that would cover this case? It looks like they're all passing now.

brendandouglas:
Using 'path' is one option, but it still boils down to auditing all inputs (these strings are communicated directly from blaze, or generated from local paths, or read from protos).

Writing tests to cover all the edge cases is definitely something we'll do as part of supporting Windows, but it's not something we have the time for now (it also requires auditing all inputs).

jin:
@brendandouglas is this PR good for import, as-is?

brendandouglas:
The patch is harmless (and a no-op) for OS's with the '/' path separator,
but may actively make things worse in some situations for windows.

I'll have a closer look tomorrow, to see how often already-canonicalized
strings are passed through here.

gmishkin:
Would it be safe enough, if, the path starts with `bazel-out` (or blaze-out?) followed by a backslash, then flip all backslashes to forward slashes, then proceed as before? There would never be any files or directories below this level with embedded backslashes that would get flipped inadvertently, would there be?

jastice:
Hi, is there still interest in getting this change merged? @gmishkin ?

gmishkin:
@jastice I don't have access to a Windows PC anymore but it should be pretty straightforward for you to do if the answer to my last comment is "yes"

odisseus:
I wonder if this is still relevant.

jastice:
> I wonder if this is still relevant.

nobody replied, so guess not

Issue #1445, opened 2019-12-01
Sync is missing imports


szinn:
I have a project where I import a number of external jar files using maven_install and have build targets that depend on these jars via an @maven//:xxx

When I was editing the BUILD files, everything was fine and all of the maven-imported jar files were showing up as external libraries.  However, when I committed and then synced again, all of them would disappear.  If I added a space to a BUILD file, they would come back.

Debugging through this issue, it seems that the working set defined which targets would be examined for libraries.  However, this doesn't make sense and that all libraries should be considered.

The line of code I narrowed it down to is:

java/src/com/google/idea/blaze/java/sync/importer/BlazeJavaWorkspaceImporter.java#245
    if (workingSet == null || workingSet.isTargetInWorkingSet(target)) {

I would submit that it doesn't matter if the working set contains the target - the target is needed to determine the imported libraries.

Removing this if statement appears to make the external libraries show up all of the time, independent of the working set.

chaoren:
Can you share your .bazelproject file? This usually happens when you don't have the correct targets in the targets section. We have a setting to automatically add the files that you've modified (working set) to the list of targets to sync, but by default that list comes from your .bazelproject targets section.

szinn:
Here is an example repo:  https://github.com/szinn/Bazel-Kotlin

If you import it from Bazel in IntelliJ, the junit jar file won't be there if you have no changed / added files to the repo.  Add a space to test/BUILD and then resync the workspace and the junit external library shows up.  Remove the space which returns you to no altered files, resync, and junit disappears.

szinn:
It appears that the rules-kotlin plugin is marking all dependencies as UNUSED in the .jdeps file except for the standard Kotlin library ones that get marked as IMPLICIT.

PR #1446, opened 2019-12-01
Working set has no impact on external library computation

szinn:
# Checklist

details.

Issue number: `https://github.com/bazelbuild/intellij/issues/1445`

# Description of this change

Remove working set check when determining external library list.

### DIFF
--- a/java/src/com/google/idea/blaze/java/sync/importer/BlazeJavaWorkspaceImporter.java
+++ b/java/src/com/google/idea/blaze/java/sync/importer/BlazeJavaWorkspaceImporter.java
@@ -242,24 +242,22 @@ public final class BlazeJavaWorkspaceImporter {
     }
 
     // Add all deps if this target is in the current working set
-    if (workingSet == null || workingSet.isTargetInWorkingSet(target)) {
-      // Add self, so we pick up our own gen jars if in working set
-      workspaceBuilder.directDeps.add(targetKey);
-      for (Dependency dep : target.getDependencies()) {
-        if (dep.getDependencyType() != DependencyType.COMPILE_TIME) {
-          continue;
-        }
-        // forward deps from java proto_library aspect targets
-        TargetIdeInfo depTarget = targetMap.get(dep.getTargetKey());
-        if (depTarget != null
-            && JavaBlazeRules.getJavaProtoLibraryKinds().contains(depTarget.getKind())) {
-          workspaceBuilder.directDeps.addAll(
-              depTarget.getDependencies().stream()
-                  .map(Dependency::getTargetKey)
-                  .collect(Collectors.toList()));
-        } else {
-          workspaceBuilder.directDeps.add(dep.getTargetKey());
-        }
+    // Add self, so we pick up our own gen jars if in working set
+    workspaceBuilder.directDeps.add(targetKey);
+    for (Dependency dep : target.getDependencies()) {
+      if (dep.getDependencyType() != DependencyType.COMPILE_TIME) {
+        continue;
+      }
+      // forward deps from java proto_library aspect targets
+      TargetIdeInfo depTarget = targetMap.get(dep.getTargetKey());
+      if (depTarget != null
+          && JavaBlazeRules.getJavaProtoLibraryKinds().contains(depTarget.getKind())) {
+        workspaceBuilder.directDeps.addAll(
+            depTarget.getDependencies().stream()
+            .map(Dependency::getTargetKey)
+            .collect(Collectors.toList()));
+      } else {
+        workspaceBuilder.directDeps.add(dep.getTargetKey());
       }
     }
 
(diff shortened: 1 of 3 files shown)


szinn:
@googlebot I signed it!

szinn:
The test failure seems to be an infrastructure issue not related to the code change

java/src/com/google/idea/blaze/java/sync/importer/BlazeJavaWorkspaceImporter.java
```diff
@@ -242,24 +242,22 @@ private void addTargetAsSource(
     }
 
     // Add all deps if this target is in the current working set
-    if (workingSet == null || workingSet.isTargetInWorkingSet(target)) {
```
brendandouglas: This isn't a desirable change for our currently-supported workflows. We don't want to unconditionally add transitive dependencies to the project libraries.

If dependencies which are directly referenced by your project targets aren't being added as libraries during sync, that's a bug specific to your use case -- we'd need to see detailed steps to reproduce to understand why that's happening and how to fix it.

szinn: I will put together an example.  However, I would submit that the working set should have no impact at all on determining the referenced libraries that should be put into IntelliJ's external libraries list.

brendandouglas: We could achieve that by removing this block entirely (rather than removing the 'if' statement).

However we have this special-case 'working set' check to support in-progress modifications. If you're currently modifying files in a target, there's a legitimate use-case for unconditionally adding all direct deps without jdeps trimming.

For 'setlled' targets not currently being modified, we rely on the jdeps compilation output to tell us which deps are actually required to resolve the source code, and ignore the rest.

szinn: I believe the jdeps list was empty - My example was using Kotlin and the rules_kotlin environment.

I will try out both java and Kotlin in the next couple of days to see if there's a difference there.

szinn: Here is an example repo:  https://github.com/szinn/Bazel-Kotlin

If you import it from Bazel in IntelliJ, the junit jar file won't be there if you have no changed / added files to the repo.  Add a space to test/BUILD and then resync the workspace and the junit external library shows up.  Remove the space which returns you to no altered files, resync, and junit disappears.

brendandouglas: Thanks for the example. If there are no jdeps produced (or it's an empty file) that would explain what you're seeing.

We rely on jdeps to trim the deps, and assume that it's successfully produced. If the external kotlin rules aren't creating a jdeps file, we'd need to relax that assumption for external kotlin projects.

However this is a serious regression in the external kotlin rules, and something that really should be fixed on the rules side. Without jdeps-based library trimming, IDE performance will scale badly with the size of the kotlin project.

brendandouglas: I'm taking a closer look now -- doesn't seem to be a problem with the jdeps. I'll update here when it's tracked down.

szinn: I mistakenly thought the "reading 0 jdeps" meant there weren't any.  However, they are there they just hadn't changed.

In unpacking the jdeps files, it appears that the junit et al are marked as UNUSED rather than EXPLICIT, so that could be a possible issue?  (although I think it unlikely)

brendandouglas: That is the problem -- junit is marked as unused, so we ignore it (see the 'relevantDep' method in JdepsFileReader). That's definitely a bug in the external kotlin rules.

szinn: I will dig into that - just tracking down where it actually gets written.  I'll update here when I've found more.  Thanks @brendandouglas !

brendandouglas: It's not just junit, src/main/greet_lib.jar is also marked as unused, as are all deps other than the kotlin-stdlib ones.

szinn: It appears to be in the rules-kotlin package.

in JdepsParser.kt, there is code to parse the output of jdep execution to create the .jdeps file.

            val filename = Paths.get(classJar).fileName.toString()
            val jdepsParser = JdepsParser(filename, isImplicit)
            classPath.forEach { x -> jdepsParser.consumeJarLine(x, Deps.Dependency.Kind.UNUSED) }
            lines.forEach { jdepsParser.processLine(it) }

If it isn't one of the Kotlin standard libraries, it gets marked as UNUSED.  

Obviously, that would cause this issue.

szinn:
Found the issue in the rules_kotlin implementation - it was not invoking
jdeps correctly.

Submitted PR https://github.com/cgruber/rules_kotlin/pull/42/commits to
cgruber's mirror of rules_kotlin

PR #1625, opened 2020-02-06
Be compatible with --incompatible_load_proto_rules_from_bzl

Yannic:
# Checklist

details.

This PR is required to flip `--incompatible_load_proto_rules_from_bzl` in a future release of Bazel.
It's purely build-related and not a feature or bug-fix, so I did not open an issue (https://github.com/bazelbuild/intellij/issues/1512 was automatically created to notify about the upcoming breaking change but was closed without comment). I did, however, reach out to @jin (https://bazelbuild.slack.com/archives/CHSV3RSR0/p1580932545028100) to ask whether such a change would be accepted.

# Description of this change

This change adds load statements for `proto_library` to become
compatible with https://github.com/bazelbuild/bazel/issues/8922
Also, Protobuf is updated to v3.11.3 (which also has the necessary load
statements).

### DIFF
--- a/aspect/testing/rules/BUILD
+++ b/aspect/testing/rules/BUILD
@@ -2,6 +2,8 @@
 
 licenses(["notice"])  # Apache 2.0
 
+load("@rules_proto//proto:defs.bzl", "proto_library")
+
 proto_library(
     name = "intellij_aspect_test_fixture_proto",
     srcs = ["intellij_aspect_test_fixture.proto"],
@@ -27,7 +29,7 @@ java_binary(
         ":intellij_aspect_test_fixture_java_proto",
         "//aspect/testing:guava",
         "//proto:intellij_ide_info_java_proto",
-        "@com_google_protobuf_java//:protobuf_java",
+        "@com_google_protobuf//:protobuf_java",
     ],
 )
 
@@ -75,7 +77,7 @@ java_binary(
         ":fast_build_aspect_test_fixture_java_proto",
         "//aspect/testing:guava",
         "//proto:fast_build_info_java_proto",
-        "@com_google_protobuf_java//:protobuf_java",
+        "@com_google_protobuf//:protobuf_java",
     ],
 )
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -361,30 +361,24 @@ bazel_external_dependency_archive(
     },
 )
 
+# BEGIN-EXTERNAL-PROTO
 # LICENSE: The Apache Software License, Version 2.0
-# proto_library rules implicitly depend on @com_google_protobuf//:protoc
 http_archive(
-    name = "com_google_protobuf",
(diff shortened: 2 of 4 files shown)


jin:
FYI @brendandouglas, this is the change I discussed with you about.

aspect/testing/rules/BUILD
```diff
@@ -2,6 +2,8 @@
 
 licenses(["notice"])  # Apache 2.0
 
+load("@rules_proto//proto:defs.bzl", "proto_library")
```
chaoren: We should probably move these into the global preamble if possible.

aspect/testing/rules/BUILD
```diff
@@ -27,7 +29,7 @@ java_binary(
         ":intellij_aspect_test_fixture_java_proto",
         "//aspect/testing:guava",
         "//proto:intellij_ide_info_java_proto",
-        "@com_google_protobuf_java//:protobuf_java",
+        "@com_google_protobuf//:protobuf_java",
```
chaoren: This is probably added through coypbara transforms, I'll need to change this from our coypbara config.

aspect/testing/rules/BUILD
chaoren: Is this really necessary? I seem to able to still run everything without it.

Yannic: Right now, `rules_proto` is in the global preamble and automatically loaded if not defined, but the goal is to make these rules independent from Bazel releases, which isn't possible if they are in the preamble. Also, `rules_proto_{dependencies,toolchains}()` isn't called in the global preamble (and shouldn't be to avoid forcing users to download `rules_proto` if they don't use it).

Yannic: The incompatible flag has not been flipped yet, so everything works without these loads in Bazel 2.X. Once the flag is flipped (possibly in 3.0, https://github.com/bazelbuild/bazel/issues/8922), the loads are mandatory. See https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/404#6ec2b71a-4b6b-4e42-9c6d-0823aa81c642/1099-1101 which runs the CI with the flag flipped.

chaoren: > Right now, `rules_proto` is in the global preamble and automatically loaded if not defined

Sorry I meant the `tools/build_rules/prelude_bazel` file. I forgot the name of it when I wrote the comment, but it is prepended to every BUILD file, so that's why I called it the global preamble.

chaoren: Just tested it out with the flag. It seems to work fine. I'm gonna go with that instead of adding loads to every file.

chaoren:
The import commit is here: https://github.com/bazelbuild/intellij/pull/1627.
I'm gonna close this one for now.

Issue #1256, opened 2019-10-11
Alternative to jar cache for solving "red code"


ittaiz:
Bazel IJ plugin configures IJ libraries to point to outputs from bazel's execroot folder.
This is problematic since on every Bazel run (sync/build/test) the symlinks are recalculated and  *external binary* dependencies which aren't in the reach of the current execution are removed.
This is not true for outputs of source dependencies (same workspace or external workspaces).   

This causes "voodoo" red code/squigglies since suddenly IJ's libraries are misconfigured and pointing to a dangling symlink. The current recommendation is to use the jar-cache.  
The jar-cache is indeed much better than without it since sync will deterministically solve red code issues but red code happens more often (if you pull from upstream and have a partial view of the repo or if you build from command line) and additionally means double disk space.  

I'd like to suggest Bazel IJ plugin will traverse symlinks under `$execroot/$repo_name/external` and configure IJ libraries to where this symlinks points to since it's a transient location.  
This is preferable to the jar cache since it will eliminate red-code similarly to jar-cache but at a much smaller cost (only new nodes in the graph which are outside of the IJ sources will require a sync).

I'll try to publish a repro soon

jin:
https://github.com/bazelbuild/intellij/issues/884 may be related

ittaiz:
@brendandouglas wdyt about this issue? I think that for people who don't use remote execution this small change might make jar cache un-needed which is great (especially now that we've been bitten by #1336 )

brendandouglas:
Is the suggestion to resolve symlinks immediately after sync (since the symlinks, but not the files they point to, are transient)? If so, I don't see any problems with doing this if the jar cache is disabled.

I'd be curious to see some examples of the paths before and after (is this only a problem for external workspaces?).

Separately though, if you can reproduce jar cache bugs (or notice it happening more often) we'll be very interested in fixing them.

ittaiz:
The suggestion is indeed to resolve symlinks immediately after sync or even during sync (similarly to jar cache) since we know at that time those are there (again similarly to jar cache).
Can I send a PR? If so anything in particular you'd like to me pay attention to?

> I'd be curious to see some examples of the paths before and after (is this only a problem for external workspaces?).

One example:
```
/private/var/tmp/_bazel_ittaiz/c98f9042ba10a6c48fd4fdb03e36d7b0/execroot/io_bazel_rules_scala/external/io_bazel_rules_scala_guava/guava-21.0.jar 
-> 
[...]

PR #1817, opened 2020-05-07
Heuristically fix IntelliJ pointing to dangling symlinks

ianoc:
# Checklist

details.

Issue number: `https://github.com/bazelbuild/intellij/issues/1256`

# Description of this change

This is to solve the issue where jars that are seen to be external binaries to Bazel are reported to IntelliJ with symlink paths which are removed by bazel when targets not requiring those binaries are built. (More description of issue in issue).

We take the approach here following:
1) The jar is not in the cache
2) The jar was in an external path and the jar is not produced by bazel (no Bazel-out)
3) Remove the exec root part of the path. The external repositories under here symlink up one level under external. 

Expected result:

Much less red when syncing different targets in IntelliJ

### DIFF
--- a/java/src/com/google/idea/blaze/java/libraries/JarCache.java
+++ b/java/src/com/google/idea/blaze/java/libraries/JarCache.java
@@ -359,6 +359,14 @@ public class JarCache {
     return Optional.ofNullable(cacheState.get(cacheKey));
   }
 
+  private static File patchExternalFilePath(File maybeExternal) {
+    String externalString = maybeExternal.toString();
+    if(externalString.contains("/external/") && !externalString.contains("/bazel-out/")) {
+      return new File(externalString.replaceAll("/execroot.*external", "/external"));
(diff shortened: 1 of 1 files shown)


ianoc:
@googlebot I signed it!

ittaiz:
@brendandouglas do you think this can be merged (once @ianoc resolves conflicts of course)?
Would be very interesting for us to see how things behave when we turn off the jar cache

ianoc:
@ittaiz I think this is actually merged/and we can close this

https://github.com/bazelbuild/intellij/commit/81afc2cd1212828f33531d2705ce61172cfd237e

some aspect of how google merges in external changes I guess?

ittaiz:
Right...
Forgot about that even though you told me 🤦🏽‍♂️
Thanks!

alice-ks:
I'm closing this PR as #1819 was merged.

PR #2148, opened 2020-10-25
Allow for more flexibility when reporting test failures

avshenuk:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/2149

# Description of this change

When using certain assertion libraries, such as https://github.com/google/truth, the expected and actual value are not always present and the patterns defined in com.intellij.junit4.ExpectedPatterns are not covering all possible messages (actually I don't see them covering any of Google Truth messages).

But even without them, IMO the test should be considered a "failure" rather than an "error" when an assertion fails. This can be controlled by the means of the "error" XML element in the generated JUnit report making it much more flexible for the plugin users.

Additionally, a "system-err" element of the "testsuite" element is currently converted to a "system-out" output in the SM runner test console which seems to be a bug.

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlToTestEventsConverter.java
+++ b/base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlToTestEventsConverter.java
@@ -351,7 +351,7 @@ public class BlazeXmlToTestEventsConverter extends OutputToGeneralTestEventsConv
       processor.onTestOutput(new TestOutputEvent(displayName, test.sysOut, true));
     }
     if (test.sysErr != null) {
-      processor.onTestOutput(new TestOutputEvent(displayName, test.sysErr, true));
+      processor.onTestOutput(new TestOutputEvent(displayName, test.sysErr, false));
     }
 
     if (isIgnored(test)) {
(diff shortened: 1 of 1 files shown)


avshenuk:
@sgowroji And could you please tell how I can do it? :) It's been a long time, sorry...

avshenuk:
@sgowroji Done :)

base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlToTestEventsConverter.java
```diff
@@ -351,7 +351,7 @@ private static void processTestCase(
       processor.onTestOutput(new TestOutputEvent(displayName, test.sysOut, true));
     }
     if (test.sysErr != null) {
-      processor.onTestOutput(new TestOutputEvent(displayName, test.sysErr, true));
+      processor.onTestOutput(new TestOutputEvent(displayName, test.sysErr, false));
```
mai93, line 354: I'm not sure why we need to change this to `false`

avshenuk: Last argument to `TestOutputEvent` here is `true` for outputting to stdOut, and `false` for stdErr.   
So we were reporting both sysOut and sysErr elements as standard output but it seems logical to differentiate between those and let Intellij present them differently?

mai93 (approved):

PR #2168, opened 2020-10-30
Correct where stdout and stderr are read from

### LINKED ISSUE #2167: Stdout and stderr of tests not displayed
When running a test in IntelliJ, it is expected that the stdout and stderr of a specific test is displayed when clicking on that test in the runner UI. This does not happen with the Bazel IJ plugin.

The output is in the junit XML file that is generated, so the problem is within the plugin.

shs96c:
# Checklist

The junit.xml format outputs stdout and stderr in elements, not attributes. Fixing the encoding, fixes that issue.

Issue number: #2167 

Closes #2167

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlSchema.java
+++ b/base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlSchema.java
@@ -145,10 +145,10 @@ public class BlazeXmlSchema {
     @XmlAttribute public String result;
     @XmlAttribute public String time;
 
-    @XmlAttribute(name = "system-out")
+    @XmlElement(name = "system-out")
     String sysOut;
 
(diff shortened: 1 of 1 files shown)


base/src/com/google/idea/blaze/base/run/smrunner/BlazeXmlSchema.java
```diff
@@ -145,10 +145,10 @@ static TestSuite mergeSuites(List<TestSuite> suites) {
     @XmlAttribute public String result;
     @XmlAttribute public String time;
 
-    @XmlAttribute(name = "system-out")
```
alice-ks, line 148: Thanks for the fix! Please also add tests for std-out and std-err to BlazeXmlSchemaTest to avoid future regressions.

alice-ks (changes requested):

Issue #2595, opened 2021-05-07
Add tests against all supported IDE versions to the project CI


mai93:
To ensure that contributions targeted towards a certain IDE version do not break other supported versions, we intend to add tests against all the [supported IDE versions](https://github.com/bazelbuild/intellij/blob/master/intellij_platform_sdk/build_defs.bzl#L4) in the plugin CI pipeline.

PR #2596, opened 2021-05-07
Add tests against all supported IDE versions to the project CI

mai93:
# Checklist

Issue number: #2595

# Description of this change

This pull request adds tasks for testing against all the IDE versions supported internally and externally. I update the CI config files to include a task for testing against every supported version.

I also added a soft fail option for the IDE version that we currently work towards supporting (labeled as under development) so that its failure won't affect the overall check status. Testing against this version is expected to fail for most PRs until we fully support it but we want the whole presubmit test pipeline to succeed if the tests against older versions passes. This will allow us to merge the pull request, of course, after checking that the failed test log does not contain anything that should have passed and was broken by the PR.

### DIFF
--- a/.bazelci/android-studio.yml
+++ b/.bazelci/android-studio.yml
@@ -24,8 +24,8 @@ tasks:
       - --test_output=errors
     test_targets:
       - //:aswb_tests
-  Android-Studio-internal-canary:
-    name: Android Studio Internal Canary
+  Android-Studio-internal-under-dev:
+    name: Android Studio Internal Under Development
     platform: ubuntu1804
     build_flags:
       - --define=ij_product=android-studio-canary
@@ -36,6 +36,8 @@ tasks:
       - --test_output=errors
     test_targets:
       - //:aswb_tests
+    soft_fail:
+      - exit_status: 1
   Android-Studio-OSS-stable:
     name: Android Studio OSS Stable
     platform: ubuntu1804
--- a/.bazelci/android-studio.yml
+++ b/.bazelci/android-studio.yml
@@ -1,6 +1,20 @@
 ---
-platforms:
-  ubuntu1804:
+tasks:
+  Android-Studio-internal-stable:
+    name: Android Studio Internal Stable
+    platform: ubuntu1804
+    build_flags:
+      - --define=ij_product=android-studio-latest
+    build_targets:
+      - //aswb/...
+    test_flags:
+      - --define=ij_product=android-studio-latest
+      - --test_output=errors
+    test_targets:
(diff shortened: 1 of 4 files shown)


.bazelci/intellij-ue.yml
```diff
@@ -10,3 +24,54 @@ platforms:
       - --test_output=errors
     test_targets:
       - //:ijwb_ue_tests
+  IntelliJ-UE-internal-canary:
```
alice-ks, line 27: Here and also for CE: intellij-ce-canary and intellij-ue-canary are comparable to oss-under-dev. Hence, we should either not include them here or also use the soft-fail strategy. (Note: android-studio-canary is a bit different. See above.)

.bazelci/android-studio.yml
```diff
@@ -10,3 +24,54 @@ platforms:
       - --test_output=errors
     test_targets:
       - //:aswb_tests
+  Android-Studio-internal-canary:
```
alice-ks, line 27: This is typically targeting an actual Canary version of Android Studio and hence is different to the -canary version aliases for IntelliJ CE/UE. We could consider to use different task names to make this clear (e.g. under-dev for IntelliJ CE/UE).

What I'm currently not sure about is whether this canary version correctly builds for the Bazel plugin at all times as it's a preliminary version after all. We might want to also use a soft fail strategy here just in case.

alice-ks (changes requested):

.bazelci/android-studio.yml
mai93: I updated it.

.bazelci/intellij-ue.yml
mai93: I updated it.

.bazelci/android-studio.yml
```diff
@@ -10,3 +24,56 @@ platforms:
       - --test_output=errors
     test_targets:
       - //:aswb_tests
+  Android-Studio-internal-under-dev:
```
alice-ks, line 27: We might have misunderstood each other. I meant leaving this name as canary (as this version is officially Android Studio Canary) and to rename the tasks for IntelliJ UE+CE instead (e.g. from IntelliJ-UE-internal-canary to IntelliJ-UE-internal-under-dev). Can you please adjust the names?

alice-ks (changes requested):

.bazelci/android-studio.yml
mai93: Sorry about that :) I updated it now.

alice-ks (approved):

PR #2638, opened 2021-05-25
Import Android Studio 2020.3.1 Beta 2 (2020.3.2.17)

Bencodes:
Importing the latest 2020.3.1 beta release for Android Studio


alice-ks:
Thanks for originally sending the PR. We typically need to move Android Studio versions ahead internally first. By now, the plugin is fully on a released version of Android Studio 2020.3 and the pointers to 2021.1 and 2021.2 have also been updated regularly.

PR #2652, opened 2021-06-01
Add aspect tests for all supported IJ versions

mai93:
This pull request adds tasks for testing aspects against all the IDE versions supported internally and externally.


alice-ks (approved):
Thanks!

PR #2698, opened 2021-06-16
Add tests for CLion Internal Under Development version

mai93:
For the sake of completeness, add a test for CLion with the internal under development version defined by clion-under-dev.


alice-ks (approved):
Thanks!

PR #2805, opened 2021-07-28
Rename canary to under-dev in aspect CI

mai93:
Following the change proposed in #2794, this PR renames canary to under-dev in aspect tests as well.


AlexeyGy (approved):
LGTM, thanks for taking care of this

Issue #2338, opened 2021-02-02
TSConfig - Workspace path must be relative; cannot start with '/'


wardellbagby:
I've been getting an error in Intellij IDEA Ultimate 2020.2.4 when using the Bazel plugin and opening a TS file.

It seems like it's failing to resolve my TSConfig file, which means that things like IDE support for the TS paths don't exist.

I don't have a small sample repo but I do have the exact repo it's failing against, along with my `.bazelproject` file.

Things I've tried:

- Downgrading my Bazel plugin to `2021.01.06.0.1`
- Downgrading my Typescript version to `3.9.7`
- Deleting the `.ijwb` folder and reimporting.
- Making a dedicated `//:tsconfig` rule in my root `BUILD.bazel` file and referencing that in my `.bazelproject` instead of using the `//:tsconfig.json` label.
- Removing `javascript` from `additional_languages`
- A few `bazel clean` s thrown about.
- Creating a `tsconfig.json` that only extends the root `tsconfig` and referencing that in `.bazelproject`.

My setup:
- macOS Catalina - 10.15.7
- Bazelisk 1.7.4, fetching Bazel 3.7.2 for my project
- `rules_nodejs` 3.0.0
- Typescript 4.1 (this was previously 3.9.7 and I still saw the issue there)

Notes:
While I did try downgrading my IJWB plugin version, I've seen this across multiple, from approx. version `2020.10.19.0.3` onward, which is about when I started using the plugin.

Repo: https://github.com/wardellbagby/lyricistant/tree/c3a1e5b9735a6a68b0c231517e269db54d8d5e86
Stack trace:
```
java.lang.IllegalArgumentException: Invalid workspace path '/tsconfig.json': Workspace path must be relative; cannot start with '/': /tsconfig.json
	at com.google.idea.blaze.base.model.primitives.WorkspacePath.<init>(WorkspacePath.java:54)
	at com.google.idea.blaze.base.model.primitives.WorkspacePath.<init>(WorkspacePath.java:60)
	at com.google.idea.blaze.typescript.AddTsConfigNotificationProvider.getTsConfigDirectoryForFile(AddTsConfigNotificationProvider.java:177)
	at com.google.idea.blaze.typescript.AddTsConfigNotificationProvider.getTsConfigLabelForFile(AddTsConfigNotificationProvider.java:139)
[...]

PR #2829, opened 2021-08-03
Break out of while loop in AddTsConfigNotificationProvider.java if we are at the root of the workspace

mscolnick:
# Checklist

- [x] I have filed an issue about this change and discussed potential changes with the maintainers.
- [ ] I have received the approval from the maintainers to make this change.
- [x] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: #2338 

# Description of this change

Break out of while loop if we are at the root of workspace in `AddTsConfigNotificationProvider.java`

PR #2840, opened 2021-08-09
Conditionally prepend parentPath in WorkspacePath

CoopeyB:
# Checklist

details.

Issue number: #2338

# Description of this change
Prepending `parentPath` when the `parentPath` is the workspace root could end up with a path that starts with `/` (`BLAZE_COMPONENT_SEPERATOR`), which is incorrect and results in an invalid WorkspacePath. This change will only prepend the `parentPath` if the `parentPath` is not the workspace root.

Fixes: https://github.com/bazelbuild/intellij/issues/2338

Takes over from PR: https://github.com/bazelbuild/intellij/pull/2829

### DIFF
--- a/base/src/com/google/idea/blaze/base/model/primitives/WorkspacePath.java
+++ b/base/src/com/google/idea/blaze/base/model/primitives/WorkspacePath.java
@@ -57,7 +57,7 @@ public class WorkspacePath implements ProtoWrapper<String>, Serializable {
   }
 
   public WorkspacePath(WorkspacePath parentPath, String childPath) {
-    this((parentPath.isWorkspaceRoot() ? "" : (parentPath.relativePath() + BLAZE_COMPONENT_SEPARATOR)) + childPath);
+    this(parentPath.isWorkspaceRoot() ? childPath : parentPath.relativePath() + BLAZE_COMPONENT_SEPARATOR + childPath);
   }
 
   public static boolean isValid(String relativePath) {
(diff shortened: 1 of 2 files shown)


AlexeyGy:
Thanks for submitting the PR! The change, in and of itself, looks good to me, however, it would be great to also add a test for the new logic to [WorkspacePathTest](https://github.com/bazelbuild/intellij/blob/f894c76e3f492cf9322a9142b43fc0cf2526b39f/base/tests/unittests/com/google/idea/blaze/base/model/primitives/WorkspacePathTest.java). The constructor should be easy to test: you can initialize and check the #relativePath method output.

base/src/com/google/idea/blaze/base/model/primitives/WorkspacePath.java
```diff
@@ -57,7 +57,7 @@ public WorkspacePath(String relativePath) {
   }
 
   public WorkspacePath(WorkspacePath parentPath, String childPath) {
-    this(parentPath.relativePath() + BLAZE_COMPONENT_SEPARATOR + childPath);
+    this((parentPath.isWorkspaceRoot() ? "" : (parentPath.relativePath() + BLAZE_COMPONENT_SEPARATOR)) + childPath);
```
chaoren, line 60: I think this might be better for readability:

```
this(parentPath.isWorkspaceRoot() ? childPath : parentPath.relativePath() + BLAZE_COMPONENT_SEPARATOR + childPath);
```

CoopeyB:
Updated with a unit test to test the concatenation constructor. The test would fail before this change (since having a workspace path at the root would throw). Let me know if there's anything else that needs updating.

base/src/com/google/idea/blaze/base/model/primitives/WorkspacePath.java
CoopeyB: Agreed. Updated and pushed.

AlexeyGy:
Thanks, looks all good!

PR #2882, opened 2021-08-25
Remove reference to support matrix from README

alice-ks:
We removed the support matrix from the documentation and hence cannot reference it in the README anymore.


alice-ks:
Instead of completely dropping the matrix, I decided to replace it with another documentation entry. Hence, I'll drop this PR and will create another more suitable one.

Issue #2365, opened 2021-02-12
External libraries do not maintain package structure


decentralgabe:
![Screen Shot 2021-02-12 at 12 28 33 PM](https://user-images.githubusercontent.com/7622243/107820329-7a704280-6d2f-11eb-8b4f-423eb61507e4.png)

When using just go modules (not bazel) and you click on the “source” for a given import, you’re brought to the import package. With bazel, all the deps are single files — we lose the package structure of the deps. Viewing the deps in a given package is a useful feature. 

Is there a way to retain this functionality using bazel+goland?

blico:
@AlexeyGy would you be open to a PR on this issue? 

I have a change making it so external Go libraries are displayed based on their import paths (similar to how external Go deps are shown without using the Bazel plugin). For example:

![image](https://user-images.githubusercontent.com/4976073/134405945-40d52af7-8da0-403a-af19-17f9f81cf290.png)

AlexeyGy:
Hey blico,

this makes total sense from a user's perspective, especially for large projects withs lots of dependencies in individual go files. 

Since adapting the change would only touch the way that libraries are attached and made visible for IntelliJ in a slight way, I would not see any danger.

Happy to accept a PR.

Best regards,
Alexey

sgowroji:
Closing this issue w.r.t this [comment](https://github.com/bazelbuild/intellij/pull/3848#issuecomment-1249998647) as it is already rolled out. @decentralgabe could you please verify the latest release and let us know if you have any queries on the same.

Issue #1744, opened 2020-03-20
No structure in External Libraries -> Go Libraries


AntonIvanov87:
There is no structure in External Libraries -> Go Libraries
![Screenshot from 2020-03-20 15-43-38](https://user-images.githubusercontent.com/3289141/77174602-99d00280-6ac1-11ea-8e94-dbde4a248e6a.png)
I find it useful to "Select Opened File" in the Project view and then to explore neighboring files and packages.
But if you do that with a library file - it will be selected in the External Libraries -> Go Libraries, which does not have any structure, just a list of all files from all libraries.

mai93:
Fixed with https://github.com/bazelbuild/intellij/commit/593a8014a48194d2a23479bbd09d230efc2b5b5c

PR #2973, opened 2021-09-30
Golang: Display Go external libraries based on importpath structure

blico:
# Checklist

Issue number: #1744 #2365 

# Description of this change

Modifies the structure of the external "Go Libraries" project view to be based on importpath rather than a flat list of files. See my comment within #2365 for an example of how the new structure looks.

This is essentially achieved by extending Go's `additionalLibraryRootsProvider` to carry external dependency importpath mappings and then using that information from a new Go `treeStructureProvider` to create the importpath based directory structure. 

Due to my unfamiliarity with this project and the lack of pre-existing tests I struggled to write tests for this feature. If anyone more familiar with the codebase could provide some guidance that would be very helpful.

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/libraries/BlazeExternalLibraryProvider.java
+++ b/base/src/com/google/idea/blaze/base/sync/libraries/BlazeExternalLibraryProvider.java
@@ -34,7 +34,7 @@ public abstract class BlazeExternalLibraryProvider extends AdditionalLibraryRoot
       Project project, BlazeProjectData projectData);
 
   @Override
-  public final Collection<SyntheticLibrary> getAdditionalProjectLibraries(Project project) {
+  public Collection<SyntheticLibrary> getAdditionalProjectLibraries(Project project) {
     SyntheticLibrary library = ExternalLibraryManager.getInstance(project).getLibrary(getClass());
     return library != null ? ImmutableList.of(library) : ImmutableList.of();
   }
--- a/base/src/com/google/idea/blaze/base/sync/libraries/BlazeExternalSyntheticLibrary.java
+++ b/base/src/com/google/idea/blaze/base/sync/libraries/BlazeExternalSyntheticLibrary.java
@@ -36,7 +36,7 @@ import javax.swing.Icon;
  * A {@link SyntheticLibrary} pointing to a list of external files for a language. Only supports one
  * instance per value of presentableText.
  */
-public final class BlazeExternalSyntheticLibrary extends SyntheticLibrary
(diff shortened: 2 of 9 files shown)


AlexeyGy:
Thank you very much for this PR and sorry that it took me a while to review!

Unfortunately, I don't think that we can upstream the changes as-is. In particular, it worries me to let state leak out out of `ExternalLibraryManager` in the form of a `importpathToFilesMap`. We heavily rely on controlling the state of libraries and sources in our plugin. For example, this state controls what libraries are attached after the sync (this is why "Go external libraries" will initially disappear when you reopen a project). I fear that changing this may cause unforeseeable problems like libraries not disappearing after removing them from BUILD files.

TreeStructureprovider is the right way to go since changes there will only be cosmetic for the user. We can get the current project state via `BlazeProjectDataManager.getInstance(project).getProjectData` or even via using the `SyncCache`. 

Then, we only need to use the DFS that you've already drafted up to create the right structure. 

## Testing
In terms of tests, I think, the ideal point would be a unit test which checks if the tree structure is properly generated (i.e., a test for different scenarios of`createSyntheticDirectoryStructure`) and that covers several bad scenarios (e.g., what if we don't have all entries in importPath->Files map).

## How to move ahead
Let me move ahead with drafting up a PR based on this and then you can tell me if it would work for you (let's flip the script). I'd make it so you are co-author of the change on GitHub when it gets submitted. I'd probably also want to have this change behind a feature flag, just to be sure that users who don't want this can fall back on something.

blico:
That makes sense about not wanting to leak state and sounds like a good plan forward. Thank you @AlexeyGy !

blico:
@AlexeyGy @alice-ks friendly ping. Do you have any plans on taking this PR?

AlexeyGy:
> @AlexeyGy @alice-ks friendly ping. Do you have any plans on taking this PR?

Still on it and have a WIP PR but did not yet get to push it out mainly because I want this functionality to be well covered by unit tests.

loeffel-io:
Can someone please merge this thing - it's so annoying

linzhp:
@jastice FYI

jastice:
@AlexeyGy maybe you could publish your draft PR so that somoebody could pick it up?

AlexeyGy:
New PR is finally out:  #3833. Under internal review.

loeffel-io:
🎉

tpasternak:
@blico once the change is merged, can we close this PR?

blico:
Done!

linzhp:
@AlexeyGy In the future, it's a better to separate your change from @blico's in different commits, so the original author get some credits in the list of commits

PR #3052, opened 2021-11-08
Upgrade scala version to support latest Bazel

comius:
Fixes Downstream builds:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2241#708a7870-2e5a-4e59-bdf7-da99206e3fde


alice-ks:
Thanks for the fix! I'll start the process to import.

Issue #3001, opened 2021-10-12
IntelliJ 2021.3 Support


mai93:
The IntelliJ 2021.3 EAP was [released](https://www.jetbrains.com/idea/nextversion/#section=windows) a few weeks ago and we have made the necessary changes to enable working towards supporting this new version. You can now build the Bazel plugin for the EAP using: `intellij-oss-under-dev` or `intellij-ue-oss-under-dev` product-version aliases for CE and UE respectively.  For compatibility fixes, we encourage you to follow the [contributions guidelines](https://github.com/bazelbuild/intellij#contributions) described in the plugin documentation.

Now, we are looking for a volunteer from the community to organize the community work and contributions towards 2021.3 support. The volunteer should identify the breaking changes and drive the community work to fix them. Don't worry if you don't know how to do this. We can give you pointers if necessary. 

Please nominate yourself before the community meeting on October 28, 2021 at 8am Pacific Time. To do so, send us a quick note with your name and contact details to product@bazel.build. Also, please show up at the community meeting on Oct 28th.

mai93:
Please note that the Bazel plugin can now be built successfully for IntelliJ 2021.3. However, some plugin tests are failing for this version [[failing tests logs](https://buildkite.com/bazel/intellij-plugin/builds/14171#581812c8-56cf-45eb-a31e-6fac334cd710)].

We are still looking forward to your contributions to fix these tests so that we can be able to release a plugin version compatible with IntelliJ 2021.3 EAP.

maomaobug:
When will the fix be released?  IntelliJ 2021.3 was released out but the plugin's broken.

philwo:
Hi @maomaobug and others,

We're on it! We were on a pretty good track after @jwillebrands kindly contributed the required fixes for the failing tests in the Community Edition (https://github.com/bazelbuild/intellij/pull/3074), but unfortunately between the last EAP and the final 2021.3 release (https://github.com/bazelbuild/intellij/pull/3095), there were again some incompatible changes in the plug-in API that we couldn't prepare for, so more stuff broke. :(

Here's where you can track the current progress (click the latest job that ran on the `master` branch and check for the soft-failing "IntelliJ Under Development" jobs):
- https://buildkite.com/bazel/intellij-plugin
- https://buildkite.com/bazel/intellij-ue-plugin

[...]

PR #3074, opened 2021-11-23
Fix/2021.3 integration tests

jwillebrands:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/3001

# Description of this change

- Replaced annotator test code based on deprecated `AnnotationHolderImpl` with the `CodeInsightTestUtil.testAnnotator` utility method.
- Replaced deprecated use of `AnnotationHolder.createXAnnotation` with the new `AnnotationBuilder` based API.
- Run tests with `idea.force.use.core.classloader` set to `true`. Solution proposed by IntelliJ [here](https://plugins.jetbrains.com/docs/intellij/api-changes-list-2021.html), and a bit more context [here](https://github.com/JetBrains/gradle-intellij-plugin/blob/0873e0418a2bd01e8fa74c98811af1a567cdbced/src/main/kotlin/org/jetbrains/intellij/IntelliJPlugin.kt#L812).


testing/src/com/google/idea/testing/ServiceHelper.java
```diff
@@ -141,8 +142,6 @@
     ServiceContainerUtil.registerServiceInstance(componentManager, key, implementation);
     Disposer.register(
         parentDisposable,
-        () ->
-            ((MutablePicoContainer) componentManager.getPicoContainer())
-                .unregisterComponent(key.getName()));
+        () -> new ComponentManagerWrapper(componentManager).unregisterComponent(key.getName()));
```
alice-ks, line 145: Even though this can be implemented as a Wrapper, it's not directly necessary. It would also be possible to introduce a static method to `BaseSdkCompat` which takes `componentManager` and `key.getName()` as method parameters. Going for a method in `BaseSdkCompat` is more light-weight and hence preferred. Can you please switch to that approach?

testing/testcompat/v203/com/google/idea/sdkcompat/ComponentManagerWrapper.java
```diff
@@ -0,0 +1,19 @@
+package com.google.idea.sdkcompat;
+
+import org.picocontainer.ComponentAdapter;
+import org.picocontainer.MutablePicoContainer;
+
+import com.intellij.openapi.components.ComponentManager;
+
+// #api212
```
alice-ks, line 8: Here and for the other classes: Please add a short explanation on what to do when addressing the #api212 tag in the future.

alice-ks (changes requested):
Thank you for the adjustments for 2021.3 and figuring out how to fix the tests. The IntelliJ CE 2021.3 tests are green now. The IntelliJ UE 2021.3 tests still fail, but I guess that requires additional adjustments. Let's get this PR in to build on top of it.

I attached some comments. Please take a look. They should be quick to address. I'll do the next review round much more quickly (sorry that the initial review took a bit of time).

base/src/com/google/idea/blaze/base/lang/buildfile/validation/HighlightingAnnotator.java
```diff
@@ -22,38 +22,46 @@
...
 import com.google.idea.blaze.base.lang.buildfile.psi.ReferenceExpression;
 import com.intellij.lang.ASTNode;
-import com.intellij.lang.annotation.Annotation;
+import com.intellij.lang.annotation.HighlightSeverity;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.util.PsiTreeUtil;
 
-/** Additional syntax highlighting, based on parsed PSI elements. */
+/**
+ * Additional syntax highlighting, based on parsed PSI elements.
```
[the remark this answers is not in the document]
jwillebrands: Sorry, I couldn't get the formatter to play along. Thought I had omitted all of the mangled comments.

testing/testcompat/v212/com/google/idea/sdkcompat/BaseSdkTestCompat.java
```diff
@@ -15,10 +15,29 @@
...
 public final class BaseSdkTestCompat {
   private BaseSdkTestCompat() {}
+
+  /** #api212: inline into test cases */
+  public static List<Annotation> testAnnotator(Annotator annotator, PsiElement... elements) {
+    PsiFile file = elements[0].getContainingFile();
+    AnnotationHolderImpl holder = new AnnotationHolderImpl(new AnnotationSession(file));
+    for (PsiElement element : elements) {
+      holder.runAnnotatorWithContext(element, annotator);
+    }
```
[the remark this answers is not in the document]
jwillebrands: Yeah, that makes sense. Especially having also refactored the annotation creation to using the Builder.

testing/src/com/google/idea/testing/ServiceHelper.java
jwillebrands: Sure, will do. Just double-checking: I'm assuming this should go in `BaseSdkTestCompat` then. Is that correct?

alice-ks: Right. Since it's only used from test code, it should go into `BaseSdkTestCompat`. Thanks for double-checking.

jwillebrands:
Thanks for the review. No worries about taking a bit of time, I was still investigating the last failure. Agreed that we should move that to a separate PR.

I've addressed your comments. Should be good now. Thanks!

alice-ks (approved):

alice-ks:
Great. Thanks. I'll start the import of the changes.

PR #3155, opened 2022-01-10
Fix CI for 2021.3

fkorotkov:
# Checklist

details.

Issue number: #3001

# Description of this change

Attempt to fix CI for 2021.3. There were two issue with Scala part:

1. Bazel run configuration was specifying the whole Scala file as a source where the standard `ScalaApplicationConfigurationProducer` was specifying a more specific element which resulted in more priority.
2. Flaky tests when a source root was already removed on a test cleanup. Used a clearer way for cleanup which fixed the issue.

### DIFF
--- a/golang/BUILD
+++ b/golang/BUILD
@@ -102,4 +102,9 @@ intellij_integration_test_suite(
         "//third_party/go:go_for_tests",
         "@junit//jar",
     ],
+    env = {
+        # Disable runtime checks because GoGenericsLibraryRootsProvider causes them to trigger
+        # genericsStub folder containing builtin.go should be added to source roots, but it's not
+        "NO_FS_ROOTS_ACCESS_CHECK": "true"
+    },
 )
--- a/golang/BUILD
+++ b/golang/BUILD
@@ -105,7 +105,7 @@ intellij_integration_test_suite(
     env = {
         # Disable runtime checks because GoGenericsLibraryRootsProvider causes them to trigger
         # genericsStub folder containing builtin.go should be added to source roots, but it's not.
-        # This affects only 2021.3 version of the platform. Should be fixed in 2022.1 and potentially
+        # #api212: This affects only 2021.3 version of the platform. Should be fixed in 2022.1 and potentially
         # in a Go plugin release after 213.6461.48 version.
         "NO_FS_ROOTS_ACCESS_CHECK": "true"
     },
--- a/golang/BUILD
+++ b/golang/BUILD
@@ -104,7 +104,9 @@ intellij_integration_test_suite(
     ],
     env = {
         # Disable runtime checks because GoGenericsLibraryRootsProvider causes them to trigger
-        # genericsStub folder containing builtin.go should be added to source roots, but it's not
+        # genericsStub folder containing builtin.go should be added to source roots, but it's not.
+        # This affects only 2021.3 version of the platform. Should be fixed in 2022.1 and potentially
+        # in a Go plugin release after 213.6461.48 version.
         "NO_FS_ROOTS_ACCESS_CHECK": "true"
     },
 )
--- a/java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
+++ b/java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
@@ -74,6 +74,7 @@ public class MultipleJavaClassesTestContextProviderTest
         WorkspaceFileFinder.Provider.class, () -> file -> file.getPath().contains("test"));
 
     // required for IntelliJ to recognize annotations, JUnit version, etc.
+    // Adding into java source root so resolve scopes of all files in tests is the same.
     workspace.createPsiFile(
         new WorkspacePath("java/org/junit/runner/RunWith.java"),
         "package org.junit.runner;"
--- a/java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
+++ b/java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
@@ -53,19 +53,17 @@ import org.junit.runners.JUnit4;
 public class MultipleJavaClassesTestContextProviderTest
     extends BlazeRunConfigurationProducerTestCase {
 
-  private SourceFolder javaSourceRoot;
-
   @Before
   public final void addSourceFolder() {
-    // create a source root so that the package prefixes are correct.
-    VirtualFile pkgRoot = workspace.createDirectory(new WorkspacePath("java"));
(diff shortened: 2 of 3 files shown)


fkorotkov:
Seems there are golang issues as well. Converted PR to draft and will investigate.

golang/BUILD
```diff
@@ -102,4 +102,9 @@ intellij_integration_test_suite(
         "//third_party/go:go_for_tests",
         "@junit//jar",
     ],
+    env = {
+        # Disable runtime checks because GoGenericsLibraryRootsProvider causes them to trigger
+        # genericsStub folder containing builtin.go should be added to source roots, but it's not
+        "NO_FS_ROOTS_ACCESS_CHECK": "true"
```
philwo, line 108: I don't know enough about the IntelliJ plug-in build setup to judge the impact of this, have to defer to the assigned reviewers here.

The rest looks good to me, thank you!

philwo (approved):

fkorotkov:
Updated to 2021.3.1 and the latest Go plugin as well but it didn't fix the issue with `GoGenericsLibraryRootsProvider`.

java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
```diff
@@ -77,17 +75,17 @@ public final void addSourceFolder() {
 
     // required for IntelliJ to recognize annotations, JUnit version, etc.
     workspace.createPsiFile(
-        new WorkspacePath("org/junit/runner/RunWith.java"),
+        new WorkspacePath("java/org/junit/runner/RunWith.java"),
```
alice-ks, line 78: I was a bit puzzled that the "java/" prefix needed to be added here and below. After all, these JUnit files only need to be present in the PSI so that IntelliJ can resolve them when they are used in the example classes in the tests. Those JUnit files don't represent actual sources of the example project and don't need to be in the `java` source folder.

After seeing this, I dug a bit deeper and noticed that the `clearSourceFolders()` call below removes also the default source folder, meaning that later test runs won't start with the same configuration as the first executed test. Due to this, IntelliJ can't properly resolve the JUnit files anymore (which were apparently automatically placed in the default source folder).

I'm not convinced that calling `clearSourceFolders()` is the right thing to do. If we keep it, we'll at least need a comment for the JUnit classes explaining why they are explicitly added to the `java` source folder (compared to other tests which don't do this).

An alternative could be to not remove the source folder after tests as that won't remove the content anyway. Simply overwriting the `java` source folder at the beginning of each tests (if that folder still exists) seems to work.

golang/BUILD
alice-ks: Simply disabling this check looks a bit suspicious. Have you tried to adjust the test setup to add the necessary folder as source root?

alice-ks (commented):
Thanks a lot for fixing the tests. I added some comments. Please take a look.

java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
fkorotkov: I noticed that this test is very flaky and my investigation let to [these two lines of code](https://github.com/JetBrains/intellij-community/blob/2f53252d6cfcbb7f258fd4f9e151f662594b1797/java/execution/impl/src/com/intellij/execution/junit/JUnitUtil.java#L250-L251) where IntelliJ fails to find `org.junit.Test`. My theory is it's not finding the class because it's not in a source root and just added to the project.

I decided to go with this approach because it seems cleaner to me. There is no need to clean only java source root but rather to remove all roots. Anyways the JUnit classes are added before each test case.

On other note all other classes that add `org/junit/runner/*.java` files don't add source roots like `MultipleJavaClassesTestContextProviderTest`. I guess I can add a comment here on why this case is a special case.

golang/BUILD
fkorotkov: I agree it doesn't look clean. I tried a few things but without success. From decompiled code of `GoGenericsLibraryRootsProvider` I see that it's trying to access these folders in `getAdditionalProjectLibraries` method while the extension is loaded. So not much can be done on the configuration level. I've reported the issue to JetBrains folks though. They told me it's fixed but not released yet.

fkorotkov: Got news that it's fixed for 2022.1 and might be cherry picked to 2021.3.2.

alice-ks (changes requested):
Please add the comments as indicated and revert the changes I pointed out. Afterwards, this PR should be good to go.

golang/BUILD
fkorotkov: I asked @artspb in a private message. There is no YouTrack issue. Added the comment in 42ac35a916f10e04fa88622954ce54e3ffbf04e5

java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
fkorotkov: Added in 8769d07c2c3ac9662e413ad99f8954bdcd6793d6.

WORKSPACE
```diff
@@ -37,8 +37,8 @@ http_archive(
 http_archive(
     name = "intellij_ce_2021_3",
     build_file = "@//intellij_platform_sdk:BUILD.idea213",
-    sha256 = "7dea1bc8473974d8c49412b7c2c79cb4360025a9114b59145aef957afefe1111",
-    url = "https://www.jetbrains.com/intellij-repository/releases/com/jetbrains/intellij/idea/ideaIC/2021.3/ideaIC-2021.3.zip",
+    sha256 = "27cebd703daa14f4a87e19877adfa7d3b093b7f1f0478f2f5fd23b58e3582668",
+    url = "https://www.jetbrains.com/intellij-repository/releases/com/jetbrains/intellij/idea/ideaIC/2021.3.1/ideaIC-2021.3.1.zip",
```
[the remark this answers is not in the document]
fkorotkov: Got it. I also noticed that `ProjectIndexingHistoryImpl` is marked as an internal API. I'll take a look if it's possible to workaround the use of it so the plugin is compatible through out all the versions of IJ 2021.3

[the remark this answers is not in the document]
fkorotkov: @alice-ks it seems `ProjectIndexingHistoryImpl` is not used anywhere actually. I think it can safely be removed. See #3164

alice-ks (approved):

golang/BUILD
artspb: We fixed the problem within [GO-12171](https://youtrack.jetbrains.com/issue/GO-12171) by removing `genericsStub`. The fix was back-ported to 2021.3 and will appear in 213.6777.*. It's expected to be released by the end of the month. I apologize for the inconvenience caused.

PR #3165, opened 2022-01-12
Workaround symlink redness in Go

### LINKED ISSUE #2057: Intellij Ultimate + Go: Constant re-sync is necessary to keep external packages resolving
After re-sync all 3rd party dependencies are resolved fine in intellij, if I run some specific tests or targets in console they everything is highlighted in red immediately. Intellij reports and error `Build constraints exclude all Go files in '' `.

This makes the experience of using bazel in intellij pretty unusable, every time you run a command in bazel you have to re-sync. Cannot say when this broke, but it was working 6 months ago.

Intellij Ultimate 2019.3.5 (Same with 2020.1)
Latest Go Plugin
bazel 3.4.1

fkorotkov:
Similar to #1256 but for Go.

Fixes #2057

# Checklist

details.

Issue number: #2057

# Description of this change

A workaround similar to #1817.

### DIFF
--- a/golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
+++ b/golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
@@ -180,11 +180,15 @@ public class BlazeGoPackage extends GoPackage {
     if (maybeExternal == null) {
       return null;
     }
-    try {
-      return maybeExternal.toPath().toRealPath().toFile();
-    } catch (IOException e) {
-      return maybeExternal;
+    // do string manipulation instead of .toPath().toRealPath().toFile()
+    // because there might be a race condition and symlink won't be resolved at the time
+    String externalString = maybeExternal.toString();
+    if (externalString.contains("/external/")
+        && !externalString.contains("/bazel-out/")
+        && !externalString.contains("/blaze-out/")) {
+      return new File(externalString.replaceAll("/execroot.*/external/", "/external/"));
     }
+    return maybeExternal;
   }
 
   private static ImmutableSet<File> getSourceFiles(
--- a/golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
+++ b/golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
@@ -58,7 +58,6 @@ import com.intellij.psi.PsiFile;
 import com.intellij.psi.PsiManager;
 import com.intellij.util.Processor;
 import java.io.File;
-import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
--- a/golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
+++ b/golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
@@ -58,6 +58,7 @@ import com.intellij.psi.PsiFile;
 import com.intellij.psi.PsiManager;
 import com.intellij.util.Processor;
 import java.io.File;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
@@ -157,11 +158,33 @@ public class BlazeGoPackage extends GoPackage {
       if (target.getGoIdeInfo() == null) {
         continue;
       }
-      builder.putAll(
-          target.getKey().getLabel(),
-          getSourceFiles(target, project, projectData, libraryToTestMap));
+      ImmutableSet<File> sourceFiles =
+          getSourceFiles(target, project, projectData, libraryToTestMap).stream()
+              .map(BlazeGoPackage::toRealFile)
+              .filter(Objects::nonNull)
+              .collect(toImmutableSet());
+      builder.putAll(target.getKey().getLabel(), sourceFiles);
     }
     return builder.build();
+  }
+
+  /**
+   * Workaround for https://github.com/bazelbuild/intellij/issues/2057. External workspace symlinks
+   * can be changed externally by practically any bazel command. Such changes to symlinks will make
+   * IntelliJ red. This helper resolves such symlink to an actual location.
+   *
+   * @see com.google.idea.blaze.java.libraries.JarCache.patchExternalFilePath()
+   */
+  @Nullable
+  private static File toRealFile(@Nullable File maybeExternal) {
+    if (maybeExternal == null) {
+      return null;
+    }
+    try {
+      return maybeExternal.toPath().toRealPath().toFile();
+    } catch (IOException e) {
+      return maybeExternal;
+    }
   }
 
   private static ImmutableSet<File> getSourceFiles(


fkorotkov:
@alice-ks PTAL. I'm not sure how to go here. Not sure how to write a test and if it's needed.

Enrico2:
I've tested this change on 2021.3.1 and can confirm that I've not had to re-sync any file when before I normally had to do so after running any bazel command in any project other than the synced file's.

This is a HUGE QoL improvement.

Enrico2:
Fedor sent me a zip, but I can only assume you could check out the branch
and build it locally.

Enrico2:
FWIW this may need further testing.
1. my CPU usage spiked a lot due to IJ, but I only noticed after I was working for a while (headphones masked the fan noise)
2. I closed IJ and reverted the plugin version to the beta version from https://plugins.jetbrains.com/plugin/8609-bazel/versions/beta
3. I re-opened IJ and tried syncing some files, so far CPU usage is normal, but this is inconclusive because when I noticed the high CPU, it was after a while and I had >100 changes in my workspace.

MichaelSnowden:
I'm on GoLand 2021.2.5 and I tried this by building with

```
bazel build //ijwb:ijwb_bazel_zip --define=ij_product=intellij-ue-oss-stable
```

I then deleted my project's `.ijwb` directory, installed this zip from disk, and tested it by:
1. Syncing a file (all references resolved at this point, no red squigglies)
2. Running a Bazel command that usually triggers the underlying issue (requiring a re-sync)
3. I then observed that the references were unresolved again (red squigglies everywhere), and I needed to resync again

Maybe the command I used to build this is wrong? Not sure why this fix isn't working for me. Would appreciate your help @fkorotkov

golang/src/com/google/idea/blaze/golang/sync/BlazeGoAdditionalLibraryRootsProvider.java
```diff
@@ -71,6 +72,7 @@ protected String getLibraryName() {
     return BlazeGoPackage.getUncachedTargetToFileMap(project, projectData).values().stream()
         .filter(isExternal)
         .filter(f -> f.getName().endsWith(".go"))
+        .map(FileUtil::patchExternalFilePath)
```
alice-ks, line 75: As you noticed, patching the file paths is necessary in several places in the code. All of the ones you modified get their data from `BlazeGoPackage#getUncachedTargetToFileMap`. Is there a reason you didn't directly patch the file paths in that method?

When I checked for usages, I found at least one additional code path which depends on it. In the current state of this PR, that code path wouldn't get patched file paths, which could lead to overall inconsistencies and thus further bugs.

Doing the file patching in `BlazeGoPackage#getUncachedTargetToFileMap` would additionally have the advantage that the patching wouldn't need to be done on every use of the targetToFileMap as `BlazeGoPackage#getTargetToFileMap` would then cache the patched version.

golang/src/com/google/idea/blaze/golang/util/FileUtil.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.io.File;
+
+public class FileUtil {
+    // Workaround for https://github.com/bazelbuild/intellij/issues/2057
+    // External workspace symlinks can be changed externally by practically any
+    // bazel command. Such changes to symlinks will make IntelliJ red.
+    // This helper resolves such symlink to an actual location.
+    public static File patchExternalFilePath(File maybeExternal) {
+        String externalString = maybeExternal.toString();
```
alice-ks, line 11: Instead of `File#toString()` we should call `File#getPath()` here.

golang/src/com/google/idea/blaze/golang/util/FileUtil.java
```diff
@@ -0,0 +1,19 @@
+package com.google.idea.blaze.golang.util;
+
+import java.io.File;
+
+public class FileUtil {
+    // Workaround for https://github.com/bazelbuild/intellij/issues/2057
+    // External workspace symlinks can be changed externally by practically any
+    // bazel command. Such changes to symlinks will make IntelliJ red.
+    // This helper resolves such symlink to an actual location.
```
alice-ks, line 9: Please also leave a comment here that this solution uses the same approach as the one in `JarCache`. Otherwise, I would have expected to rather see symlink resolution using file APIs instead of the string manipulation this implementation does.

golang/src/com/google/idea/blaze/golang/util/FileUtil.java
```diff
@@ -0,0 +1,19 @@
+package com.google.idea.blaze.golang.util;
+
+import java.io.File;
+
+public class FileUtil {
```
alice-ks, line 5: If we end up using patchExternalFilePath() in only one location, please convert it to a private method.

If we need to keep this extra class, we 
1) need a private constructor so that the class can never be instantiated and
2) need Javadoc descriptions on the class and public method.

alice-ks (changes requested):
Thanks for the fix. I saw your question about adding tests. Ideally, if you can figure out how to write a test, it would be great to cover this new code. If you end up putting the path patching into `BlazeGoPackage` as I suggested, you could add a test to `BlazeGoPackageTest`, which already has some test setup.

After the adjustments, please also make sure to run google-java-format of the latest version.

alice-ks:
Changing the implementation in `JarCache` from `getFallbackFile()` to `OutputArtifactResolver#resolve` introduces a different behavior as the latter method has a different logic for remote outputs. That doesn't seem right to me.

In addition, touching `OutputArtifactResolver#resolve` will modify the behavior for many languages. If we get something wrong in there, the blast radius will be large. Hence, we'd need to apply careful testing. I don't think we should make this change right now.

To limit the scope, it would be better to move back to the original approach of this PR, which focused on fixing the resolution for Go. Please see my other comments for a good place to put the path patching logic.

fkorotkov:
I don't think `JarCache` behaviour is changed that much. If the [remote cache is not yet populated then the behavior will be the same](https://github.com/bazelbuild/intellij/blob/11acaac819346f74e930c47594f37d81e274efb1/base/src/com/google/idea/blaze/base/filecache/RemoteOutputsCache.java#L109). And if the remote cache is populated then it will just use the file with no extra computations. Maybe @brendandouglas can add some input/context around why `null` is returned in case of a remote artifact location?

I agree that the change is affecting a lot of places. As we see in these two examples IntelliJ doesn't play nicely with symlinks that change. There is third place in #3228 (external protocol buffers) I need this logic too. In that case it seems reasonable to introduce a new constrain in the plugin to always use real files instead of symlinks and put it in some abstraction layer that is easy to use. `OutputArtifactResolver` looks like an ideal place. If you are concerned about the funky string manipulation I've changed it in d9ad485aa3021a1c0f8b9e8527e8dbcf28f413fd. If you still think this is too risky of a change and instead I should put workarounds in all three places I'll do that.

alice-ks:
> If you still think this is too risky of a change and instead I should put workarounds in all three places I'll do that.

I think the change is still too risky and requires quite some time for ensuring that it doesn't break anything and that it doesn't introduce performance issues. I can't cover that part at the moment.

To continue, I'd suggest going back to the original version of the PR and following up on my review comments as suggested in them (e.g. moving the string manipulation to the indicated location). That seemed safe enough to not introduce major issues.

> If you are concerned about the funky string manipulation I've changed it in https://github.com/bazelbuild/intellij/commit/d9ad485aa3021a1c0f8b9e8527e8dbcf28f413fd.

I indeed stumbled a bit about the string manipulation and wondered why the symlinks weren't properly resolved in the code with file system operations. However, I have some concerns with that latest version.

First of all, Brendan mentioned performance concerns on https://github.com/bazelbuild/intellij/issues/1256 for some solutions and the current approach with resolving symlinks with file system operations frequently in `OutputArtifactResolver` sounds very much like one of these solutions.

We'd also need to ensure that the symlinks are only resolved once after sync and not on every resolution of an artifact reference. I haven't checked all the code paths which rely on `OutputArtifactResolver`, but we'd need to carefully do so. Since you're updating the JarCache to rely on  `OutputArtifactResolver`, I assume that this PR would regularly do the symlink resolution during the use of the plugin, which we definitely shouldn't do due to performance reasons. Furthermore, the symlinks could be invalid upon further execution of Bazel commands. Hence, we need to resolve them while they are still valid (-> directly after the Bazel sync). The string manipulation currently avoids this as the original file paths don't need to exist anymore.

> Maybe brendandouglas can add some input/context around why null is returned in case of a remote artifact location?

Brendan is not working on the team anymore.

Enrico2:
I can confirm that the performance issues I had were not related to this version of the plugin. I've been running with it for a few days without performance issues.

fkorotkov:
@alice-ks gentle ping. I reverted the changes to scope the to only Go.

golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
```diff
@@ -58,6 +58,7 @@
 import com.intellij.psi.PsiManager;
 import com.intellij.util.Processor;
 import java.io.File;
+import java.io.IOException;
```
alice-ks, line 61: This is an unused import. Please run google-java-format on future PRs. Having to run google-java-format myself means that I need an additional manual step during the import of such PRs.

alice-ks (approved):
Thanks for the adjustments!

alice-ks:
I triggered the import of this PR. As soon as it's submitted, this PR will be automatically closed.

Since we missed the cut-off date of the current release, it means that this PR won't get into the plugin version which will be pushed at the end of next week. The next release will be ~two weeks after the current one.

golang/src/com/google/idea/blaze/golang/resolve/BlazeGoPackage.java
fkorotkov: My bad, I installed the formatter and ran it but it seems it doesn't optimize imports. Fixed in 
1ea0f3a

Issue #3215, opened 2022-01-26
Disable "Expand Sync to Working Set" by default


fkorotkov:
I want to bring to attention one scenario that I experiences which confused me and my investigation into it.

**TLDR:** I propose to change the default for [`BlazeUserSettings#expandSyncToWorkingSet `](https://github.com/bazelbuild/intellij/blob/4c91d565062e152cace081bf3bfc8860008dc145/base/src/com/google/idea/blaze/base/settings/BlazeUserSettings.java#L72) from `true` to `false` because IMO it can confuse users more then help them.

In my case I have a huge mono repository where we practice fork driven approach (each engineer has a fork to work on and we submit PRs to the main repository). I have two origins in my Git repository: `origin` and `fork`. For example, I have a feature branch  `fork/fedor/feature-1` which was checkout from `fork/master`. I did some development, opened a PR, got some feedback after a few days and now I'm rebasing my feature branch on top of the latest changes from `origin/master`. Now when I run Sync from IntelliJ I see that the plugin **syncs every single target in the repository**.

After some investigation I found the root cause: because my `fork/master` hasn't been rebased as well this causes the "working set" to contain all files changed between `fork/fedor/feature-1` and the outdated `fork/master` instead of `origin/master`. [Here](https://github.com/bazelbuild/intellij/blob/4c91d565062e152cace081bf3bfc8860008dc145/base/src/com/google/idea/blaze/base/vcs/git/GitBlazeVcsHandler.java#L53-L66) is how `GitBlazeVcsHandler` calculates the working set of files. The troubling part is how the [SHA is calculated in `getUpstreamSha`](https://github.com/bazelbuild/intellij/blob/4c91d565062e152cace081bf3bfc8860008dc145/base/src/com/google/idea/blaze/base/vcs/git/GitBlazeVcsHandler.java#L123-L141).

I propose to change the default for [`BlazeUserSettings#expandSyncToWorkingSet`](https://github.com/bazelbuild/intellij/blob/4c91d565062e152cace081bf3bfc8860008dc145/base/src/com/google/idea/blaze/base/settings/BlazeUserSettings.java#L72) from `true` to `false`. It seems this automatic expansion of the target set based on changed files can bring more harm than benefits for use cases similar to mine.

Unfortunately I wasn't able to find more information about motivation for `expandSyncToWorkingSet` feature because this feature predates the publicly available Git history.

fkorotkov:
[...]

PR #3218, opened 2022-01-27
Invalidate Plugin Cache

fkorotkov:
# Checklist

details.

None

# Description of this change

As part of the IntelliJ's standard cache invalidation process (Files -> Invalidate Caches...).


alice-ks:
Is there a specific reason to invalidate the plugin's cache as part of the IDE's standard cache invalidation? Typically, our users need to invalidate the IDE's caches from time to time due to inconsistencies in the IDE's indices and it would be unfortunate if they also lost the plugin's caches with it.

Right now, you can avoid that the plugin uses its previously cached data by using "Non-Incrementally Sync Project with BUILD files".

fkorotkov:
I've disabled it by default in d5f4adb

Screen Shot 2022-02-09 at 7 41 30 AM

If I recall correctly I was in a situation when my project got in a weird state related to #3215, I wanted to import project from scratch and I ended up taking a look into the code in order to figure out that the caches are never invalidated.

Issue #675, opened 2019-04-16
Feature Request: Rust Support


ghost:
I've been working on integrating a Rust project with a large multi-language repository, and we're successfully building Rust binaries via `rules_rust` and `cargo raze`. However, we have not been able to get the projects working within CLion or Intellij IDEA, I think in part because the plugin does not yet support it.

Is Rust support something that the Intellij plugin authors would be open to? I am happy to dedicate time to making this come to fruition, if it is desired. Roughly speaking, what would it take to get this working?

petroseskinder:
@jin there seems to already be support for go, cpp, dart, javascript, python, scala, kotlin and java. 

What is the process for adding a new language? Barring scala which wix added support for, each of the existing languages are used internally within Google. I presume one concern is who would be on the hook for supporting the language support after the initial implementation

jin:
Currently, we don't have bandwidth or domain expertise to support new languages like Rust, and there isn't a process to add a new language yet. It's possible to work on a fork of this project to add Rust support, but until the details are ironed out, we cannot support this yet.

One of the things that's high on my list to do is to implement a pluggable framework so that rule owners / domain experts (e.g. rules_rust) can integrate with this plugin without having to extend the core IntelliJ plugin, similar to how Starlark has provided Bazel core with extensibility.

mattboardman:
Any updates on this? Will this be scoped out or prioritized in/for 2021?

vlad20012:
By the way, could someone link an example of a project that uses Bazel with Rust?

jelmansouri:
https://github.com/bazelbuild/rules_rust/tree/master/examples has a couple of samples.
I can build a repo that mimics the common components used in internal monorepos using bazel if it might help.
And here you can find an example of crates where cargo raze was used to generate build files:
https://github.com/bazelbuild/rules_rust/tree/master/bindgen/raze/remote

briansmith:
> Currently, we don't have bandwidth or domain expertise to support new languages like Rust, and there isn't a process to add a new language yet. It's possible to work on a fork of this project to add Rust support, but until the details are ironed out, we cannot support this yet.

Is forking this project still seen as the best way to getting Rust support added given the current state of the project?

Would you accept a PR to add direct Rust support now? Or...

[...]

PR #3222, opened 2022-01-31
Rust Support

alexjpwalker:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/675

# Description of this change

This integration depends on https://github.com/intellij-rust/intellij-rust/pull/8491.

The IntelliJ Rust plugin is built to work with Cargo projects, so to make it work with Bazel projects, we generate Cargo manifest files using Bazel.

This change generates a `Cargo.toml` (and a `RustIdeInfo` file) in the Bazel sandbox for every `rust_binary` and `rust_library` target in the project.

`rust_test` targets need to be handled a little differently. Normally, in Cargo, we specify test targets as part of the same `Cargo.toml` file that contains the library (or binary) that we're writing tests for. However, Aspects can't modify files generated by, say, `rust_library` while processing a `rust_test`. So instead, we force IntelliJ to detect the test target by writing a blank file to the Bazel sandbox with the same name and path as the `rust_test`'s source file.

An example project, and demo of the plugin's capabilities, is available here: https://github.com/vaticle/bazel-intellij-rust-example

https://user-images.githubusercontent.com/14253102/151391801-a7551fcb-0322-4ed5-adc1-f11784a3b106.mov


WORKSPACE
```diff
@@ -337,6 +337,22 @@ jvm_maven_import_external(
     server_urls = ["https://repo1.maven.org/maven2"],
 )
 
+jvm_maven_import_external(
```
alexjpwalker, line 340: `CargoTomlBuilder` uses `electronwill.nightconfig` to generate a TOML file

WORKSPACE
```diff
@@ -510,3 +526,12 @@ jvm_maven_import_external(
     licenses = ["notice"],  # Apache 2.0
     server_urls = ["https://repo1.maven.org/maven2"],
 )
+
+http_archive(
```
alexjpwalker, line 530: We need to import `rules_rust` to get access to `CrateInfo` exposed by each `rust_library` and `rust_binary` target.

base/src/com/google/idea/blaze/base/ideinfo/RustIdeInfo.java
```diff
@@ -0,0 +1,76 @@
...
+package com.google.idea.blaze.base.ideinfo;
+
+import com.google.common.collect.ImmutableList;
+import com.google.devtools.intellij.ideinfo.IntellijIdeInfo;
+
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+/** Ide info specific to Rust rules. */
+public final class RustIdeInfo implements ProtoWrapper<IntellijIdeInfo.RustIdeInfo> {
```
alexjpwalker, line 25: I've added `RustIdeInfo` because every other language has one, but I don't know if it's being used for anything.

base/src/com/google/idea/blaze/base/sync/SyncProjectTargetsHelper.java
```diff
@@ -132,11 +132,7 @@ private static boolean shouldDeriveSyncTargetsFromDirectories(ProjectViewSet vie
     }
     ImmutableList<TargetExpression> retained =
         SourceToTargetFilteringStrategy.filterTargets(targets).stream()
-            .filter(
-                t ->
-                    t.getKind() != null
-                        && t.getKind().getLanguageClasses().stream()
-                            .anyMatch(languageSettings::isLanguageActive))
+            .filter(t -> t.getKind() != null)
```
alexjpwalker, line 135: This change is clearly incorrect. What's the proper process for getting a new language marked as `isLanguageActive` in `bazelbuild/intellij`?

WORKSPACE
```diff
@@ -342,6 +342,22 @@ jvm_maven_import_external(
     server_urls = ["https://repo1.maven.org/maven2"],
 )
 
+jvm_maven_import_external(
+    name = "night_config_core",
+    artifact = "com.electronwill.night-config:core:3.6.5",
+    artifact_sha256 = "d47fd261e4c20c6f800a0f354644b5e83840347dde17d46537a9cb1c5a4207ba",
+    licenses = ["notice"],  # Apache 2.0
```
tpasternak, line 349: It's LGPL, not Apache

https://mvnrepository.com/artifact/com.electronwill.night-config/core/3.6.5

WORKSPACE
```diff
@@ -342,6 +342,22 @@ jvm_maven_import_external(
...
+    server_urls = ["https://repo1.maven.org/maven2"],
+)
+
+jvm_maven_import_external(
+    name = "night_config_toml",
+    artifact = "com.electronwill.night-config:toml:3.6.5",
+    artifact_sha256 = "12774bf07aeb893044fba8c38d80419f6f3adc0a28cc3b88325018dcf12e8bdf",
+    licenses = ["notice"],  # Apache 2.0
```
tpasternak, line 357: LGPL

https://mvnrepository.com/artifact/com.electronwill.night-config/toml/3.6.5

WORKSPACE
alexjpwalker: Updated to
```
licenses = ["notice"],  # Lesser General Public License 3.0
```

base/src/com/google/idea/blaze/base/sync/SyncProjectTargetsHelper.java
tpasternak: Afaik it is done in `BlazeSyncPlugin` implementations

tpasternak:
Hey @alexjpwalker, thank you for this PR! 

I'd like to know your opinion on one idea. If I understand correctly, it seems that the Rust Toml file generated here is not used anywhere inside IntelliJ Bazel Plugin code, it is utilized in IntelliJ Rust Plugin. I'm wondering if we could make aspects execution logic extensible, so it could be delegated to the Rust Plugin. This could simplify future development, as we wouldn't have to synchronize both sides. I don't know how to do it yet, just asking what you think about it.

alexjpwalker:
> it seems that the Rust Toml file generated here is not used anywhere inside IntelliJ Bazel Plugin code, it is utilized in IntelliJ Rust Plugin. I'm wondering if we could make aspects execution logic extensible, so it could be delegated to the Rust Plugin.

@tpasternak : Your understanding is correct! The Bazel plugin does not use the generated `Cargo.toml` files, these are used by the Rust plugin to identify project structure and the set of files to provide code analysis over.

The Rust plugin currently assumes that every Rust project is a Cargo project:
> Each Rust project in IDE consists of multiple `CargoProject`s.
>
> `CargoProject` corresponds to a `Cargo.toml` file inside a project source folder that is linked to the current IDE project via `attachCargoProject`.

(source: https://github.com/intellij-rust/intellij-rust/blob/e8b5860f6f31f520bd54b75d62fcbb917888e4e1/ARCHITECTURE.md#project-model)

Because of this assumption, one of the key goals of this PR has been to generate `Cargo.toml` files. This does introduce unwanted coupling between the Bazel and Rust plugins. If there was a way to avert that, we would be absolutely open to the idea - but we also don't currently know how that could be achieved.

alexjpwalker:
@tpasternak However - do correct me if I'm wrong:

- The only consumer of this plugin will be the Rust IntelliJ plugin;
- There is only one Rust IntelliJ plugin at the time of writing, and any future IntelliJ Rust work would ideally contribute to this plugin, rather than creating a new plugin;
- Even if a new Rust plugin was created, it would, almost certainly, assume the use of Cargo as the build system.

Therefore, creating a coupling between the two seems like it would be acceptable, as the two will always be coupled in practice.

Is there a specific reason why we need a solution that is not tied to IntelliJ Rust plugin(s)? And if so, what would that solution look like?

tpasternak:
cc @mai93

alexjpwalker:
It is true that the current approach is "needed because of the Rust plugin's expectation that we are working in a Cargo project". One unwanted consequence, for example, is that while Bazel downloads all the external Crate dependency sources to `bazel-{project_name}/external/{dependency_name}` (using Cargo!), we generate `Cargo.toml` and then the Rust plugin runs `cargo` again, re-downloading the dependency sources (if they are not already there) to the _system_ Cargo registry, then provides code analysis over them.

Whereas in Java, the dependency sources are installed into the Bazel sandbox and the IDE provides code analysis over these specific sources.

This was a trade-off made to avoid an unknown, potentially large amount of development work on the Rust plugin side. If we follow the same approach as in Java, this is how I (approximately) foresee it looking:

#### Bazel
- `intellij_info_aspect` generates IDE metadata for each `rust_library`, `rust_binary` (and perhaps `rust_test`). It specifies the exact paths to the sources of all of its dependencies. These paths may be in the project, or in `bazel-bin` (in the case of generated sources), or in `bazel-{project_name}` (in the case of external crates fetched via `crate_universe`/`cargo raze`/etc.)

#### Rust
- In addition to detecting Cargo projects (`Cargo.toml` files), also detect Bazel projects (IDE metadata files inside `bazel-bin`).
- Parse the Bazel IDE metadata to construct the project structure.

base/src/com/google/idea/blaze/base/sync/SyncProjectTargetsHelper.java
alexjpwalker: This turned out to be correct - by adding the `AlwaysPresentRustSyncPlugin` and the new `WorkspaceType` `RUST`, the behaviour is now in line with other optional languages:

- issues an IDE warning if the Rust plugin is not installed, which is clickable and prompts the user to install it;
- prompts the user to `enable Bazel plugin rust support` on detecting `rust_*` targets in the project, and only once that is enabled, flags those targets for syncing

As a result, the plugin now behaves properly and passes BuildKite's checks

aspect/intellij_info_impl.bzl
```diff
@@ -414,6 +415,157 @@ def collect_go_info(target, ctx, semantics, ide_info, ide_info_file, output_grou
     update_sync_output_groups(output_groups, "intellij-resolve-go", depset(generated))
     return True
 
+def _crate_info(ctx):
```
alexjpwalker, line 418: Each Rust target needs to know the following info about its dependencies:

- The crate name
- The crate version
- The crate's [features](https://doc.rust-lang.org/cargo/reference/features.html)

These three pieces of data are written directly into the Cargo manifest

WORKSPACE
alexjpwalker: Not that I could find. There appear to be a couple of MIT-licensed TOML libraries in the wild. I haven't tested any of them.

alexjpwalker:
The way that this PR currently functions is to generate `Cargo.toml` files and trick the Rust plugin into using them. However, this is not in line with the methods employed by other languages in the Bazel plugin. Moreover, I would expect the complexity of implementing a new language such as Rust _correctly_ (ie: not through hacks) to be very high, requiring considerable knowledge of IntelliJ extensions and the internals of the Bazel and Rust plugins which I don't have myself.

As such, we've changed course on Bazel + IntelliJ + Rust support and put together a script that runs a standalone Bazel aspect that generates metadata similar to `Cargo.toml`, then recursively scans `bazel-bin` and generates `Cargo.toml` files in the project directory to construct a Cargo project. We've retained some modifications to the Rust plugin, namely detecting the Rust toolchain and stdlib in `$(bazel info output_base)/external`, but we are able to proceed with this strategy with no code changes to the Bazel plugin at all, making this PR redundant.

I will create a copy of the branch this PR is based off and link to it here.

jastice:
Thank you, that seems like a better approach for a workaround until we get a more "proper" implementation. Does that mean we can close this PR?

alexjpwalker:
Yes, let's close this PR, @jastice . Thank you to everyone who's put in their time to review it.

andrew-otiv:
@alexjpwalker is the code for your new approach available?

alexjpwalker:
> @alexjpwalker is the code for your new approach available?

The code will need some massaging to be usable in the general case, but you can find our implementation at https://github.com/vaticle/dependencies/tree/3acb49da653059e93346e539666c9a8f1d88f65c/tool/ide .

anonrig:
Any plans on adding Rust support to the plugin soon?

tpasternak:
cc @jastice

jastice:
Hello @anonrig

Unfortunately Rust is not in the scope of the development work on this plugin. Wo do have plans to support it in the [new Bazel plugin](https://lp.jetbrains.com/new-bazel-plugin/) in the future, but will not be able to deliver it in the near-term.

PR #3229, opened 2022-02-02
Don't expand working set by default

### LINKED ISSUE #3215: Disable "Expand Sync to Working Set" by default

fkorotkov:
Fixes #3215

# Checklist

details.

Issue number: #3215

# Description of this change

See motivation in #3215

### DIFF
--- a/base/src/com/google/idea/blaze/base/settings/BlazeUserSettings.java
+++ b/base/src/com/google/idea/blaze/base/settings/BlazeUserSettings.java
@@ -69,7 +69,7 @@ public class BlazeUserSettings implements PersistentStateComponent<BlazeUserSett
   private FocusBehavior showBlazeConsoleOnRun = FocusBehavior.ALWAYS;
   private FocusBehavior showProblemsViewOnRun = FocusBehavior.NEVER;
   private boolean syncStatusPopupShown = false;
-  private boolean expandSyncToWorkingSet = true;
+  private boolean expandSyncToWorkingSet = false;
   private boolean showPerformanceWarnings = false;
   private boolean collapseProjectView = true;
   private boolean formatBuildFilesOnSave = true;


alice-ks:
Thanks for sending the PR. See https://github.com/bazelbuild/intellij/issues/3215#issuecomment-1032885692 for why we can't simply change the default.

PR #3487, opened 2022-04-17
fix clion can not found javax.annotation.Nullable;

kekxv:
fix clion-2022.1 can not found `javax.annotation.Nullable`;


alice-ks:
Thanks! The same is already covered by #3479, which was submitted by now. The CLion 2022.1 build is green with it.

PR #3553, opened 2022-04-29
Add Scala support to plugin workspace

liucijus:
At Wix we use Scala code to write internal plugins, but Scala language is not supported if `workspace_type` is `intellij_plugin`. This PR adds support to allow Scala language to be used in plugin workspaces.

Implementation is mostly copy of `IntellijPluginSyncPlugin` with added modifications to allow Scala to be used in the workspace. The test only adds `scala` to `additional_languages` project view section, but still uses Java code, because compiling code with Rules Scala would require complex workspace setup as the ruleset is not bundled with Bazel. I think this is enough to verify Scala language is supported.

I am open to make any changes, if needed.

### DIFF
--- a/plugin_dev/src/com/google/idea/blaze/plugin/sync/IntellijScalaPluginSyncPlugin.java
+++ b/plugin_dev/src/com/google/idea/blaze/plugin/sync/IntellijScalaPluginSyncPlugin.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2016 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.idea.blaze.plugin.sync;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.idea.blaze.base.model.BlazeProjectData;
+import com.google.idea.blaze.base.model.BlazeVersionData;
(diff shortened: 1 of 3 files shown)


liucijus:
👋🏻  @mai93 @jastice would appreciate a review on this one

plugin_dev/src/com/google/idea/blaze/plugin/sync/IntellijScalaPluginSyncPlugin.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    LanguageLevel javaLanguageLevel =
+        JavaLanguageLevelHelper.getJavaLanguageLevel(
+            projectViewSet, blazeProjectData);
+
+    // Leave the SDK, but set the language level
+    Transactions.submitWriteActionTransactionAndWait(
+        () -> {
+          LanguageLevelProjectExtension ext = LanguageLevelProjectExtension.getInstance(project);
```
tpasternak, line 98: Not 100% sure if this is required, but `BlazeGoSdkUpdater.java` modifies SDK in WriteAction context

tpasternak (commented):
Thank you. LGTM, apart from one issue mentioned in an inline comment.

I'm only wondering, does rules_scala support IntelliJ Plugins development out of the box?

plugin_dev/src/com/google/idea/blaze/plugin/sync/IntellijScalaPluginSyncPlugin.java
liucijus: @tpasternak I just did the same impl as in [IntellijPluginSyncPlugin](https://github.com/bazelbuild/intellij/blob/master/plugin_dev/src/com/google/idea/blaze/plugin/sync/IntellijPluginSyncPlugin.java#L96). Should I change it, or keep consistent with the existing code?

liucijus:
> I'm only wondering, does rules_scala support IntelliJ Plugins development out of the box?

It's just Java/Scala interop, nothing is specifically done to support Intellij plugins development.

plugin_dev/src/com/google/idea/blaze/plugin/sync/IntellijScalaPluginSyncPlugin.java
tpasternak: Ok, probably it's fine

jastice:
Is this okay to merge @tpasternak ?

jastice (approved):

Issue #3670, opened 2022-06-07
SourceFolderProviders can't set test folders


kamildoleglo:
If a test folder is not conforming to any regex in the project view file but is marked as a test source by the SourceFolderProvider, then the `setSourceFolderForLocation` method is called on the relevant SourceFolderProvider. However, if the provider then adds the folder as a test source anyway, the ContentEntryEditor removes it in the subsequent step, because the ContentEntry map has the (folder, isTestSource) pair as the unique key. This prevents the SourceFolderProviders from setting the test source folders altogether

mai93:
Hello @kamildoleglo can you provide us with repro steps to better understand the issue and the fix you are proposing in https://github.com/bazelbuild/intellij/pull/3671?

kamildoleglo:
Yeah, sorry, got a lot of work lately. I'll try to keep the discussion for this issue and the two PRs (https://github.com/bazelbuild/intellij/pull/3671/, https://github.com/bazelbuild/intellij/pull/3832/) here.
> Can you provide us with repro steps to better understand the issue and the fix you are proposing (...)

Just try to create any custom `SourceFolderProvider ` and mark directories as test ones. My use case is that I'm writing a Python language support, we use custom Python rules internally and the current solution (with `GenericFolderProvider`) doesn't work well. I thus created a separate plugin and plugged into some extension points provided by this Bazel IntelliJ plugin. I then created a custom `PythonSourceFolderProvider` that marks source and test directories. While the source directory marking works well, the test directory marking does not work as expected (explained a bit in my first comment). 

Upon further inspection and a small refactoring (see https://github.com/bazelbuild/intellij/pull/3832) it turns out that there is a way for me to make the test directories marking work (in most cases at least) by marking the folder's parent as a test source root, since the `contentEntry.removeSourceFolder(current)` is not called in the `if` branch for the parent. 
TBF the whole logic there is a bit unclear to me, I don't know why someone wrote it like that and it's not visible in the Git blame, since it has been copied from a private repo. I guess you could try to find that out if you have the access. 

[...]

PR #3671, opened 2022-06-07
Fix a bug with setting test folders in ContentEntryEditor

kamildoleglo:
# Checklist

details.

Issue number: [3670](https://github.com/bazelbuild/intellij/issues/3670)

# Description of this change
If a test folder is not conforming to any regex in the project view file but is marked as a test source by the SourceFolderProvider, then the `setSourceFolderForLocation` method is called on the relevant SourceFolderProvider. However, if the provider then adds the folder as a test source anyway, the ContentEntryEditor removes it in the subsequent step, because the ContentEntry map has the (folder, isTestSource) pair as the unique key. This prevents the SourceFolderProviders from setting the test source folders

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/projectstructure/ContentEntryEditor.java
+++ b/base/src/com/google/idea/blaze/base/sync/projectstructure/ContentEntryEditor.java
@@ -104,7 +104,7 @@ public class ContentEntryEditor {
     if (currentOrParent != null && isTest != currentOrParent.isTestSource()) {
       currentOrParent =
           provider.setSourceFolderForLocation(contentEntry, currentOrParent, file, isTest);
-      if (current != null) {
+      if (current != null && !current.equals(currentOrParent)) {
         contentEntry.removeSourceFolder(current);
       }
     }


tpasternak:
Hey Kamil, could you share an example showing the issue?

tpasternak:
> However, if the provider then adds the folder as a test source anyway

And when can this happen?

tpasternak:
```
    SourceFolder current = sourceFolders.get(new File(file.getPath()));
    SourceFolder next;
    if (current != null) {
      if (isTest != current.isTestSource()) {
        next = provider.setSourceFolderForLocation(contentEntry, current, file, isTest);
        contentEntry.removeSourceFolder(current);
      } else {
        next = current;
      }
    } else {
      if (parent != null && isTest != parent.isTestSource()) {
        next = provider.setSourceFolderForLocation(contentEntry, parent, file, isTest);
      } else {
        next = parent;
      }
    }
    for (Map.Entry<WorkspacePath, DirectoryStructure> child :
        directoryStructure.directories.entrySet()) {
      walkFileSystem(
          workspaceRoot,
          testConfig,
          excludedDirectories,
          contentEntry,
          provider,
          sourceFolders,
          next,
          child.getKey(),
          child.getValue());
    }
 ```

tpasternak:
So the intention is to let providers add test source folders even if they are not specified in `projectview`'s `test_sources`

1. Extract the code to a simpler form, without `currentOrParent` variable
2. Move `removeTestSources` to provider's interface, as a new method with default implementation, and call this in place of `contentEntry.removeSourceFolder`
3. Provide default implementation for the new method

This will keep the old behavior, but allow `provider` implementations to skip the `remove` step

tpasternak:
Or we may even do item 1 in a separate PR

kamildoleglo:
Currently done item 1. in https://github.com/bazelbuild/intellij/pull/3832 and responded with some additional context in https://github.com/bazelbuild/intellij/issues/3670

Issue #3685, opened 2022-06-14
Improve performance of Java package parser by making it a Bazel worker


keithl-stripe:
Here at Stripe, project sync performance is a big issue, and costs our engineers a lot of time. We found that project syncs often spend 10-30% of the overall build time just doing "package parsing," which is used when constructing IntelliJ project source roots.

The package parser tool is a Java application and the majority of the time spent executing it is in starting up the JVM and loading classes. It would benefit from being a Bazel worker.

Bencodes:
This is also something that we've seen here at Lyft in our Android mono repo. We also built out worker support for all of the JVM processes inside the aswb plugin and have seen a ton of performance improvements in sync times as a result.

mai93:
Closing as https://github.com/bazelbuild/intellij/pull/3686 is merged

Bencodes:
Created a separate issue to track worker support in the other two JVM based actions https://github.com/bazelbuild/intellij/issues/4860

PR #3686, opened 2022-06-14
Improve performance of Java package parser by making it a Bazel worker

keithl-stripe:
# Checklist

- [x] I have filed an issue about this change 
- [ ] and discussed potential changes with the maintainers.

details.

Issue number: #3685

# Description of this change
Makes the package parser tool (a Java CLI application) able to be run as a Bazel worker. Also modify the Starlark to enable worker mode.

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -694,7 +694,15 @@ def build_java_package_manifest(ctx, target, source_files, suffix):
         join_with = ":",
         map_each = _package_manifest_file_argument,
     )
-    args.use_param_file("@%s")
+
+    # Bazel has an option to put your command line args in a file, and then pass the name of that file as the only
+    # argument to your executable. The PackageParser supports taking args in this way, we can pass in an args file
+    # as "@filename".
+    # Bazel Persistent Workers take their input as a file that contains the argument that will be parsed and turned
+    # into a WorkRequest proto and read on stdin. It also wants an argument of the form "@filename". We can use the
+    # params file as an arg file.
+    # Thus if we always use a params file, we can support both persistent worker mode and local mode (regular) mode.
+    args.use_param_file("@%s", use_always = True)
     args.set_param_file_format("multiline")
 
     ctx.actions.run(
@@ -704,6 +712,10 @@ def build_java_package_manifest(ctx, target, source_files, suffix):
         arguments = [args],
         mnemonic = "JavaPackageManifest",
         progress_message = "Parsing java package strings for " + str(target.label),
+        execution_requirements = {
+            "supports-workers": "1",
+            "requires-worker-protocol": "proto",
+        },
     )
     return output
 
--- a/aspect/tools/src/com/google/idea/blaze/aspect/PackageParser.java
+++ b/aspect/tools/src/com/google/idea/blaze/aspect/PackageParser.java
@@ -22,6 +22,8 @@ import com.google.common.io.Files;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
+import com.google.devtools.build.lib.worker.WorkerProtocol.WorkResponse;
+import com.google.devtools.build.lib.worker.WorkerProtocol.WorkRequest;
 import com.google.devtools.intellij.aspect.Common.ArtifactLocation;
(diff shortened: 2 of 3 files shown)


aspect/intellij_info_impl.bzl
```diff
@@ -704,6 +712,10 @@ def build_java_package_manifest(ctx, target, source_files, suffix):
         arguments = [args],
         mnemonic = "JavaPackageManifest",
         progress_message = "Parsing java package strings for " + str(target.label),
+        execution_requirements = {
```
Bencodes, line 715: I think it would be great to also support multiplexed workers here.

Bencodes:
@mai93 can you take a look at this?

aspect/intellij_info_impl.bzl
jastice: @keithl-stripe is this something that makes sense from your perspective?

jastice (approved):
LGTM. Any problems you see @mai93 ?

Comments from @keithl-stripe in chat:

> AFAIK generally workers have some risks:
> could add to memory footprint of a build, e.g accidental state or uncollected garbage, or simply the fact that bazel keeps workers running between builds
> bugs could be hard to track down when they are the result of built up state
> 
> :one: doesn’t seem likely, as this worker doesn't do much, but it might be worth experimenting with Xmx values
> 
> :two: this worker doesn't carry any meaningful state beyond just the JVM’s own state

aspect/tools/src/com/google/idea/blaze/aspect/PackageParser.java
```diff
@@ -68,14 +70,61 @@ static PackageParserOptions parseArgs(String[] args) {
...
+      WorkResponse.Builder builder =
+          WorkResponse.newBuilder().setRequestId(workRequest.getRequestId());
+
+      try {
+        PackageParserOptions workRequestOptions = parseArgs(getWorkRequestArgs(workRequest));
+
+        parsePackagesAndWriteManifest(parser, workRequestOptions);
+
+        builder.build().writeDelimitedTo(System.out);
+      } catch (Exception e) {
```
tpasternak, line 107: I'm wondering whether we should catch all exceptions this way here. In case an exception is thrown, the loop will still run. Maybe we should allow continuation of the loop only when "parsePackagesAndWriteManifest" throws a parsing-specific exception, and for other errors break the loop. The BazelJavaBuilder is doing so

https://github.com/bazelbuild/bazel/blob/a4251eab6988d6cf4f5e35681fbe2c1b0abe48ef/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/BazelJavaBuilder.java#L45-L55

But I'm not sure, otherwise LGTM :+1:

mai93 (approved):

Bencodes:
@keithl-stripe There are a few other actions in this repo like `CreateAar` and `JarFilter` that would benefit greatly from having worker support. Can this worker implementation be made a bit more generic so that we can support those as well?

aspect/intellij_info_impl.bzl
keithl-stripe: Sure, but for now that's a "nice to have" - I believe it would simply improve overall performance, but should not be a blocker for this PR.

aspect/tools/src/com/google/idea/blaze/aspect/PackageParser.java
keithl-stripe: I wanted to reduce the risk that we crash hard for an issue that only affects the current action. I honestly don't know what is best here. If you'd like me to be more conservative that's fine - will update the PR.

tpasternak: Personally I would prefer to follow the same way as `BazelJavaBuilder`. For recoverable exceptions we can continue, but for unexpected ones System.exit would be more transparent. We could also print a message telling the user it is possible to run it without persistent workers.

mai93:
@estebandlc I would like to get your feedback on this PR.

keithl-stripe:
Sorry, we are moving away from using this plugin here at Stripe, so we are investing our effort in our homegrown IntelliJ plugin. Someone else will have to pick up this PR and get it landed if people think it would be useful. (It sped up our sync builds substantially!)

jastice:
Okay, then let's keep this open as a work item for maintainers or other contributors. Thank you for the effort so far!

mai93:
I'm ok with that, can you fix the conflict then merge it? It is already approved.

Issue #3726, opened 2022-06-25
Add .bazelrc to .gitignore


abhillman:
Allows devs to specify settings for all build targets. Example use-case:

```
$ cat .bazelrc
build --define=ij_product=intellij-ue-oss-stable
```

abhillman:
cc/ @mai93 @vsiva please take a look

abhillman:
It looks like the "better" way to do this is to add `build_flags` to `.bazelproject`

```
build_flags:
  --define=ij_product=intellij-ue-oss-stable
```

I would propose, instead of this approach, perhaps adding some documentation for how to import and run sources to the README.

mai93:
@abhillman I see that we mention this in the repo README [here](https://github.com/bazelbuild/intellij#compatibility-with-ide-versions) and instructions about using existing .bazelproject files are mentioned in the last line of [Building the plugin](https://github.com/bazelbuild/intellij#building-the-plugin) section
"You can import the project into IntelliJ (with the Bazel plugin) via importing the ijwb/ijwb.bazelproject file."

blorente:
@abhillman Could you let me know if the instructions pointed to by Mai (plus the dev instance instructions) are enough?

PR #3727, opened 2022-06-25
Add .bazelrc to .gitignore

abhillman:
# Checklist

Issue number: https://github.com/bazelbuild/intellij/issues/3726

# Description of this change

Allows devs to specify settings for all build targets without using a global `$HOME/.bazelrc` and without having to add things like `--define=<stuff>` to `bazel build //<path/to/target>`.

```
$ cat .bazelrc
build --define=ij_product=intellij-ue-oss-stable
```

Note that something like this is also useful for getting sync'ing to work; i.e.:

image

Results in this attribute error:

```
ERROR: /Users/ahillman/Development/intellij/aswb/BUILD:106:13: configurable attribute "srcs" in //aswb:aswb_lib doesn't match this configuration: define an intellij product version, e.g. --define=ij_product=intellij-latest
```

... which a `.bazelrc` with the above contents would rectify.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ bazel-*
 /.ijwb/
 /.aswb/
 /.clwb/
+.bazelrc


abhillman:
Closing to move discussion to the ticket for now

PR #3744, opened 2022-06-30
Allow running single Scalatest tests in Intellij.

### LINKED ISSUE #3425: Running a single test with ScalaTest
Clicking the gutter "Play" icon on a single test in a test suite written with ScalaTest runs the entire TestSuite.

srdo-humio:
ScalaTestContextProvider is responsible for figuring out which parameters
to pass to the Scalatest runner. It contains code capable of deciding which
test class to run, but not which individual test case in the class to run.

This commit replaces the Bazel plugin's own logic for this with a call to
the corresponding logic in the Scala plugin, which is capable of deciding
both the class and test name.

When no test name is detected (e.g. due to running a whole class), the
plugin will continue passing the class name in --test_filter as before.

When a test name is detected, the plugin will instead set --test_arg=-t
--test_arg="testName", and leave --test_filter out. This is because
rules_scala turns --test_filter into the -s Scalatest runner parameter,
which selects a full class for execution. Passing both -s and -t would
cause Scalatest to run both the full class, as well as the selected test.

Fixes #3425

### DIFF
--- a/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
+++ b/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
@@ -48,7 +48,7 @@ class ScalaTestContextProvider implements TestContextProvider {
   @Nullable
   @Override
   public RunConfigurationContext getTestContext(ConfigurationContext context) {
-    ClassAndTestNames classAndTestNames = getTestClass(context);
+    ClassAndTestNames classAndTestNames = getClassAndTestNames(context);
 
     if (classAndTestNames == null) {
       return null;
@@ -85,7 +85,7 @@ class ScalaTestContextProvider implements TestContextProvider {
   }
 
   @Nullable
-  private static ClassAndTestNames getTestClass(ConfigurationContext context) {
+  private static ClassAndTestNames getClassAndTestNames(ConfigurationContext context) {
     Location<?> location = context.getLocation();
     if (location == null) {
       return null;
@@ -106,13 +106,15 @@ class ScalaTestContextProvider implements TestContextProvider {
 
   public static class ScalatestTestSelectorFlagsModification implements TestContext.BlazeFlagsModification {
 
+    private static final String testClassSelectorFlag = BlazeFlags.TEST_ARG + "-s";
+
+    private static final String testNameSelectorFlag = BlazeFlags.TEST_ARG + "-t";
+
     private final List<String> testNames;
 
     private final List<String> testNameFlags;
 
-    private final String testClassSelectorFlag;
     private final String testClassFlag;
-    private final String testNameSelectorFlag;
 
     public ScalatestTestSelectorFlagsModification(String testClassFqn, @Nullable String testNames) {
       if (testNames != null) {
@@ -129,9 +131,7 @@ class ScalaTestContextProvider implements TestContextProvider {
         this.testNames = Collections.emptyList();
         this.testNameFlags = Collections.emptyList();
       }
-      testClassSelectorFlag = BlazeFlags.TEST_ARG + "-s";
       testClassFlag = BlazeFlags.TEST_ARG + testClassFqn;
-      testNameSelectorFlag = BlazeFlags.TEST_ARG + "-t";
     }
 
     private boolean flagsMatchSettings(List<String> flags) {


scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -76,27 +104,31 @@ private static ScClass getTestClass(ConfigurationContext context) {
       // handled by a different producer
       return null;
     }
-    return getTestClass(location);
+    ScalaTestConfigurationProducer scalaTestConfigurationProducer = new ScalaTestConfigurationProducer();
```
tpasternak, line 107: We could use ScalaTestConfigurationProducer.instance() here

tpasternak (commented):
Hey @srdo-humio . Thank you for the contribution. I think it would be great to add a test for it in BlazeScalaTestClassConfigurationProducerTest.java

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: :+1: Addressed

srdo-humio:
@tpasternak Yes, I attempted this but gave up. The test infrastructure in that test fakes too much. The current test in that class defines a couple of interfaces that simulate the ones in Scalatest closely enough to work for the class lookup code. 

The find-the-test code in the Scala plugin relies on an "org.scalatest.Finders" annotation on the test interface, such as the one on `AnyFunSuite` `@Finders(Array("org.scalatest.finders.FunSuiteFinder"))`, which then points to a class in Scalatest that contains logic on how to find the relevant class.

I think in order to add a test for this, I'd have to define a workspace that actually depends on Scalatest, I can't just fake a project with some dummy interfaces. I didn't spot any good tooling to do this, or any tests that did something similar, so I decided to drop it.

What do you think?

srdo-humio:
My actual problem is that in order to figure out how to make a test work here, I need to debug into code from the Scala IJ plugin, specifically `ScalaTestConfigurationProducer`, but when I open this project in IJ, sources are not attached for the `scalaCommunity` jar. 

If you can tell me how to fix that, I can probably get some kind of test working.

tpasternak:
I'm not sure, it would be good to have integration tests for this code, but at the moment it would be indeed hard to do it due to limitations of the test framework you have found. Let me ask someone about it.

----

On the other hand, maybe it would be good not to split code into "class run" and "single test run". In this case, we would always use `addBlazeFlagsModification`. This would mean, we would cover whole configuration builder logic by in the test class test case. 

----

In the meantime, I noticed there's a bug here. The Run Configuration executes all tests of the specified name, even if they are in separate suites. For example:

```
import org.scalatest.flatspec.AnyFlatSpec
import org.scalatest.matchers.must.Matchers

class ExampleTest extends AnyFlatSpec with Matchers {
  "Test1" should "pass" in {
    2 must be(2)
  }

  "Test2" should "pass" in {
    2 mustBe 2
  }
}

class ExampleTest22 extends AnyFlatSpec with Matchers {
  "Test1" should "pass" in {

  }
}

```

srdo-humio:
> In the meantime, I noticed there's a bug here. The Run Configuration executes all tests of the specified name, even if they are in separate suites. For example:

Yes, I noticed this and initially thought it was an issue with ScalaTest, so I just left it in. I tried passing both `--test_filter` and `-t`, and the result was that both the specified test, and all tests in the specified class were run.

It turns out it's actually an oversight in `rules_scala`. `rules_scala` turns `--test_filter=className` into an `-s className` flag for ScalaTest's runner, putting it at the end of whichever other arguments are specified.

ScalaTest cares about the ordering of these flags. "-s className -t testName" means "run the test in className with the specified name". "-t testName -s className" means "run all tests in className and also run tests that match the specified name".

I'll submit an issue to `rules_scala` and see how they feel about moving `-s` to the front of the parameter list instead of the back. I don't think this behavior is on purpose. If we can make the change there, this plugin can switch to passing both `--test_filter` and `-t` at some point. For now I've made it use `--test_filter` only when running a class, while running a single test specifies `-s` and `-t` via `--test_arg` flags instead.

srdo-humio:
> On the other hand, maybe it would be good not to split code into "class run" and "single test run". In this case, we would always use addBlazeFlagsModification. This would mean, we would cover whole configuration builder logic by in the test class test case.

I gave making a test for this another shot, and it turns out that the Scala plugin's code for figuring out which single test is being pointed at depends on loading the test class via a classloader, and then reading the annotation I mentioned above. I'm pretty sure the test in BlazeScalaTestClassConfigurationProducerTest doesn't even compile the test class, so we'd need to figure out another way to test it.

srdo-humio:
@tpasternak I've updated the code to always use `addBlazeFlagsModification`. I don't really see the point of setting `--test_filter` over setting the `-s` flag directly, so I've merged the two code paths so the flag modification always sets `-s`, and only sets `-t` if a test case is selected.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -76,32 +92,49 @@ private static ScClass getTestClass(ConfigurationContext context) {
...
-      return testClass;
-    }
-    return null;
-  }
+  static TestContext.BlazeFlagsModification setScalatestTestSelectorArguments(ClassAndTestName classAndTestName) {
+    String testClassSelectorFlag = BlazeFlags.TEST_ARG + "-s";
+    String testClassFlag = BlazeFlags.TEST_ARG + classAndTestName.testClass.getQualifiedName();
+    String testNameSelectorFlag = BlazeFlags.TEST_ARG + "-t";
+    // Scalatest names can contain spaces, so the name needs to be quoted
+    String testNameFlag = BlazeFlags.TEST_ARG + "\"" + classAndTestName.testName + "\"";
```
tpasternak, line 106: We should also escape some characters here, at least `"` and `\`

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -76,32 +92,49 @@ private static ScClass getTestClass(ConfigurationContext context) {
...
 
-  private static String getTestFilter(PsiClass testClass) {
-    // TODO: may need to append '#' if implementation changes.
-    // https://github.com/bazelbuild/rules_scala/pull/216
-    return testClass.getQualifiedName();
+      @Override
+      public boolean matchesConfigState(RunConfigurationFlagsState state) {
+        List<String> rawFlags = state.getRawFlags();
+        boolean testClassFlagsSet = rawFlags.contains(testClassSelectorFlag) && rawFlags.contains(testClassFlag);
+        boolean relevantNameFlagsSet = !hasTestName() || (rawFlags.contains(testNameSelectorFlag) && rawFlags.contains(testNameFlag));
```
tpasternak, line 135: Personally I'd prefer to check these flags more strictly. So to check whether -s is directly followed by suite name and -t by test case name. Theoretically a test could have the same name as a suite and this code would fail. It does not happen now only because we surround test name with double quotes.

Maybe we could extract and unit test this logic?

tpasternak (commented):
So in general it looks good to me, apart from the two minor issues I made just now.

But as @srdo-humio pointed in one of the previous messages, we can't write good integration tests for it due to limitations of the test framework. Is this OK to merge it then? @mai93 @jastice

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: I understand the need for `"`, but why is escaping `\` needed?

tpasternak: That's tricky, but this fails when you try to run this
```
class Test2 extends AnyFunSuite {
  test("Test\\") {

  }
}
```

In order to fix it I had to manually edit configuration and set it to 
```
--test_arg=-s
--test_arg=foo.Test2
--test_arg=-t
--test_arg=Test\
```

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
tpasternak: Oh, and it's also wrong when you first try to run only Test1, and the whole Suite afterwards
```
class Suite extends AnyFunSuite {
  test("Test1") {

  }
  test("Test2") {

  }
}
```

srdo-humio: Did this. Let me know if you have a better place in mind for the flag modifier class, I wasn't sure where to put it so for now it's just a static inner class.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: :+1: Tried to address this, and added a test for it.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: > Oh, and it's also wrong when you first try to run only Test1, and the whole Suite afterwards

Good catch, it wasn't clear to me why matchesConfigState even needed to exist, but it makes sense if it's caching old flags.

So presumably in `modifyFlags`, I'd need to remove any previous occurrences of the `-s` and `-t` flags?

tpasternak: I don't think so. If the flags are already there, something must have gone really bad, so I'd consider it a non-recoverable error and just throw, but I'm not sure what would happen then. Other solution is to indeed replace the flags, but log error with IntelliJ logging mechanism.

srdo-humio: I'd rather just leave it as is if the intended way these methods are called is that `modifyFlags` isn't supposed to get called on the same flag list twice.

srdo-humio: I think your example works now, at least it seems to when I run it. Running test 1 and then trying to run the test class seems like it works as expected for me.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: Having dug into this, I think the backslash issue isn't solvable with changes in just this code.

We postprocess the flags in RunConfigurationFlagsState
```
flags.stream()
    .map(s -> ParametersListUtil.parse(s, false, true).get(0))
    .collect(Collectors.toList());
```

I ran a few tests on this util, and I'm not sure these lines do what they're supposed to.

`parse("--test_arg=this is a string").get(0)` gets you `--test_arg=this`, with anything after the whitespace stripped. 
I think this is why I have to add quotes around the test name, or we lose most of the name here.

`parse("--test_arg=\"this is a string\"").get(0)` gets you `--test_arg=this is a string`. 
Notice that this string doesn't contain quotes yet. Something later in the process must handle adding quotes for flags containing whitespace.

`parse("--test_arg="this is a string\\\"")` yields `--test_arg=this is a string"`. 
`parse("--test_arg="this is a string\\\\\"")` yields `--test_arg=this is a string\"`
I don't see a way to deal with a trailing backslash when the strings pass through this method, since any backslashes here get interpreted as an escape for the following quote.

I don't really understand why the postprocessing code I posted above throws away most of the parameter list, but adding quotes on my end doesn't seem like the right solution. Maybe RunConfigurationFlagsState should keep the entire flag including whitespaces, since it seems like something later on is adding quotes if necessary?

I'd prefer not dealing with this issue here. I think the inability to deal with backslashes could be tracked as a separate issue. I don't think it's something I can fix locally in the flags modifier, and I don't think backslashes in testnames are so crucial they should block merging this.

What do you think?

srdo-humio: I've removed the backslash replacement and the misleading test, since it doesn't work.

tpasternak:
Ok, LGTM

@mai93 @jastice  we don't have integration tests for this, but it seems we can't develop them due to limitations of current tests framework. Can we merge this?

jastice:
> @mai93 @jastice we don't have integration tests for this, but it seems we can't develop them due to limitations of current tests framework. Can we merge this?

Thanks! I think that's fine.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -45,25 +45,41 @@ class ScalaTestContextProvider implements TestContextProvider {
...
+
+    public ClassAndTestName(ScTypeDefinition testClass, @Nullable String testName) {
+      this.testClass = testClass;
+      this.testName = testName;
+    }
+  }
+
   @Nullable
-  private static ScClass getTestClass(ConfigurationContext context) {
+  private static ClassAndTestName getTestClass(ConfigurationContext context) {
```
mai93, line 82: Maybe renaming this method to getClassAndTestName()

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -76,32 +92,83 @@ private static ScClass getTestClass(ConfigurationContext context) {
...
+      }
+      int expectedSize = hasTestName() ? 4 : 2;
+      if (flags.size() < classSelectorIndex + expectedSize) {
+        return false;
+      }
+      String classSelector = flags.get(classSelectorIndex);
+      String clazz = flags.get(classSelectorIndex + 1);
+      boolean matchesClassFlags = classSelector.equals(testClassSelectorFlag) && clazz.equals(testClassFlag);
+      if (!hasTestName()) {
+        boolean hasTestSelector = flags.contains(testNameSelectorFlag);
```
mai93, line 144: Based on that `testNameSelectorFlag` is set in line 116 before checking that testName exists or not. this check can cause the method to return false if flags list has `testNameSelectorFlag` but no `testName`.

mai93:
Thank you @srdo-humio 
I've a question about this case:

```import org.scalatest.WordSpec
 import scala.collection.mutable.Stack

 class StackSpec extends WordSpec {

   "A Stack" should {

     "pop values in last-in-first-out order" in {
       val stack = new Stack[Int]
       stack.push(1)
       stack.push(2)
       assert(stack.pop() === 2)
       assert(stack.pop() === 1)
     }

     "throw NoSuchElementException if an empty stack is popped" in {
       val emptyStack = new Stack[String]
       intercept[NoSuchElementException] {
         emptyStack.pop()
       }
     }
   }
 }`

When I try to run the test from `"A Stack" should {` line, the methods are not shown in the Test Results window in the bottom left of the screen. Have you checked if this can be fixed?

File: scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java (hunk 4 above)
Comment (srdo) [reply scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java:82]: Will do.

File: scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java (hunk 5 above)
Comment (srdo) [reply scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java:144]: If flags contains `testNameSelectorFlag` but no `testName`, I think we want this method to return false in this case? What do you see as the issue here?

**Comment by srdo** [comment]
> When I try to run the test from `"A Stack" should {` line, the methods are not shown in the Test Results window in the bottom left of the screen. Have you checked if this can be fixed?

No, I haven't. Is this an issue introduced with these changes, or was it also broken before?

**Comment by mai93** [comment]
>No, I haven't. Is this an issue introduced with these changes, or was it also broken before?

It was not broken before because it used to run the whole test class independent of from where you hit the run tests button. So it used to always list all the tests it ran.

File: scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java [hunk 6]
```diff
@@ -76,32 +92,83 @@ private static ScClass getTestClass(ConfigurationContext context) {
...
+    @Nullable
+    private final String testNameSelectorFlag;
+    @Nullable
+    private final String testNameFlag;
+
+    public ScalatestTestSelectorFlagsModification(String testClassFqn, @Nullable String testName) {
+      this.testName = testName;
+      testClassSelectorFlag = BlazeFlags.TEST_ARG + "-s";
+      testClassFlag = BlazeFlags.TEST_ARG + testClassFqn;
+      testNameSelectorFlag = BlazeFlags.TEST_ARG + "-t";
```
mai93, line 116: To better understand what I'm missing in [this comment](https://github.com/bazelbuild/intellij/pull/3744/files/ce9cc2f4f3664ab2161c57edec9a27abba3b2fae#r948508380), why the `testNameSelectorFlag` is not added only if `testName` is not null, i.e. move this line inside the next if?

srdo: This isn't adding the flag, it's just setting a constant field in this object for later reference. The flags are set here 

https://github.com/bazelbuild/intellij/pull/3744/files/ce9cc2f4f3664ab2161c57edec9a27abba3b2fae#diff-f78adffc6524eb9e6e20bacab71d8bbea8b46ed0d16b60b421ecd064d3f2a122R159-R166

I could make the two fields holding the selector variables static, that might make this clearer?

srdo:
> It was not broken before because it used to run the whole test class independent of from where you hit the run tests button. So it used to always list all the tests it ran.

Okay, I looked at it a bit, and it appears that when running tests from that line, the test name ends up being the concatenation of both test names, so it fails to run. I'll look into what we can do here once I'm back in the office.

That being said, I think it was also broken before, just in a different way. Before you could only run the whole class. Now you can run the whole class (by putting your cursor on the class) or run each of the two tests individually.

srdo-humio:
@mai93 I believe the WordSpec example you posted should work now. Please take another look when you get a chance.

mai93:
@srdo-humio thank you, it is working fine now! 
If you can resolve the other open comments and @tpasternak does not have further comments we can merge this PR.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: Made the two fields static.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: Did this

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: Resolving this, since I think the behavior is correct

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: Resolving this, since it is a likely rare edge case that seems to require a decent amount of work to fix. Even if this is worth fixing, I don't think it should block the rest of this code from going in, since it is not a regression.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: Resolving since the given example works now

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
srdo-humio: Made the two variables static

srdo-humio:
@mai93 I believe every open comment has been resolved now.

PR #3755, opened 2022-07-08
Update 2022.2 Beta archives

mai93:
Use CLion 2022.2 Beta release (2022.2 EAP build 222.3345.16) and fix the Go 2022.2 plugin url.


jastice:
I used the Go plugin url with the parameters because initially it would give a 403 without them. Does it work for you like this?

mai93:
> I used the Go plugin url with the parameters because initially it would give a 403 without them. Does it work for you like this?

Yes, it is working now.

PR #3789, opened 2022-07-22
RemovedAction should delegate isDumbAware 

### LINKED ISSUE #3788: RemovedAction prevents DumbAware actions from being triggered in Dumb Mode
We noticed that some actions, such as Build Project, were no longer available during indexing (AKA Dumb Mode) even in a non-bazel project. After some debugging I identified the cause was ReplaceActionHelper wrapped the action with RemovedAction which abstracted the original action from the DumbAware check. 

This can also easily be reproduced via Tools > Internal Actions > Enter Dumb Mode.

The fix seems to be simply to implement PossiblyDumbAware

brian-mcnamara:
# Checklist

details.

Issue number: Fixes #3788

# Description of this change

### DIFF
--- a/common/actions/src/com/google/idea/common/actions/ReplaceActionHelper.java
+++ b/common/actions/src/com/google/idea/common/actions/ReplaceActionHelper.java
@@ -20,6 +20,7 @@ import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.actionSystem.EmptyAction;
 import com.intellij.openapi.actionSystem.Presentation;
+import com.intellij.openapi.project.PossiblyDumbAware;
 import com.intellij.openapi.project.Project;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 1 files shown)


tpasternak (approved):
Hey @brian-mcnamara, thanks for this change, LGTM.

common/actions/src/com/google/idea/common/actions/ReplaceActionHelper.java
```diff
@@ -67,7 +68,7 @@ public static void replaceAction(
   }
 
   /** Wraps an action and makes it conditionally invisible. */
-  private static class RemovedAction extends AnAction {
+  private static class RemovedAction extends AnAction implements PossiblyDumbAware {
```
jastice, line 71: The explicit implements here seems to be redundant.

jastice (approved):
Thanks!

common/actions/src/com/google/idea/common/actions/ReplaceActionHelper.java
brian-mcnamara: You're absolutely right. Removed.

Issue #2328, opened 2021-01-29
[ASWB] Mobile-install run configuration failed


ThomasCJY:
**Background**
Running mobile-install from bazel configuration doesn't work. This issue is regressed from https://github.com/bazelbuild/intellij/pull/1895/commits/12d8faf1cd404840acc9e2f4c1416dade89a05db

![Screenshot from 2021-01-28 17-59-36](https://user-images.githubusercontent.com/6951238/106221712-fcd6ff00-6192-11eb-8bc5-af5c002cedcd.png)

```
Waiting for target device...
Invoking mobile-install...
Command: /home/tchen/01_Projects/android/app/bazelisk mobile-install --tool_tag=ijwb:AndroidStudio "--adb_arg=-s " --adb_arg=801KPSL1432918 --adb /home/tchen/Android/Sdk/platform-tools/adb --device 801KPSL1432918:tcp:5037 --curses=no --color=yes --progress_in_terminal_title=no --build_event_binary_file=/tmp/intellij-bep-7b690874-d9b7-460d-805e-fb6de61ce23e --nobuild_event_binary_file_path_conversion -- //platform/core-ui/samples/playground:playground-apk --nolaunch_app

ANDROID_SDK is not set! This must point to the root directory of the Android SDK
2021/01/28 17:58:52 Using unreleased version 114 from gs://snap-bazel-releases
2021/01/28 17:58:52 Authenticating against GCS using default application credentials
INFO: Invocation ID: 35b58e51-d73a-4b4d-ac93-124359378e14
Loading:  (0 packages loaded)
Loading: 0 packages loaded
ERROR: Skipping '-nolaunch_app': no such target '//:-nolaunch_app': target '-nolaunch_app' not declared in package '' defined by /home/tchen/01_Projects/android/app/BUILD.bazel
WARNING: Target pattern parsing failed.
ERROR: no such target '//:-nolaunch_app': target '-nolaunch_app' not declared in package '' defined by /home/tchen/01_Projects/android/app/BUILD.bazel
INFO: Elapsed time: 3.468s (0.06m)
INFO: 0 processes (0.00%).
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
INFO: Build Event Protocol files produced successfully.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

rogchap:
[...]

Issue #2084, opened 2020-09-15
Unable to launch app: unknown flag --nolaunch_app


jagobagascon:
When I try to run the app from the Android Studio UI the flag `--nolaunch_app` at the end of the command. That is triggering the following error:
```
Waiting for target device...
Invoking mobile-install...
Command: /usr/local/bin/bazel mobile-install --tool_tag=ijwb:AndroidStudio --device 94PAY0RKQL:tcp:5037 "--adb_arg=-s " --adb_arg=94PAY0RKQL --adb /Users/jagoba/Library/Android/sdk/platform-tools/adb --curses=no --color=yes --progress_in_terminal_title=no --build_event_binary_file=/var/folders/0p/db4t1g1d2zxgd91vl02ff0fw0000gn/T/intellij-bep-a9ded8a1-4a1a-4aca-95f9-923c12f3b377 --nobuild_event_binary_file_path_conversion -- //example/MyApp/src/main:android --nolaunch_app

Loading: 
Loading: 0 packages loaded
ERROR: Skipping '-nolaunch_app': couldn't determine target from filename '-nolaunch_app'
ERROR: couldn't determine target from filename '-nolaunch_app'
INFO: Elapsed time: 0.725s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
INFO: Build Event Protocol files produced successfully.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

I haven't found any documentation about that flag in the bazel docs or in Google.

I managed to run the app by disabling the option "Use mobile-install". But I'm still curious about the `--nolaunch_app` flag that is failing.

memostark:
I have the same problem with Android Studio 4.0.1 on Windows 10, using Bazel 3.7.1, plugin 2020.06.25.0.3. Can build and install apps using the terminal with no problems

```
ERROR: Skipping '-nolaunch_app': couldn't determine target from filename '-nolaunch_app'
ERROR: couldn't determine target from filename '-nolaunch_app'
```

c1139676919:
**I'm having the same problem on macos catalina 、Bazel3.4.0，** 

`Waiting for target device...
Invoking mobile-install...
[...]

PR #3793, opened 2022-07-25
Remove nolaunch_app to fix mobile-install

TylerCarberry:
## Checklist

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See  the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more details.

## Discussion thread for this change
https://github.com/bazelbuild/intellij/issues/2328 and https://github.com/bazelbuild/intellij/issues/2084

## Description of this change

When running `mobile-install` through Android Studio, it appends the `--nolaunch_app` flag to the command. This flag cannot be found and prevents the app from being installed. This PR removes it. Perhaps `nolaunch_app` was removed or it only exists in Blaze?

```
Invoking mobile-install...
Command: bazel mobile-install --tool_tag=ijwb:AndroidStudio --adb sdk/platform-tools/adb --device ABC123:tcp:7890 --curses=no --color=yes --progress_in_terminal_title=no --build_event_binary_file=/tmp/intellij-bep-3 --nobuild_event_binary_file_path_conversion -- //src/main:template_app --nolaunch_app

Loading: 
Loading: 0 packages loaded
ERROR: Skipping '-nolaunch_app': no such target '//:-nolaunch_app': target '-nolaunch_app' not declared in package '' defined by android-bazel-java-app-template/BUILD
WARNING: Target pattern parsing failed.
ERROR: no such target '//:-nolaunch_app': target '-nolaunch_app' not declared in package '' defined by android-bazel-java-app-template/BUILD
```

This PR https://github.com/bazelbuild/intellij/pull/3202 attempted to fix the issue but I was still able to reproduce using that branch.

## Root Cause Analysis
This was added in https://github.com/bazelbuild/intellij/pull/1895

## Reproduction Steps

1. Have an android bazel project. I used https://github.com/jaredsburrows/android-bazel-java-app-template
2. On the run configuration check `mobile-install`
3. Press the run button

Screenshot from 2022-07-25 16-00-46

Android Studio: `Chipmunk 2021.2.1 Patch 1`
Bazel Plugin: `2022.06.14.0.1-api-version-212`
bazel: `5.2.0`

### DIFF
--- a/aswb/src/com/google/idea/blaze/android/run/binary/mobileinstall/MobileInstallBuildStep.java
+++ b/aswb/src/com/google/idea/blaze/android/run/binary/mobileinstall/MobileInstallBuildStep.java
@@ -150,6 +150,7 @@ public class MobileInstallBuildStep implements ApkBuildStep {
       return;
     }
 
+    BuildSystemName buildSystemName = Blaze.getBuildSystemName(project);
     BuildInvoker invoker =
         Blaze.getBuildSystemProvider(project).getBuildSystem().getBuildInvoker(project, context);
     BlazeCommand.Builder command = BlazeCommand.builder(invoker, BlazeCommandName.MOBILE_INSTALL);
@@ -167,7 +168,7 @@ public class MobileInstallBuildStep implements ApkBuildStep {
     }
 
     WorkspaceRoot workspaceRoot = WorkspaceRoot.fromProject(project);
-    final String deployInfoSuffix = getDeployInfoSuffix(Blaze.getBuildSystemName(project));
+    final String deployInfoSuffix = getDeployInfoSuffix(buildSystemName);
 
     try (BuildResultHelper buildResultHelper = invoker.createBuildResultHelper();
         AdbTunnelConfigurator tunnelConfig = getTunnelConfigurator(context)) {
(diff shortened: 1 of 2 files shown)


alexrwegener:
@jastice @mai93 This would be awesome to get in. Everyone who tries to run the bazel plugin runs into this the first time they try to deploy an Android app to a device.

mai93:
@estebandlc Do you think it is safe to remove this flag since it is not available in Bazel version of mobile-install?

estebandlc:
Well, there might be two parts of this issue because I haven't checked if the flag does not exist, but one thing is for sure the error shown above is not because the flag does not exist. The command shown is:

```
bazel mobile-install --tool_tag=ijwb:AndroidStudio [...] -- //src/main:template_app --nolaunch_app
```

And the error is 
```
Skipping '-nolaunch_app': no such target '//:-nolaunch_app':
```

because that flag is interpreted as a target as it's after the `--`

The issue there is that the flag is in the wrong place.

estebandlc:
Ah my bad, this is an execution flag, could we do this only if BuildSystem type is BAZEL?

estebandlc:
What happens with this change when Studio will want the app to be installed but not launched and it's now being launched?

TylerCarberry:
Assuming you're talking about this setting, it still works correctly with this change. The app doesn't get launched if you have `Do not launch Activity` checked

Screenshot from 2022-08-01 11-26-21

TylerCarberry:
This is with bazel, not sure if it breaks blaze. How would I check the BuildSystem?

mai93:
This LGTM, @estebandlc do you have any other comments?

estebandlc:
LGTM

mai93 (approved):

PR #3812, opened 2022-08-02
IntelliJ 2022.2 compatibility fixes

mai93:
- Replace `boolean getWasFullRescanning()` with `ScanningType getScanningType()` and return `ScanningType.PARTIAL`.
-  Use bundled IntelliLang plugin in golang integration tests

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -100,8 +100,8 @@ http_archive(
 http_archive(
     name = "clion_2022_2",
     build_file = "@//intellij_platform_sdk:BUILD.clion222",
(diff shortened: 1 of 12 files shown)


AlexeyGy (approved):
LGTM, I am assuming some features of the Go plugin were migrated into the IntelliLang plugin (bundled with UE)?

mai93:
> LGTM, I am assuming some features of the Go plugin were migrated into the IntelliLang plugin (bundled with UE)?

I'm not sure what features were moved to IntelliLang but it is now a dependency of Go plugin. So the Go tests fail with `Required plugin 'org.jetbrains.plugins.go' is not enabled` if IntelliLang is not added.

PR #3838, opened 2022-08-12
Update product versions aliases

mai93:
Since the plugin will be supporting the last 2 versions of the IDE, this PR changes the versions aliases to reflect this better. 
- `{PRODUCT}-oss-oldest-stable`: the oldest stable IDE version supported
- `{PRODUCT}-oss-latest-stable`: the second stable IDE version supported
- `{PRODUCT}-oss-under-dev`: the EAP version we are working towards supporting


jastice (approved):
Thanks, I think this will make things clearer. Personally I would prefer `old-stable` over `oldest-stable`, but that's a nit.

PR #3885, opened 2022-08-24
Prevent auto import of project if registry key is set

brian-mcnamara:
# Checklist

details.

We are slowly migrating a project to bazel and the auto import code is causing issues and making it hard for developers to work the project in the mean time. 

Ultimatly a easy fix for this is to use the `external.system.auto.import.disabled` key (defined [here](https://github.com/JetBrains/intellij-community/blob/master/platform/util/resources/misc/registry.properties#L1476)) to allow us to disable the auto import functionality. 

Its worth noting, this is used in IntelliJ community already to disable maven importing, we can use a different registry key, but I figure this is specific enough as is to qualify for this use case.

The change is pretty minimal. Returning false to canOpenProject if the registry value is set to true (defaults to false) to allow bypassing the auto import functionality.

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -312,6 +312,7 @@
         implementationClass="com.google.idea.blaze.base.run.producers.BuildFileRunLineMarkerContributor"/>
     <actionConfigurationCustomizer implementation="com.google.idea.blaze.base.plugin.BlazeHideMakeActions"/>
     <notificationGroup displayType="BALLOON" id="BuildifierBinaryMissing"/>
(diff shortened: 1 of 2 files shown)


liucijus (approved):

jastice:
I don't think `external.system.auto.import.disabled` is the right registry key, using it here can have side effects in other plugins. I suggest adding a custom key, or perhaps better a UI-accessible option in the plugin settings.

brian-mcnamara:
I can create a new key. I don't think a UI setting is appropriate since 99% of the time you do want auto import to work if a bazel build is detected.

tpasternak:
As #3885 came in, we probably don't need the checkbox

PR #3901, opened 2022-09-01
Detect project data directory per IDE product

liucijus:
Fixes hardcoded Intellij project data path to support other IDE products as reported in https://github.com/bazelbuild/intellij/pull/3601#issuecomment-1219955183


base/src/com/google/idea/blaze/base/sync/data/BlazeDataStorage.java
```diff
@@ -33,7 +33,7 @@ public class BlazeDataStorage {
   public static final String BLAZE_DATA_SUBDIRECTORY = ".blaze";
   public static final String PROJECT_DATA_SUBDIRECTORY = getProjectDataSubdirectory();
```
jastice, line 34: why not use the public field instead?

liucijus: thanks, good point!

mai93 (approved):

PR #3902, opened 2022-09-01
Add checkbox controlling if auto-import feature is enforced

### LINKED ISSUE #3895: Auto Import Feature might cause conflicts with other build tools
### Description of the bug:

Recently, we've introduced Auto Import feature in #3601. It works correctly but seems to be conflicting with other build tools. If you have more than one plugin that is able to import a project, bazel will always take precedence. 

The root cause is that, the `AutoImportProjectOpenProcessor`'s `canOpenProject` field is doing only a very brief check - only if there is a WORKSPACE file in the directory. On the other hand, it is marked as `isStrongProjectInfoHolder`, which means it will run even if there's an already existing `.idea` file.  (according to docs https://github.com/JetBrains/intellij-community/blob/9e68e0f60c4e58083a7ca7cfc35514f5da8fb4e0/platform/platform-api/src/com/intellij/projectImport/ProjectOpenProcessor.kt#L40)

_No response_

_No response_

_No response_

_No response_

_No response_

_No response_

tpasternak:
Closes #3895

# Checklist

details.


liucijus:
@tpasternak do you know a usecase when someone may want to enforce auto import? I guess for teams working with Bazel projects it's expected to have managed IDE and have less per user toggling of such setting. Maybe it should be controlled by env var or some other configuration mean. WDYT?

tpasternak:
Oh, that's a good point, but it still can be programatically modified by just editing `options/blaze.user.settings.xml`

tpasternak:
Probably, we don't need it, as #3885 is in

tpasternak:
@jastice

mai93:
@tpasternak do you think we can merge https://github.com/bazelbuild/intellij/pull/3900 as a fix for https://github.com/bazelbuild/intellij/issues/3895 then?

tpasternak:
Oh, honestly I think this problem is already covered by #3885 . If not, then I would prefer to take this PR instead of #3900, as enforcing might sometimes be useful (in gateway mode)

Issue #3216, opened 2022-01-27
Run gazelle on project sync


blorente:
## Problem
For projects that use gazelle, moving a single file is a two-step process: First they must run gazelle, then they must sync to trigger re-indexing if IntelliJ didn't pick those changes up.

Ideally, this would be a single-click action, whereby a user only has to click "sync" after a refactoring is done.

## Proposed Solution
Run Gazelle on sync, at least for full, non-incremental syncs of projects, before we run the bazel build that enables the actual sync.

Regarding performance, Gazelle accepts packages as arguments, so I believe it should be possible to inspect the `.bazelproject` file to gather which packages it should sync, therefore correlating performance to the size of the indexed project.

The changes I imagine would be needed:
- Additional entries in the Bazel Plugin preferences, where the user can specify:
  - The label of the Gazelle target to run.
  - The frequency of the gazelle runs (on every sync, or only on full syncs).
- Custom logic to derive go packages from the `directories` entry in the `.bazelproject`.
- A call to gazelle during project sync, which would run before even the initial query.

I'm happy to spend the effort of implementing this, but as per the contributing guidelines I thought I'd ask first if this would be an interesting contribution, and if I missed important pieces.

blorente:
I have end-to-end example working, implementing `SyncTask` that runs between `ProjectStateSyncTask` and `ProjectUpdateSyncTask`.

After diving a bit deeper into the code, I think I see three possible options for implementing this:
- The above, where we introduce a new `ProjectPreUpdateSyncTask` between `ProjectState` and `ProjectUpdate`.
- To merge it into the `ProjectUpdateSyncTask`, and run it before any of the other `SyncPlugins` run. Should be an easy change if we decide to go in this direction.
- To create a `SyncPlugin` to run gazelle. This has the benefit that it makes configuring it trivial by hooking up to the existing infrastructure to parse `.bazelproject`, but (as far as I know) there is no way to specify ordering, and this step should run before any other Bazel/Blaze builds.

I'd appreciate any input in this matter.

fkorotkov:
I'm very much interested in the Gazelle use case as well. IMO the option with `SyncListener` and having it configured in `ProjectViewSet` seems like the most clean solution for me. To @blorente points:

1. We can get `ProjectViewSet` in the listener via `ProjectViewManager`.
[...]

PR #3910, opened 2022-09-09
feat: Add gazelle to plugin

blorente:
This is a WIP because I can't figure out how to make the tests work.

# Checklist

details.

Issue number: `3216`

# Description of this change

See linked issue for a description and general design.
This is a port of the existing plugin: https://github.com/blorente/intellij_gazelle. Refer to that README for usage.

Architecturally, this plugin does two things:
- Implements a `GazelleSyncPlugin`, which inserts itself between the `bazel info` and `bazel build` phases of a sync cycle, to run gazelle in between, via a `BlazeGazelleRunnerImpl`.
- Hooks into UserSettings, to add a global field to specify a gazelle target to run on sync.

Refer to `gazelle/src/META-INF/blaze-gazelle.xml` for a reference of the extension points to the plugin.

**NOTE:** The added tests in this PR set up a full repository and run gazelle. This is potentially expensive, in the order of single-digit seconds.

### DIFF
--- a/gazelle/src/com/google/idea/blaze/gazelle/BlazeGazelleRunner.java
+++ b/gazelle/src/com/google/idea/blaze/gazelle/BlazeGazelleRunner.java
@@ -21,7 +21,8 @@ public abstract class BlazeGazelleRunner {
   /**
    * Run the provided Gazelle target via Blaze.
    *
-   * @return
+   * @return GazelleRunResult,
+   *         an enum determining whether the command succeeded, failed, and/or had errors.
    */
   public abstract GazelleRunResult runBlazeGazelle(
       BlazeContext context,
--- a/gazelle/src/com/google/idea/blaze/gazelle/BlazeGazelleRunner.java
+++ b/gazelle/src/com/google/idea/blaze/gazelle/BlazeGazelleRunner.java
@@ -1,0 +1,33 @@
+package com.google.idea.blaze.gazelle;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.idea.blaze.base.bazel.BuildSystem;
+import com.google.idea.blaze.base.model.primitives.Label;
+import com.google.idea.blaze.base.model.primitives.WorkspaceRoot;
+import com.google.idea.blaze.base.projectview.section.sections.DirectoryEntry;
+import com.google.idea.blaze.base.scope.BlazeContext;
+import com.intellij.openapi.components.ServiceManager;
+
+import java.util.List;
+
+/** Runs the blaze run command, on gazelle. The results may be cached in the workspace. */
+public abstract class BlazeGazelleRunner {
+    public static BlazeGazelleRunner getInstance() {
+        return ServiceManager.getService(BlazeGazelleRunner.class);
+    }
+
+    /**
+     * This will call Blaze run with the argument of the gazelle target provided.
+     *
+     * @param blazeFlags The blaze flags that will be passed to Blaze.
+     * @return stdout of the gazelle process.
+     */
+    public abstract ListenableFuture<String> runBlazeGazelle(
+            BlazeContext context,
+            BuildSystem.BuildInvoker invoker,
+            WorkspaceRoot workspaceRoot,
+            List<String> blazeFlags,
+            Label gazelleTarget,
+            List<DirectoryEntry> directories
+    );
+}
--- a/gazelle/src/com/google/idea/blaze/gazelle/BlazeGazelleRunner.java
+++ b/gazelle/src/com/google/idea/blaze/gazelle/BlazeGazelleRunner.java
@@ -1,33 +1,34 @@
 package com.google.idea.blaze.gazelle;
 
-import com.google.common.util.concurrent.ListenableFuture;
-import com.google.idea.blaze.base.bazel.BuildSystem;
+import com.google.common.collect.ImmutableList;
+import com.google.idea.blaze.base.bazel.BuildSystem.BuildInvoker;
+import com.google.idea.blaze.base.issueparser.BlazeIssueParser.Parser;
 import com.google.idea.blaze.base.model.primitives.Label;
+import com.google.idea.blaze.base.model.primitives.WorkspacePath;
 import com.google.idea.blaze.base.model.primitives.WorkspaceRoot;
-import com.google.idea.blaze.base.projectview.section.sections.DirectoryEntry;
 import com.google.idea.blaze.base.scope.BlazeContext;
 import com.intellij.openapi.components.ServiceManager;
-
+import java.util.Collection;
 import java.util.List;
 
 /** Runs the blaze run command, on gazelle. The results may be cached in the workspace. */
 public abstract class BlazeGazelleRunner {
-    public static BlazeGazelleRunner getInstance() {
-        return ServiceManager.getService(BlazeGazelleRunner.class);
-    }
 
-    /**
-     * This will call Blaze run with the argument of the gazelle target provided.
-     *
-     * @param blazeFlags The blaze flags that will be passed to Blaze.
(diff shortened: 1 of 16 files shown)


HaimingTian:
If I'm not mistaken this will run gazelle on a list of preconfigured targets, could potentially take a long time if working on several different projects in a large monorepo. I guess this is in no way blocking since if users don't configure anything, it won't run at all.

Just FYI I've been working on something similar but only run gazelle on a list of paths that contain dependency graph changes, I can open another PR with my changes when they are ready.

gazelle/tests/integrationtests/com/google/idea/blaze/gazelle/sync/RunGazelleOnSyncTest.java
```diff
@@ -0,0 +1,184 @@
...
+                "directories:",
+                "  ."
+        );
+
+        runBazelSync();
+
+        errorCollector.assertNoIssues();
+
+        File resultBuildFile = workspaceRoot.fileForPath(new WorkspacePath("src/BUILD.bazel"));
+        assertThat(workspaceRoot.isInWorkspace(resultBuildFile)).isFalse();
```
mai93, line 127: I'm not sure but I think `workspaceRoot.isInWorkspace()` will compare the path of the file to the path of the workspace directory instead of checking the existence of the build file which we want here. So if you can confirm that, we can replace this line with `assertThat(resultBuildFile.exists()).isFalse();`  
Same for the other tests.

gazelle/src/com/google/idea/blaze/gazelle/GazelleSyncPlugin.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.google.common.collect.ImmutableList;
+import com.google.idea.blaze.base.projectview.section.SectionParser;
+import com.google.idea.blaze.base.sync.BlazeSyncPlugin;
+
+import java.util.Collection;
+
+public class GazelleSyncPlugin implements BlazeSyncPlugin {
+    public Collection<SectionParser> getSections() {
+        return ImmutableList.of(GazelleSection.PARSER);
```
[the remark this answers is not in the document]
blorente: - As far as I understand, the plugin settings are global, for all projects. For instance, the bazel binary you want to use tends to be the same, regardless of which project you're using. The Gazelle target, however, can vary from project to project. One repository might have it in `//:gazelle`, whereas another repo might store it under `//tools:gazelle`, or something different. Therefore, this configuration should be per-project. The plugin settings exist as a stopgap, so that one can start using the Gazelle functionality somewhat, before one's entire team has switched to the new version.

- Thanks! Will do.

gazelle/tests/integrationtests/com/google/idea/blaze/gazelle/sync/RunGazelleOnSyncTest.java
blorente: Good catch, you were right. I've changed it for that test, and added the actual existence check for the other tests.

gazelle/src/com/google/idea/blaze/gazelle/GazelleSyncListener.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+    public void onSyncStart(Project project, BlazeContext context, SyncMode syncMode)
+            throws SyncScope.SyncFailedException, SyncScope.SyncCanceledException {
+        if (syncMode == SyncMode.NO_BUILD) {
+            return;
+        }
+        ProjectViewSet projectViewSet = ProjectViewManager.getInstance(project).getProjectViewSet();
+        Optional<Label> gazelleBinary = this.getGazelleBinary(projectViewSet);
+        BuildSystem.BuildInvoker invoker = Blaze.getBuildSystemProvider(project).getBuildSystem().getBuildInvoker(project, context);
+        List<DirectoryEntry> importantDirectories = projectViewSet.listItems(DirectorySection.KEY);
```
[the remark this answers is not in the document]
blorente: - Switched to `ImportRoots`, thanks for the tip! It does seem safer.
- Yes, that will be the case at the moment. It's not very efficient, but @HaimingTian and I could collaborate on finding a good subset of changed files to run it on.

gibfahn:
I believe this fixes https://github.com/bazelbuild/intellij/issues/3216

blorente:
@mai93 Thank you so much for your comments, I think the plugin has improved a lot as a result.
I've tried to address all of them, if you could take a look at your earliest convenience, it would be great.

gazelle/src/com/google/idea/blaze/gazelle/BlazeGazelleRunner.java
```diff
@@ -1,33 +1,34 @@
...
-            List<DirectoryEntry> directories
-    );
+  public static BlazeGazelleRunner getInstance() {
+    return ServiceManager.getService(BlazeGazelleRunner.class);
+  }
+
+  /**
+   * Run the provided Gazelle target via Blaze.
+   *
+   * @return
```
mai93, line 24: Please add a description for the method return

gazelle/src/com/google/idea/blaze/gazelle/GazelleSyncListener.java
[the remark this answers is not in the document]
mai93: Do you think you have the time to add this enhancement first or we can merge this PR and add a TODO for it?

gazelle/src/com/google/idea/blaze/gazelle/GazelleUserSettings.java
```diff
@@ -0,0 +1,58 @@
...
+  public void setGazelleTarget(String target) {
+    this.gazelleTarget = target;
+  }
+
+  public Optional<Label> getGazelleTargetLabel() throws InvalidTargetException {
+    String rawTarget = getGazelleTarget();
+    if (rawTarget.equals(EMPTY_GAZELLE_TARGET)) {
+      return Optional.empty();
+    }
+    Label targetLabel = (Label) TargetExpression.fromString(rawTarget);
```
mai93, line 45: When I tried it manually with a label like "//pkg" in the settings, the plugin crashes with this error during sync
```
Error:Internal error: class com.google.idea.blaze.base.model.primitives.TargetExpression cannot be cast to class com.google.idea.blaze.base.model.primitives.Label (com.google.idea.blaze.base.model.primitives.TargetExpression and com.google.idea.blaze.base.model.primitives.Label are in unnamed module of loader com.intellij.ide.plugins.cl.PluginClassLoader @27226178)
```
I think this is because `TargetExpression.fromString()` returns `TargetExpression` if the label is not valid ([link](https://github.com/bazelbuild/intellij/blob/master/base/src/com/google/idea/blaze/base/model/primitives/TargetExpression.java#L52)). Do you think we can replace this with `Label.validate()`?

gazelle/src/com/google/idea/blaze/gazelle/GazelleSyncListener.java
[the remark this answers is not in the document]
blorente: I'd rather leave it as a TODO, given that the plugin is opt-in. I'll add that.

gazelle/src/com/google/idea/blaze/gazelle/GazelleUserSettings.java
blorente: Sure! Thanks for the explanation.

blorente:
@mai93 Thanks for the review! The issues with the tests was that the UI setup wasn't done properly for testing. Turns out, we should probably not do any UI setup for testing at all, so added a headless mode in the latest commit.

Tests should pass now, please let me know what you think.

mai93 (approved):

PR #3969, opened 2022-09-27
Update stale.yml

mai93:
since this script is for issues only then the right label should be `more-data-needed`.


sgowroji (approved):
LGTM, Approved!

PR #4003, opened 2022-10-07
Do not include absolute paths in debug build items

### LINKED ISSUE #2101: can't debug on macos using clion
**details**
i built  a simple project   < [example](https://github.com/dwtj/clwb-bugs-example) >  successfully  and i could run it directly , but i can't debug it 
clion just stuck on debug state and disable my breakpoints,

image

i tried to add  some flags to command line , but it's not working 
-c dbg --spawn_strategy=standalone --strategy_regexp=^Linking=local
image

**my environment:**
xcode : 12.0
clion: v2019.3.6
bazel plugin : v2020.07.13
bazel : release 3.5.0

tpasternak:
This should fix CLion debugging on Mac OS

Closes #2101

Flags are valid only for clang and lld, so we need to preserve
information about compiler type in BlazeCompilerInfoMap

# Checklist


isaactorz:
@tpasternak I've built your branch and tested it on an x86 mac and can confirm that it works on this platform 👍🏼 .

Out of curiosity do you have any reference for this `-oso_prefix` option for the linker?

Can't find anything about it online.

cpp/src/com/google/idea/blaze/cpp/BlazeConfigurationResolver.java
```diff
@@ -98,9 +103,21 @@ public BlazeConfigurationResolverResult update(
         HeaderRootTrimmer.getValidRoots(
             context, blazeProjectData, toolchainLookupMap, targetFilter, executionRootPathResolver);
     builder.setValidHeaderRoots(validHeaderRoots);
+    builder.setTargetToVersionMap(targetToVersion);
     return builder.build();
   }
 
+  @NotNull
+  private static Map<TargetKey, CompilerInfo> getTargetToVersionMap(ImmutableMap<TargetKey, CToolchainIdeInfo> toolchainLookupMap, ImmutableMap<CToolchainIdeInfo, BlazeCompilerSettings> compilerSettings) {
+    Map<ExecutionRootPath, String> compilerVersionByPath =
```
mai93, line 112: Can this function be simplified to something like the following?
```
 return toolchainLookupMap.entrySet().stream()
            .collect(toImmutableMap(e -> e.getKey(), e -> compilerSettings.get(e.getValue()).getCompilerVersion()));
```
or will this be missing something?

cpp/src/com/google/idea/blaze/cpp/BlazeConfigurationResolver.java
```diff
@@ -44,17 +45,19 @@
...
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
+import static java.util.AbstractMap.*;
```
mai93, line 59: Can we avoid using import `.*`?

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -126,6 +127,7 @@ public void update(
     BlazeConfigurationResolverResult newResult =
         configurationResolver.update(
             context, workspaceRoot, projectViewSet, blazeProjectData, oldResult);
+    BlazeCompilerInfoMap.getInstance(project).setState(newResult.targetToCompilerVersion);
```
mai93, line 130: Not a review comment but just for me to understand, will this map be populated on Startup syncs? If the IDE is closed then reopened the plugin picks up the last sync state from disk, will this map be loaded from there as well or it will require manual sync?

mai93, line 130: Not a review comment but just for me to understand, will this map be populated on Startup syncs? If the IDE is closed then reopened the plugin picks up the last sync state from disk, will this map be loaded from there as well or it will require manual sync?

cpp/src/com/google/idea/blaze/cpp/BlazeConfigurationResolver.java
mai93: Also if it is possible, we can have the map key set to the target label string from here instead of transforming it in the BlazeCompileInfoMap class.

cpp/src/com/google/idea/blaze/cpp/CompilerInfo.java
```diff
@@ -0,0 +1,21 @@
+package com.google.idea.blaze.cpp;
+
+public class CompilerInfo {
```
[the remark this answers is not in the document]
mai93: I think we should remove this class and maybe we can call the map `targetToCompilerVersionStrMap` and this way it will be clear that `targetToCompilerVersionStrMap.get(target)` calls return compiler version strings.

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
mai93: I see, thank you.

cpp/src/com/google/idea/blaze/cpp/BlazeCompilerInfoMapService.java
```diff
@@ -0,0 +1,66 @@
...
+
+    BlazeCompilerInfoMapService(Project p) {
+
+    }
+
+    public void setState(Map<TargetKey, String> compilerInfoMap) {
+        this.state = new State(compilerInfoMap.entrySet().stream()
+                .collect(Collectors.toMap(e -> e.getKey().getLabel().toString(), Map.Entry::getValue)));
+    }
+    State state = new State();
```
mai93, line 51: - Can we make this variable private?

cpp/src/com/google/idea/blaze/cpp/BlazeConfigurationResolver.java
```diff
@@ -98,9 +100,27 @@ public BlazeConfigurationResolverResult update(
...
+  @NotNull
+  private static ImmutableMap<TargetKey, String> getTargetToVersionMap(ImmutableMap<TargetKey, CToolchainIdeInfo> toolchainLookupMap, ImmutableMap<CToolchainIdeInfo, BlazeCompilerSettings> compilerSettings) {
+    Map<ExecutionRootPath, String> compilerVersionByPath =
+            compilerSettings.entrySet().stream().collect(
+                    Collectors.toMap(
+                            e -> e.getKey().getCppExecutable(),
+                            e -> e.getValue().getCompilerVersion()));
+    Map<TargetKey, String> map = toolchainLookupMap.entrySet().stream()
+            .map(e -> new AbstractMap.SimpleImmutableEntry<>(
+                    e.getKey(),
```
mai93, line 116: - Can we replace this with `e.getKey().getLabel().toString()` to avoid going over the stream again in `BlazeCompilerInfoMapService$setState`?

cpp/src/com/google/idea/blaze/cpp/BlazeConfigurationResolver.java
```diff
@@ -98,9 +100,27 @@ public BlazeConfigurationResolverResult update(
...
+                    compilerVersionByPath.get(e.getValue().getCppExecutable())))
+            // In case of a broken compiler, the version string is null, but Collectors.toMap requires non-null value function.
+            .filter(e -> e.getValue() != null)
+            .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
```
mai93, line 120: I think we can use `.collect(toImmutableList())` to avoid `ImmutableList.copyOf()` at the end of the method

`toImmutableList()` can be imported as `import static com.google.common.collect.ImmutableList.toImmutableList;
`

mai93:
Thank you @tpasternak, I think we still have these two unresolved comments from the first review round
https://github.com/bazelbuild/intellij/pull/4003#discussion_r1008408930
https://github.com/bazelbuild/intellij/pull/4003#discussion_r1008416273
and we can merge the PR.

mai93 (approved):

Issue #3097, opened 2021-12-03
Current directory not correctly set when debugging Go


Enrico2:
This issue was reported here a while ago: https://github.com/bazelbuild/intellij/issues/319
I'm seeing the exact behavior again, in IntelliJ ultimate, with Golang.

bazel: 2021.11.09.0.2-api-version-212
![image](https://user-images.githubusercontent.com/161314/144683320-8f8c93dc-7ac9-4fcb-a55d-f0506dae3b7d.png)

Specifically, when using "Run" on a go_binary target I see
```
os.Getwd():
/private/var/tmp/_bazel_ranmagen/68a892d02d10473428c53fc867ebfbd4/execroot/coinbazel/bazel-out/darwin-fastbuild/bin/xyz/xyz/xyz/server.runfiles/coinbazel
```
But when using "Debug" I see
```
os.Getwd():
/Users/ranmagen/src/monorepo
```

hansonap:
I am seeing the exact same issue as #319 with CLion 2021.2.3 (Build #CL-212.5457.51, built on October 14, 2021)

alice-ks:
This is not exactly the same issue as #319. #319 was about debugging in C++. What you're describing is for Go debugging.

For Go debugging, there are two code paths. The default one uses a script file generated by Bazel. That code path uses the workspace directory as working directory.

An alternative code path directly runs the test. That code path sets the working directory to the 'runfiles' directory. This alternative code path doesn't support args, though, and hence isn't used by default.

Before we update the working directory for the script file version, we'd need to figure out what would be the right working directory for that case. In the meantime, you can use a workaround to enable the alternative code path. Do so by creating `~/.intellij-experiments` (if that file doesn't exist for you) and specify `blaze.go.script.path.enabled.2=false`.

JamesRTaylor:
@alice-ks, your workaround isn't working for me. I'm using GoLand 2022.1.4 on macOS with the latest bazel plugin. Is there somewhere I can put this setting in a project file that I could check-in instead (so I don't have to ask everyone on my team to make a change)?

alice-ks:
@JamesRTaylor I'm not working on the project anymore.

@mai93 FYI.

JamesRTaylor:
@mai93 - would it be possible to switch the default to use the alternate code path so that the debugger and runtime use the same current working directory?

sgowroji:
[...]

PR #4043, opened 2022-10-24
Issue #3097 Current directory not correctly set when debugging Go

JamesRTaylor:
# Checklist

- [ x] I have filed an issue about this change and discussed potential changes with the maintainers.
- [ x] I have received the approval from the maintainers to make this change.
- [ x] This is not a stylistic, refactoring, or cleanup change.

details.

Issue number: `https://github.com/bazelbuild/intellij/issues/3097`

# Description of this change

Turn off the experimental `blaze.go.script.path.enabled.2` config by default so that the current working directory when tests are run is the same for run vs debug time.


blorente:
Hello @JamesRTaylor !
Sorry for taking so long to respond, Alice (who had the context in the previous issue) has stepped down as a maintainer, and I wanted to consult the other maintainers about this change.
Sadly, I don't think we should merge this change, because users would lose the ability to pass arguments to the go tests.

To solve the issue, I see two alternatives:
- Figure out if pointing the script to the runfiles directory is the right thing to do, and make the script `cd` there before running. To test this, we could just make the change, and see if `dlv` picks up the right source files in `Debug` mode.
- Have a mechanism to check in experiment flags, maybe as fields in the `.bazelproject` files.

I don't have the time to implement either right now, but if you'd be interested in contributing either of those changes, I'd love to review them.

Issue #3080, opened 2021-11-26
Java 17 (Preview)


tadeoj:
Hello.

Is there any way to indicate in the java_language_level option to use the preview mode?

In Java 17 preview you can use the new switch expressions and we are very interested.

Thanks.

aarone:
i ran into the same issue and can contribute a change to support this. It looks like the plugin accepts a Java major version that is then mapped to an Intellij project language level. Language levels are (now?) a superset of Java major versions which is why the preview language levels aren't supported.

It looks like all of the possible language levels are defined in a enum (`com.intellij.pom.java.LanguageLevel`). Any issues with me extending the project view `java_language_level` option to support these enum names while keeping backwards compatibility with existing major versions?

mai93:
@aarone I think we can review a PR that implements this extension.

aarone:
thanks! see related PR

mai93:
Fixed with #4061

PR #4061, opened 2022-10-31
Support Preview Java Language Levels

aarone:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/3080

# Description of this change

Intellij project language levels may include variants of major Java
versions such as "17 (preview)". Today, the Bazel plugin
`java_language_level` option specified in in the Project View File
accepts an integer value which defines the major Java version to use.
As an integer, it is not possible to specify preview versions.

This changes adds support for specifying any Java language level by
specifying a string for `java_language_level` that is attempted to be
matched against the LanguageLevel enum value names (found in
`com.intellij.pom.java.LanguageLevel`). If no match is found, the
previous parsing logic is used to parse the given language version.

To enable Java 17 preview features, you can set:

```yaml
java_language_level: jdk_17_preview
```

in your project view file.

To use a preview language level, the version of the SDK providing that level must be
available to Intellij.

### DIFF
--- a/java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
+++ b/java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
@@ -58,7 +58,7 @@ public class JavaLanguageLevelSection {
       if (numericVersion <= 0) {
         return defaultValue;
       }
-    } catch (NumberFormatException nje) {
+    } catch (NumberFormatException e) {
       return defaultValue;
     }
 
--- a/java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
+++ b/java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
@@ -44,17 +44,17 @@ public class JavaLanguageLevelSection {
   @Nullable
   @VisibleForTesting
   static LanguageLevel getLanguageLevel(String level, @Nullable LanguageLevel defaultValue) {
-      try {
-        return LanguageLevel.valueOf(level.toUpperCase(Locale.ROOT));
-      } catch(IllegalArgumentException e) {
-        // fall through
-      }
+    try {
+      return LanguageLevel.valueOf(level.toUpperCase(Locale.ROOT));
+    } catch (IllegalArgumentException e) {
+      // fall through
+    }
 
-      LanguageLevel languageLevel = LanguageLevel.parse(level);
-      if (languageLevel != null) {
-        return languageLevel;
-      }
-      return defaultValue;
+    LanguageLevel languageLevel = LanguageLevel.parse(level);
+    if (languageLevel != null) {
+      return languageLevel;
+    }
+    return defaultValue;
   }
 
   private static class JavaLanguageLevelParser extends ScalarSectionParser<LanguageLevel> {
@@ -64,7 +64,8 @@ public class JavaLanguageLevelSection {
 
     @Nullable
     @Override
-    protected LanguageLevel parseItem(ProjectViewParser parser, ParseContext parseContext, String rest) {
+    protected LanguageLevel parseItem(ProjectViewParser parser, ParseContext parseContext,
+        String rest) {
       LanguageLevel languageLevel = getLanguageLevel(rest, null);
       if (languageLevel == null) {
         // Fall through to error handler
--- a/java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
+++ b/java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
@@ -25,10 +25,12 @@ import com.google.idea.blaze.base.projectview.section.SectionKey;
 import com.google.idea.blaze.base.projectview.section.SectionParser;
 import com.intellij.pom.java.LanguageLevel;
 import javax.annotation.Nullable;
+import java.util.Locale;
 
 /** Section to force the java language level used */
 public class JavaLanguageLevelSection {
-  public static final SectionKey<Integer, ScalarSection<Integer>> KEY =
+
+  public static final SectionKey<LanguageLevel, ScalarSection<LanguageLevel>> KEY =
       SectionKey.of("java_language_level");
   public static final SectionParser PARSER = new JavaLanguageLevelParser();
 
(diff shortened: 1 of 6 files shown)


mai93:
Thank you @aarone! 

I tried building the plugin with this change but when I try it with [this project](https://github.com/bazelbuild/intellij/tree/master/examples/java/greetings_project) and the `.bazelproject` updated with 
`java_language_level: jdk_17_preview` I get this error on sync even though I've jdk17

```
Error:Unable to find a JDK 17 (Preview) - Pattern matching for switch installed.
After configuring a suitable JDK in the "Project Structure" dialog, sync the project again.
```

aarone:
hi @mai93  - thank you so much for reviewing this change!

I think this error occurs if the plugin can't find an installed JDK that provides the requested language level. I'm pretty sure [that behavior of how SDKs are discovered](https://github.com/bazelbuild/intellij/blob/4eb50ff5415c34493536373e87e7f2542ed4aea2/java/src/com/google/idea/blaze/java/sync/projectstructure/Jdks.java#L60-L98) by selected language level is unchanged by this change. For example, today, if you have no JDKs installed and set the java_language_level to '11' you'd get a similar error.

mai93:
Thanks @aarone.

Yes, I get that but although I've JDK 17 installed I still get this error when I use `java_language_level: jdk_17_preview`. I think the problem is in our use of `LanguageLevel$isAtLeast()` as based on the order of the versions in LanguageLevel class, [JDK17 is less than JDK17 preview](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-api/src/com/intellij/pom/java/LanguageLevel.java#L40:L41). Does this make sense? Or there is another reason for why I'm getting this error even though JDK 17 is installed.

aarone:
thanks for the info, i wasn't seeing the same error with newer jdks installed so didn't realize the nature of the issue; i'll look into it

aarone:
Alright, i've made changes that should address the issue; thanks for your help and patience.

java/tests/integrationtests/com/google/idea/blaze/java/sync/projectstructure/JdksTest.java
```diff
@@ -222,6 +222,42 @@ public void testChooseDifferentJdkIfNotProvidedByProvider() {
...
+    registerJdkProvider(
+        ImmutableMap.of(
+            LanguageLevel.JDK_1_8, jdk8,
+            LanguageLevel.JDK_11, jdk11
+        ));
+
+    setJdkTable(jdk8, jdk11);
+
+    Sdk chosenSdk = Jdks.chooseOrCreateJavaSdk(jdk11, LanguageLevel.JDK_17_PREVIEW);
+    // chosenSdk may be non-null when finding a local, system-installed sdk that provides the preview version
```
mai93, line 256: I tried this locally with jdk17 installed and [`createJdk`](https://github.com/bazelbuild/intellij/blob/d8d6ddcb34fb122a8f00c45c0927d023fe300def/java/src/com/google/idea/blaze/java/sync/projectstructure/Jdks.java#L103) failed with `Unable to create JDK from path '{PATH_TO_MY_JDK17}'`, maybe this is not allowed during tests. I think it will be better to not depend on the environment on which the test is run.

java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
```diff
@@ -25,51 +25,59 @@
...
-    LanguageLevel parsed = LanguageLevel.parse(level.toString());
-    return parsed != null ? parsed : defaultValue;
+  static LanguageLevel getLanguageLevel(String level, @Nullable LanguageLevel defaultValue) {
+    try {
+      return LanguageLevel.valueOf(level.toUpperCase(Locale.ROOT));
+    } catch (IllegalArgumentException e) {
+      // fall through
+    }
+
+    LanguageLevel languageLevel = LanguageLevel.parse(level);
```
mai93, line 53: IIUC this part is for supporting the old format of java_language_level, if so we need to check that `level` is an Integer before calling `LanguageLevel.parse()`.

java/tests/integrationtests/com/google/idea/blaze/java/sync/projectstructure/JdksTest.java
aarone: This test was trying to test the case when no jdk could be found but it looks like which can trigger lookup of a jdk on the local machine which is bad like you noted. I replaced this test with another test that choses an sdk based when multiple language levels are provided.

java/src/com/google/idea/blaze/java/sync/projectstructure/Jdks.java
```diff
@@ -122,16 +121,23 @@ private static boolean jdkPathMatches(Sdk jdk, String homePath) {
...
         .min(Comparator.comparing(Jdks::getJavaLanguageLevel))
         .orElse(null);
   }
 
+  private static boolean isLanguageLevelSupportedBySdkLevel(LanguageLevel requestedLanguageLevel,
+      LanguageLevel sdkLanguageLevel) {
+    if (sdkLanguageLevel == null) {
+      return false;
+    } else if (requestedLanguageLevel.isPreview()) {
+      return requestedLanguageLevel.toJavaVersion().equals(sdkLanguageLevel.toJavaVersion());
```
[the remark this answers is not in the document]
aarone: you right; that was my misunderstanding. I've updated some JDK selection methods to check if a JDK supports a given language level instead of assuming that a JDK has only a single language level.

java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
aarone: I can add this check if you'd like but `LanguageLevel.parse` returns null for invalid language levels and this is the same assumption in the existing implementation.

mai93: I also thought that using `LanguageLevel.parse` is enough. But I tried setting the language level to `INVALID_17_INVALID` for example and it was accepted as `JDK_17` that was surprising and that is why I want to force a separate validation here.

java/src/com/google/idea/blaze/java/sync/projectstructure/Jdks.java
```diff
@@ -122,16 +120,34 @@ private static boolean jdkPathMatches(Sdk jdk, String homePath) {
...
+        .filter(sdk -> isLanguageLevelSupportedBySdk(langLevel, sdk))
         .filter(Jdks::isValid)
         .min(Comparator.comparing(Jdks::getJavaLanguageLevel))
         .orElse(null);
   }
 
+
+  private static boolean isLanguageLevelSupportedBySdk(LanguageLevel requestedLanguageLevel,
+      Sdk sdk) {
+    JavaSdkVersion version = JavaSdk.getInstance().getVersion(sdk);
```
mai93, line 132: Can you please add a check that this a JavaSDK as 
```   
if (!(sdk.getSdkType() instanceof JavaSdk)) {
      return false;
}
```

java/src/com/google/idea/blaze/java/projectview/JavaLanguageLevelSection.java
aarone: i see, looking into the implementation of LanguageLevel.parse, it looks like it is pretty liberal in accepting version strings so adding this check will reject language levels that are currently accepted today in project view files. Seems like the right thing to do imo but just wanted to be clear about the implications.

mai93 (approved):

mai93:
Thank you @aarone!

Issue #2766, opened 2021-07-08
External Include Paths Point into Unstable Execroot Instead of Stable External Directory


cpsauer:
Hello wonderful Bazel folks,

**_TL;DR / summary:_** C++ header finding--and probably more--breaks for external libraries when you run a build command that uses a different set of external libraries. This issue can be fixed by having paths go through `<outputBase>/external/`, rather than the current `<outputBase>/execroot/__main__/external/`.

**_Discovery context:_** I'd noticed this after initially making the same mistake when writing a VSCode<>Clangd<>Bazel integration. I think this has the same root cause as https://github.com/bazelbuild/intellij/issues/1256 --a deeper fix here might allow removal of the code that patched it in https://github.com/bazelbuild/intellij/commit/81afc2cd1212828f33531d2705ce61172cfd237e. A similar problem has also caused issues in Tulsi.

**_What's going wrong in more detail:_**
When the plugin gets paths to external libraries from Bazel, they usually point into the execution root's external folder, the temporary sandbox of symlinks reconfigured for each build. This leads leads the plugin to point the IDE into the temporary sandbox for external header search paths (and other things, I'd assume). 

That's a trap. When you run a build that doesn't use the same set of external workspaces (like for example, a build for one of several targets in the project), external paths start breaking--and the red squiggles come out to indicate their displeasure. The temporary build sandbox, the execution root ("execroot") has gotten torn down and reconfigured, so some paths no longer exist. You can work around this by re-syncing, but it's annoying to have building break editing--feels like clearly undesirable behavior.

You can see that this problem is happening by opening, say, an NDK file in ASwB and looking at the header include paths. You can see the header include paths for a given file in Android Studio by "Show Compiler Info" (via Help→Find Action, aka CMD+SHIFT+A). You'll see paths containing `/private/var/tmp/_bazel_<username>/<hash>/execroot/__main__/external`, and verify that the symlinks those paths go through broken if you build some other part will get broken if a build that doesn't use the external workspace in question.

The fix: Those files live more durably inside `/private/var/tmp/_bazel_<username>/<hash>/external`, so the execroot-based paths should be rewritten to point there, just as we'd do for paths in the execroot that point into the normal workspace. That way you aren't relying on the symlinks in execroot that often get deleted out from under you!

Thanks for the work you do!
Chris
(ex-Googler)

P.S. This is parallel to a lingering Bug in Tulsi: https://github.com/bazelbuild/tulsi/issues/164 [Now resolved.]
[...]

PR #4063, opened 2022-11-01
Resolve external paths to their workspaces

cpsauer:
Please see https://github.com/bazelbuild/intellij/issues/2766 for context.

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/workspace/ExecutionRootPathResolver.java
+++ b/base/src/com/google/idea/blaze/base/sync/workspace/ExecutionRootPathResolver.java
@@ -25,14 +25,16 @@ import com.google.idea.blaze.base.settings.Blaze;
 import com.google.idea.blaze.base.sync.data.BlazeProjectDataManager;
 import com.intellij.openapi.project.Project;
 import java.io.File;
+import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
 /**
  * Converts execution-root-relative paths to absolute files with a minimum of file system calls
  * (typically none).
  *
- * <p>Files which exist both underneath the execution root and within the workspace will be resolved
- * to workspace paths.
+ * <p>Files which exist both underneath the execution root and within a workspace will be resolved
+  * to paths within their workspace. This prevents those paths from being broken when a different
+  * target is built.
  */
 public class ExecutionRootPathResolver {
 
@@ -76,13 +78,13 @@ public class ExecutionRootPathResolver {
     if (isInWorkspace(path)) {
       return workspacePathResolver.resolveToFile(path.getAbsoluteOrRelativeFile().getPath());
     }
-    return path.getFileRootedAt(executionRoot);
+    return convertExternalToStable(path.getFileRootedAt(executionRoot));
   }
 
   /**
    * This method should be used for directories. Returns all workspace files corresponding to the
-   * given execution-root-relative path. If the file does not exist inside the workspace (e.g. for
-   * blaze output files or external workspace files), returns the path rooted in the execution root.
+   * given execution-root-relative path. If the file does not exist inside a workspace (e.g. for
+   * blaze output files), returns the path rooted in the execution root.
    */
   public ImmutableList<File> resolveToIncludeDirectories(ExecutionRootPath path) {
     if (path.isAbsolute()) {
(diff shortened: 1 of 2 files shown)


base/src/com/google/idea/blaze/base/sync/workspace/ExecutionRootPathResolver.java
```diff
@@ -118,4 +120,24 @@ private static String getFirstPathComponent(String path) {
...
+  * That is, converts paths under <outputBase>/execroot/<workspace_name>/external/ to paths under
+  * <outputBase>/external/.
+  * Safe to call on all paths; non-external paths are left as they are.
+  * See https://github.com/bazelbuild/intellij/issues/2766 for more details.
+  */
+  private static File convertExternalToStable(File f) {
+    String regexSep = Pattern.quote(f.separator);
+    // Workspace name defaults to __main__ per DEFAULT_REPOSITORY_DIRECTORY in 
+    // https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/cmdline/LabelConstants.java
+    // Valid workspace name regex copied from LEGAL_WORKSPACE_NAME in
```
cpsauer, line 136: If you all have a good way of referencing Bazel source from within the IntelliJ plugin, please say something. Ideally, it'd be good to reference LEGAL_WORKSPACE_NAME instead of duplicating the regex.

mai93: I do not think there is a way to use `LEGAL_WORKSPACE_NAME` here.

cpsauer: 👍🏻 Sounds like we proceed then. Figured I should ask, though :)

cpsauer:
Gentle bump @tpasternak, since it's been a good while and y'all explicitly asked me to post this one?

cpsauer:
@tpasternak, that's definitely better. Done. Thanks for great feedback.
[The previous had just been a quick workaround to get folks back moving. Really appreciate your review and suggestions on how to improve it! Sorry for not being as quick as usual; got totally buried over the holidays.]

Some more minor notes:
- If small things get in the way of merging, feel free to just modify to save a RTT. You should have access.
- Had to break out the cases a little more finely for path resolution, so I simplified things by making it explicit that the main workspace was the fallback case. The actual case logic hasn't changed.
- Separable/less important improvement opportunities I saw while editing, but didn't do:
  - BlazeConfigurationResolver's call to the ExecutionRootPathResolver constructor could be a call to ExecutionRootPathResolver .fromProject, but it's unclear to me whether it should really be taking the workspace root from the parameter in the calling function instead. Either way, the prior code probably isn't optimal, but it isn't clear which way to chance it without more context than I have. (But which you might have!)
  - This preserves the duplicated logic between buildArtifactDirectories and resolveToIncludeDirectories, with some expanded documentation. They could maybe be merged into a single function, but I'm not sure you'll gain that much cleanliness because of the different return types and workspace implementations. Up to you; feel free to do if you want. 

Thanks!
Chris

base/src/com/google/idea/blaze/base/sync/workspace/ExecutionRootPathResolver.java
```diff
@@ -73,49 +78,55 @@ public File resolveExecutionRootPath(ExecutionRootPath path) {
...
       return path.getAbsoluteOrRelativeFile();
     }
-    if (isInWorkspace(path)) {
-      return workspacePathResolver.resolveToFile(path.getAbsoluteOrRelativeFile().getPath());
+    String firstPathComponent = getFirstPathComponent(path.getAbsoluteOrRelativeFile().getPath());
+    if (buildArtifactDirectories.contains(firstPathComponent)) {
+      // Build artifacts accumulate under the execution root, independent of symlink settings
+      return path.getFileRootedAt(executionRoot);
+    }
+    if (firstPathComponent.equals("external")) { // In external workspace
```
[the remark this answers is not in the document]
cpsauer: Hey hey! Thanks for taking another look, @tpasternak.

I'd think we'd need to modify execution root paths--and not just directories--because one might try to browse to an external file path. More generally, those execroot paths are unstable for all files, not just directories, so I'd think we'd always want to point to the stable paths, lest they disappear under the editor when a build gets run that uses different external workspaces. This ended up being important for Xcode/VSCode, anyway. I think this is a case we want to defend against. (Unless I'm misunderstanding your question; you all are the experts, immersed in this day-to-day.)

cpsauer:
Yay! Thank you, @tpasternak!

PR #4115, opened 2022-11-22
Make gazelle tests respect USE_BAZEL_VERSION

blorente:
# Checklist

@mai93 brought it to attention.

# Description of this change

Some downstream tests use the `USE_BAZEL_VERSION` env var to set up the Bazel executable. Normally, this would not be set up for tests, as most tests don't run Bazel themselves. However, the Gazelle integration tests do need to run Bazel, so we must set those global settings manually.


mai93 (approved):

mai93:
@blorente Unfortunately we are still getting errors with Bazel@Head: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2744#0184a2ac-9040-4254-a361-78cc4c1fb45f

I was thinking about mocking the build step, I think other sync tests do that https://github.com/bazelbuild/intellij/blob/master/base/tests/utils/integration/com/google/idea/blaze/base/sync/BlazeSyncIntegrationTestCase.java#L342.

PR #4127, opened 2022-11-24
[gazelle] Mock bazel in the integration tests

blorente:
# Checklist

Offline discussion with @mai93 

# Description of this change

## Problem

Some downstream tests are run with a non-standard bazel. While #4115  addressed that concern, it does rely on the particular bazel binary being in the machine.
However, sometimes these test are run in a remote execution environment, which means that bazel is not available.

## Solution

Instead of running gazelle on a full mock repository, we will mock `BlazeGazelleRunner` (now renamed to `GazelleRunner`), and make assertions on the bazel command it would have run in the real world.

This diminishes the guarantees, of the code, but it tests all the codepaths of the Gazelle integration itself. In addition to that, it makes the tests significantly faster and easier to maintain.

A caveat:
- We want to mock the gazelle runner without modifying the plugin code. To accomplish that, we leverage the services architecture, so that we implement the `GazelleRunner` interface with our mock.
  - This means that we need to build different plugin xmls for tests and regular code, as we need to select the right implementation of `GazelleRunner` each time.

### DIFF
--- a/gazelle/BUILD
+++ b/gazelle/BUILD
@@ -6,6 +6,7 @@ load(
     "//build_defs:build_defs.bzl",
     "intellij_plugin",
     "intellij_plugin_library",
+    "merged_plugin_xml",
     "optional_plugin_xml",
     "stamped_plugin_xml",
 )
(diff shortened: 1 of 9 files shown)


mai93:
Closing as https://github.com/bazelbuild/intellij/pull/4129 is merged.

PR #4129, opened 2022-11-24
Mock bazel run in Gazelle integration tests

mai93:
Co-authored-by: Borja Lorente <blorente@apple.com>

This is a simplified version of https://github.com/bazelbuild/intellij/pull/4127 that uses `registerApplicationService()` to register the mock `GazelleRunner` instead of using different plugin xml for tests.


gazelle/src/com/google/idea/blaze/gazelle/GazelleRunner.java
```diff
@@ -0,0 +1,76 @@
...
+ * The results may be cached in the workspace.
+ */
+public abstract class GazelleRunner {
+
+    public static GazelleRunner getInstance() {
+        return ApplicationManager.getApplication().getService(
+                GazelleRunner.class);
+    }
+
+    protected static BlazeCommand createGazelleRunCommand(
```
blorente, line 44: If we extract this function, we should use it in `GazelleRunnerImpl.runBlazeGazelle`.

blorente (approved):
One Small nit, otherwise looks good to me! Thanks for getting it over the line.

gazelle/src/com/google/idea/blaze/gazelle/GazelleRunner.java
mai93: I missed that, thank you!

mai93:
@tpasternak Can you please review this PR? we are still missing an approval to merge, probably because @blorente and I are co-authoring.

Issue #2042, opened 2020-08-14
Add support for setting Environment Variables via Run Configuration


KaoruDev:
This may be a naive request since I have no knowledge Intellij's Plugin API, but would it be possible to add a field in Bazel run configurations to be able to set ad-hoc environment variables for a given executable?

For example, the Java run configurations have this field available allowing for environment variables to be set per run configuration:

![image](https://user-images.githubusercontent.com/5273755/90256437-98013580-ddfa-11ea-805c-da9fb134e245.png)

Intellij docs capture the ability:
https://www.jetbrains.com/help/idea/run-debug-configuration-application.html

---

At the moment my current workaround is to add an `export VAR=value` to my shell dotfile, which is error-prone and difficult to manage 😭

HunterEl:
Also running into this and would love to see support for this through Intellij.

graysonchao:
I managed to work around this using the `--action_env=VARIABLE=value` form of the `--action_env` flag ([docs](https://docs.bazel.build/versions/master/command-line-reference.html#flag--action_env)) in the "Bazel flags" section. 

e.g.:

<img width="1068" alt="Screen Shot 2021-03-18 at 5 27 37 PM" src="https://user-images.githubusercontent.com/1198863/111714643-4f39b100-880f-11eb-8cae-10b2e4467cea.png">

ns476:
As a slightly grim workaround, you can use [`--run_under`](https://docs.bazel.build/versions/master/user-manual.html#flag--run_under) like so:

![Screenshot from 2021-04-22 14-16-13](https://user-images.githubusercontent.com/632493/115720837-535c6f80-a375-11eb-847a-84a43ec63cdd.png)

graysonchao:
@ns476 That's still way better than launching intellij from the CLI with env vars set, which is what I was doing. Thanks!

rodrigodiez:
@ns476 your approach works for me but unfortunately can't debug with it

```
could not launch process: not an executable file
```

JamyDev:
[...]

PR #4138, opened 2022-11-28
feat(ijwb): add environment variables to the run configuration

JamyDev:
# Checklist

details.

Issue number: #2042

# Description of this change

This adds the option to set environment variables to your run configuration for any target type. For `Test` configurations `--test_env` should be used and thus this field is only shown when set to `Run`.

Reused the `EnvironmentVariablesData` that was used in the Python module and applied it everywhere by passing it as an argument to `ScopedBlazeProcessHandler`, which then passes it onto the ProcessGroup using `.withEnvironment`.

Since `Go` handles the Run Configuration differently, we make sure to add the environment variables there as well, in this case as the last one so the user can override variables set elsewhere.

Screenshots:

env_vars_new_1
(this was a go_binary target but I had to modify it for reasons, so it changed it to generic handler)
env_vars_new_2
env_vars_new_3
env_vars_new_py

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
+++ b/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
@@ -296,6 +296,7 @@ public final class BlazeCommandGenericRunConfigurationRunner
       return new ScopedBlazeProcessHandler(
           project,
           blazeCommand,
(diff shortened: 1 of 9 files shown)


mai93:
Thanks @JamyDev!

I also was not able to see the environment variables for `java_binary` target run configuration, I'm trying with this [example project](https://github.com/bazelbuild/intellij/tree/master/examples/java/greetings_project).

JamyDev:
Thanks for the comments, I'll address them sometime soon.

PR #4180, opened 2022-12-09
Extension point to configure the way BlazePyRunConfigurationRunner selects the executable file

odisseus:
# Checklist

details.

# Description of this change


python/src/com/google/idea/blaze/python/run/BlazePyRunConfigurationRunner.java
```diff
@@ -386,14 +389,8 @@ private static ImmutableList<String> getPythonArgsFor(
...
-    if (outputs.size() == 1) {
-      return outputs.get(0);
-    }
-    String name = PathUtil.getFileName(target.targetName().toString());
-    for (File file : outputs) {
-      if (file.getName().equals(name)) {
-        return file;
-      }
+    for (BlazePyExecutableOutputFinder outputFinder : EP_NAME.getExtensionList()) {
+      return outputFinder.findExecutableOutput(target, outputs);
```
lukaszwawrzyk, line 393: Shouldn't it for example return only if result is not null?

Also I am not sure if I understand how the current setup works. Is somehow BlazePyExecutableOutputFinder in the list of extensions? Shouldn't BlazePyExecutableOutputFinder be an empty interface and one implementation would be added through xml (with this default implementation)?

lukaszwawrzyk (approved):

Issue #1341, opened 2019-11-04
Cannot run program "glogin" exception


ittaiz:
Running a plugin built from [source](https://github.com/bazelbuild/intellij/commit/2fb993316fc7708d6a02b75f4a90d81393fc5be2) causes the following exception in the log: 
```java
java.io.IOException: Cannot run program "glogin" (in directory "/"): error=2, No such file or directory
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at com.google.idea.blaze.base.async.process.ExternalTask$ExternalTaskImpl.invokeCommand(ExternalTask.java:250)
	at com.google.idea.blaze.base.async.process.ExternalTask$ExternalTaskImpl.lambda$run$0(ExternalTask.java:179)
	at com.google.idea.blaze.base.scope.Scope.push(Scope.java:40)
	at com.google.idea.blaze.base.async.process.ExternalTask$ExternalTaskImpl.run(ExternalTask.java:172)
	at com.google.idea.blaze.base.actions.FileGitHubIssueAction.lambda$static$0(FileGitHubIssueAction.java:65)
	at com.intellij.openapi.util.NotNullLazyValue$2.compute(NotNullLazyValue.java:67)
	at com.intellij.openapi.util.NotNullLazyValue.getValue(NotNullLazyValue.java:38)
	at com.google.idea.blaze.base.actions.FileGitHubIssueAction.updateForBlazeProject(FileGitHubIssueAction.java:74)
	at com.google.idea.blaze.base.actions.BlazeProjectAction.update(BlazeProjectAction.java:58)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareUpdate(ActionUtil.java:176)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.doUpdate(ActionUpdater.java:378)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.lambda$null$1(ActionUpdater.java:78)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.callAction(ActionUpdater.java:106)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.lambda$new$2(ActionUpdater.java:79)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.update(ActionUpdater.java:364)
	at com.intellij.openapi.actionSystem.impl.ActionUpdater.expandGroupChild(ActionUpdater.java:231)
[...]

PR #4239, opened 2022-12-29
Check whether the glogin command exists before running it.

odisseus:
# Checklist

details.

Issue number: #1341 

# Description of this change

The attempt to run `glogin` is expected to fail for all users outside the Google's infrastructure. This causes a stack trace which litters the idea.log and makes it harder to look for any actual problems.

### DIFF
--- a/base/src/com/google/idea/blaze/base/actions/FileGitHubIssueAction.java
+++ b/base/src/com/google/idea/blaze/base/actions/FileGitHubIssueAction.java
@@ -62,7 +62,11 @@ public final class FileGitHubIssueAction extends BlazeProjectAction {
   private static final NotNullLazyValue<Boolean> isCorpMachine =
       NotNullLazyValue.createValue(
           () -> {
-            int retVal = ExternalTask.builder().args("glogin", "-version").build().run();
+            String which = SystemInfo.isWindows ? "where" : "which";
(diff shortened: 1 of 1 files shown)


base/src/com/google/idea/blaze/base/actions/FileGitHubIssueAction.java
```diff
@@ -62,7 +62,10 @@ public final class FileGitHubIssueAction extends BlazeProjectAction {
   private static final NotNullLazyValue<Boolean> isCorpMachine =
       NotNullLazyValue.createValue(
           () -> {
-            int retVal = ExternalTask.builder().args("glogin", "-version").build().run();
+            int retVal = ExternalTask.builder().args("which", "glogin").build().run();
```
jastice, line 65: I think this will break on Windows, or at least throw an error - which we're trying to avoid. Maybe it's expected to return false on Windows anyway, but we should check.

odisseus: Never mind. It turns out that Bazel has some support for Windows. I have changed my code accordingly.

jastice (approved):
LGTM, maybe @mai93 can check if it works as intended in Google still?

mai93:
I checked it, thank you @odisseus and @jastice!

PR #4242, opened 2022-12-29
Stop allowing `ScalaTestContextProvider` to deal with targets of `scala_junit_test` kind

azdrojowa123:
I found that `ScalaTestContextProvider` allows to add special scalatest flags (which were introduced in [3744](https://github.com/bazelbuild/intellij/pull/3744)) for runner even if test target is of `scala_junit_test` kind. It introduces a problem for such targets - `--test_filter` flag is not added so tests are not filtered in any way (`-s` and `-t` flags are not respected by JUnit runner)

This change exclude `scala_junit_test` targets  from creating their context in the `ScalaTestContextProvider`, so they are treated as they were before [3744](https://github.com/bazelbuild/intellij/pull/3744).

### DIFF
--- a/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
+++ b/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
@@ -30,7 +30,7 @@ import com.google.idea.blaze.java.run.producers.TestSizeFinder;
 import com.intellij.execution.JavaExecutionUtil;
 import com.intellij.execution.Location;
 import com.intellij.execution.actions.ConfigurationContext;
-import java.util.Optional;
+import java.util.function.Supplier;
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScTypeDefinition;
 import org.jetbrains.plugins.scala.testingSupport.test.scalatest.ScalaTestConfigurationProducer;
 import org.jetbrains.plugins.scala.testingSupport.test.scalatest.ScalaTestTestFramework;
@@ -64,23 +64,24 @@ class ScalaTestContextProvider implements TestContextProvider {
     ListenableFuture<TargetInfo> target =
         TestTargetHeuristic.targetFutureForPsiElement(
             testClass, TestSizeFinder.getTestSize(testClass));
-    TestContext.BlazeFlagsModification flags = null;
     if (target == null) {
       return null;
-    } else {
+    }
+
+    Supplier<TestContext.BlazeFlagsModification> testFlags = () -> {
       try {
         if(target.get(2, TimeUnit.SECONDS).kindString.equals("scala_junit_test")) {
-          flags = TestContext.BlazeFlagsModification.testFilter(ProducerUtils.getTestFilterForClass(testClass));
+          return TestContext.BlazeFlagsModification.testFilter(ProducerUtils.getTestFilterForClass(classAndTestNames.testClass));
         }
       } catch (InterruptedException | ExecutionException | TimeoutException e){
         //ignore
       }
-    }
+      return new ScalatestTestSelectorFlagsModification(classAndTestNames.testClass.getQualifiedName(), classAndTestNames.testNames);
+    };
 
     return TestContext.builder(testClass, ExecutorType.DEBUG_SUPPORTED_TYPES)
         .setTarget(target)
-        .addBlazeFlagsModification(Optional.ofNullable(flags)
-                .orElse(new ScalatestTestSelectorFlagsModification(testClass.getQualifiedName(), classAndTestNames.testNames)))
+        .addBlazeFlagsModification(testFlags.get())
         .setDescription(testClass.getName())
         .build();
   }
--- a/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
+++ b/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
@@ -25,10 +25,12 @@ import com.google.idea.blaze.base.run.producers.TestContext;
 import com.google.idea.blaze.base.run.producers.TestContextProvider;
 import com.google.idea.blaze.base.run.smrunner.SmRunnerUtils;
 import com.google.idea.blaze.base.run.state.RunConfigurationFlagsState;
+import com.google.idea.blaze.java.run.producers.ProducerUtils;
 import com.google.idea.blaze.java.run.producers.TestSizeFinder;
 import com.intellij.execution.JavaExecutionUtil;
 import com.intellij.execution.Location;
 import com.intellij.execution.actions.ConfigurationContext;
+import java.util.Optional;
 import org.jetbrains.plugins.scala.lang.psi.api.toplevel.typedef.ScTypeDefinition;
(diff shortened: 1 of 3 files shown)


scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -61,6 +64,14 @@ public RunConfigurationContext getTestContext(ConfigurationContext context) {
             testClass, TestSizeFinder.getTestSize(testClass));
     if (target == null) {
       return null;
+    } else {
+      try {
+        if(target.get(2, TimeUnit.SECONDS).kindString.equals("scala_junit_test")) {
```
lukaszwawrzyk, line 69: Please check if this .get or .kindString can return null, and if they can, check it.

azdrojowa123: as I checked `.get` can only throw exceptions and `.kindString` is not marked @Nullable

azdrojowa123:
> This still does not cover case if there is a scala_test target, but the class is annotated `@RunWith(classOf[JUnitRunner])`

When target is of kind scala_test and test class is annotated with `@RunWith(classOf[JUnitRunner])` it does not cause conflict (still scalatest runner is used inside).

srdo-humio:
Most of the "is this a test class this provider should handle" code is the same as before, but I noticed that this bit was removed:

```
  private static boolean isTestClass(ScClass testClass) {
    TestFramework framework = TestFrameworks.detectFramework(testClass);
    return framework instanceof ScalaTestTestFramework && framework.isTestClass(testClass);
  }
```

The provider returned null if this test failed. I'm wondering if adding this check back would fix this issue? I think we could just add a filter step here 

https://github.com/bazelbuild/intellij/blob/ba507e8f5b3b647774793254013ffa2ce5f9500b/scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java#L103

that returns null if the `classAndTestNames.testClass` is not an `ScClass` or `isTestClass` returns false?

srdo-humio:
Yes, but I meant as an alternative to looking at the target kindString for detecting whether the test class is a Scalatest class.

Regarding using `@RunWith`, isn't the test still being run via the Scalatest command line runner in that case, and then the Scalatest runner is handing off to JUnit at some point internally? 

The concern about which flags are set should only apply when `rules_scala` is actually invoking a different runner on the command line (as it would when using `scala_junit_test`), which I don't think happens when you use `scala_test` and add a `@RunWith` to your class. My best guess is `rules_scala` still invokes the regular Scalatest runner with the `-s -t` flags, and then that runner hands off to the JUnit runner at some point with whatever config it thinks is relevant.

Note that I'm just guessing here, I don't have anything set up to check this. Did you see `@RunWith` tests break from the change I did?

azdrojowa123:
> So I've tried it and can't reproduce. @azdrojowa123 could you provide a minimal, reproducible example of this problem?

Of course, here is such a reproduction [scalatest_junit_repro]( https://github.com/azdrojowa123/scalatest_junit_repro)

azdrojowa123:
> The provider returned null if this test failed. I'm wondering if adding this check back would fix this issue? I think we could just add a filter step here

I tried your solution and I added back `isTestClass` method and it doesn't work in my reproduction - `TestFrameworks.detectFramework(testClass)` return `ScalaTestTestFramework` :(

srdo-humio:
Thank you for investigating, sorry I think I misunderstood your issue.

Prior to my change, the scalatest provider would set `--test_filter=com.example.TestClass1` for your example. The flag is passed to https://github.com/srdo-humio/rules_scala/blob/23bc243126dc7948e3d5411f095c74de2a9ed9b6/src/java/io/bazel/rulesscala/scala_test/Runner.java for regular Scalatest tests. I updated the provider to instead pass several Scalatest-specific `--test_arg` flags to Bazel, which then get passed through to Scalatest via the runner.

I think the reason the old behavior also worked for junit tests before is that passing `--test_filter=com.example.TestClass1` also happens to be accepted by https://github.com/bazelbuild/bazel/blob/master/src/java_tools/junitrunner/java/com/google/testing/junit/runner/BazelTestRunner.java, which is the runner invoked when using `scala_junit_test`.

After applying the change from this PR, I see that the command line changes to `--test_filter=com.example.TestClass1#`. I figure this is because some other test provider in the IJ plugin also happens to be able to extract the relevant test class name.

@azdrojowa123 I think your change seems fine. I'm wondering though if it would be better to ensure that the scalatest provider handles these tests, instead of relying on whichever provider it is that happens to catch the test when the scalatest provider returns null?

So instead of returning null here https://github.com/bazelbuild/intellij/pull/4242/files#diff-f78adffc6524eb9e6e20bacab71d8bbea8b46ed0d16b60b421ecd064d3f2a122R70

you would change this code

https://github.com/bazelbuild/intellij/pull/3744/files#diff-f78adffc6524eb9e6e20bacab71d8bbea8b46ed0d16b60b421ecd064d3f2a122R66-R71

so it sets the test_filter if the rule is `scala_junit_test`, and otherwise it sets the selector flags I added?

If you'd rather not, then I think this is probably fine as is too.

azdrojowa123:
> 
Thank you for your answer. I think it might be a good idea - in this case when `null` is returned, flags are generated by `JavaTestContextProvider` which is probably not entirely correct (although it works)

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -59,13 +64,23 @@ public RunConfigurationContext getTestContext(ConfigurationContext context) {
     ListenableFuture<TargetInfo> target =
         TestTargetHeuristic.targetFutureForPsiElement(
             testClass, TestSizeFinder.getTestSize(testClass));
+    TestContext.BlazeFlagsModification flags = null;
     if (target == null) {
       return null;
+    } else {
```
srdo-humio, line 70: Nitpicking: You don't need the "else" here, since the if branch returns.

scala/src/com/google/idea/blaze/scala/run/producers/ScalaTestContextProvider.java
```diff
@@ -59,13 +64,23 @@ public RunConfigurationContext getTestContext(ConfigurationContext context) {
...
+      try {
+        if(target.get(2, TimeUnit.SECONDS).kindString.equals("scala_junit_test")) {
+          flags = TestContext.BlazeFlagsModification.testFilter(ProducerUtils.getTestFilterForClass(testClass));
```
srdo-humio, line 73: Nitpicking: I think this could be structured in a slightly more straightforward way as

```
TestContext.BlazeFlagsModification flags = getFlags();

private TestContext.BlazeFlagsModification getFlags() {
try {
  if (is junit) {
    return TestContext.BlazeFlagsModification.testFilter(ProducerUtils.getTestFilterForClass(testClass));
  }
} catch {
  //ignore
}

return new ScalatestTestSelectorFlagsModification(testClass.getQualifiedName(), classAndTestNames.testNames);
}
```

so you don't have to figure out which flags to use based on null a few lines further down.

srdo-humio (commented):
Looks good to me

raoarun:
@tpasternak @mai93 @jastice is this ready for review / merge?  thx.

PR #4274, opened 2023-01-11
Remove `is_new_external_version` temporary field

tpasternak:
# Checklist

details.

# Description of this change
This temporary field was used to cover old bazel behavior, which is not the case now.

By the way there was a bug here because `toProto` was not setting the "is_new_external_version" field so if an already-serialized data was deserialized again, the fixProto method was applied again

### DIFF
--- a/base/tests/unittests/com/google/idea/blaze/base/sync/workspace/ArtifactLocationDecoderTest.java
+++ b/base/tests/unittests/com/google/idea/blaze/base/sync/workspace/ArtifactLocationDecoderTest.java
@@ -58,66 +58,6 @@ public class ArtifactLocationDecoderTest extends BlazeTestCase {
   }
 
   @Test
-  public void testExternalSourceArtifactOldFormat() {
-    ArtifactLocation artifactLocation =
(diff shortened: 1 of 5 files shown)


base/tests/unittests/com/google/idea/blaze/base/sync/workspace/ArtifactLocationDecoderTest.java
```diff
@@ -57,66 +57,6 @@ public void testGeneratedArtifact() {
...
-            null,
-            RemoteOutputArtifacts.EMPTY);
-
-    assertThat(decoder.decode(artifactLocation).getPath())
-        .isEqualTo(
-            EXECUTION_ROOT + "/blaze-out/crosstool/bin/external/repo_name/com/google/Bla.java");
-  }
-
   @Test
   public void testExternalSourceArtifactNewFormat() {
```
mai93, line 61: nit: can you remove `NewFormat` from the test name here and in other test methods?

mai93 (approved):

mai93 (approved):

PR #4280, opened 2023-01-13
Add python example with proto

blorente:
Based on work by @yoyomolinas

Currently, third-party symbols do not resolve correctly for Python. Hopefully this will be a way to test that they do in the future.

# Checklist

details.

# Description of this change

### DIFF
--- a/examples/python/with_proto/library/BUILD.bazel
+++ b/examples/python/with_proto/library/BUILD.bazel
@@ -1,0 +1,30 @@
+load("@python_deps//:requirements.bzl", "requirement")
+load("@rules_python//python:defs.bzl", "py_binary", "py_library")
+
+py_library(
+    name = "utils",
+    srcs = ["utils.py"],
+)
+
+py_binary(
+    name = "server",
+    srcs = ["server.py"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//proto:python_lib",
+        requirement("grpcio"),
+        requirement("grpcio-tools"),
+    ],
+)
+
+py_binary(
+    name = "client",
+    srcs = ["client.py"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//library:utils",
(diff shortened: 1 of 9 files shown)


examples/python/with_proto/library/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
...
+        requirement("grpcio-tools"),
+    ],
+)
+
+py_binary(
+    name = "client",
+    srcs = ["client.py"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//library:utils",
```
mai93, line 25: I think using `":utils"` is easier to follow

examples/python/with_proto/library/client.py
```diff
@@ -0,0 +1,39 @@
+import grpc
+
+import proto.hello_pb2_grpc as pb2_grpc
+import proto.hello_pb2 as pb2
+
+from library import utils
+
+class Client(object):
+    """
+    Client to test gRPC functionality. Every symbol from `pb2` and `pb2_grpc` should resolve correctly,
```
mai93, line 10: but currently these references do not resolve, right?

examples/python/with_proto/library/client.py
```diff
@@ -0,0 +1,39 @@
...
+        self.server_port = 50051
+
+        # instantiate a channel
+        self.channel = grpc.insecure_channel(
+            '{}:{}'.format(self.host, self.server_port))
+
+        # bind the client and the server
+        self.stub = pb2_grpc.HelloServiceStub(self.channel)
+
+    def get_url(self, message):
```
mai93, line 25: can we rename this method to something like `send_request`?

mai93:
@tpasternak if you have some time, can you also check this example project I think you have more python experience?

mai93:
@tpasternak I asked and it seems to be safe to have these dependencies especially in an example project.

Issue #4112, opened 2022-11-17
Variables are not resolved when run Run Configuration without debugging


vicpl:
### Description of the bug:

 - Create a Run Configuration for bazel binary target
 - Specify Executable flags using variables, for example:
   --envfile=$WorkspaceRoot$/current.env

 - Run configuration with debugger - "$WorkspaceRoot$" is resolved into an actual path
 - Run configuration without debugger - "$WorkspaceRoot$" is not resolved and application receives "$WorkspaceRoot$/current.env" as an argumen

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

See above

### Which Intellij IDE are you using? Please provide the specific version.

IntelliJ IDEA 2022.2.3 (Ultimate Edition)

### What programming languages and tools are you using? Please provide specific versions.

go

### What Bazel plugin version are you using?

2022.10.18.0.1-api-version-222

sgowroji:
Hello @vicpl, Can you provide an sample code to reproduce the above issue. Thanks!

vicpl:
Hey @sgowroji,

Thanks for looking into this. Here is a sample repo: https://github.com/vicpl/bazel-4114

vicpl:
Is there any update on this? This issue is a blocker for us to migrate from vs code.

blorente:
@vicpl Please check if this was solved.

blorente:
Closing, feel free to reopen if the issue wasn't fixed.

PR #4368, opened 2023-02-01
fix(run): Properly expand IntelliJ macros in run configurations

blorente:
# Checklist

Issue number: #4112

# Description of this change

IntelliJ makes available several macros for flag values ([link](https://www.jetbrains.com/help/objc/built-in-macros.html)). In addition to that, we define our own for `$WorkspaceRoot$` ([link](https://github.com/bazelbuild/intellij/blob/f08795f8eb18da91d88be13f91b29cb0681dce8c/base/src/META-INF/blaze-base.xml#L297)). However, the expansion of these macros is currently broken for all languages.

The original issue denotes how Debug runners in Go happened to work, but this is due to the underlying Go plugin expanding the macros independently.

This patch replaces the current, broken mechanism to expand macros with a new one.

Tests are only in an example project, because they require actually running the run configuration to see the changes.


examples/go/with_proto/run_configurations/test_macro_expansion.xml
```diff
@@ -0,0 +1 @@
+<configuration name="Bazel test go:go_test" type="BlazeCommandRunConfigurationType" factoryName="Bazel Command" nameIsGenerated="true"><blaze-settings handler-id="BlazeGoRunConfigurationHandlerProvider" blaze-command="test" kind="go_test"><blaze-target>//go:go_test</blaze-target><blaze-user-flag>--test_env=WORKSPACE_ROOT=$WorkspaceRoot$</blaze-user-flag><blaze-user-flag>--test_env=FILE_NAME=$FileName$</blaze-user-flag></blaze-settings><method v="2"><option name="Blaze.BeforeRunTask" enabled="true" /></method></configuration>
```
mai93, line 1: I think we can add (--test_filter=^TestFlagMacros$ --test_env=GO_TEST_WRAP_TESTV=1) blaze flags for the test results to be viewed correctly in the IDE.

examples/go/with_proto/go/main_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+func TestFlagMacros(t *testing.T) {
+	testCases := []struct {
+		envVar     string
+		validation func(string) bool
+		error      string
+	}{
+		{
+			envVar:     "FILE_NAME",
+			validation: func(expanded string) bool { return expanded == "main_test.go" },
```
mai93, line 16: Because `$FileName$` requires the `main_test.go` file to be the active file for the test to pass, it can be confusing. I think we can avoid this by testing on something like $ProjectName$.

blorente: Good point! Changed.

examples/go/with_proto/run_configurations/test_macro_expansion.xml
blorente: Done! Thanks for the suggestion.

mai93 (approved):

PR #4412, opened 2023-02-09
Add ignore for other Jetbrains directories on Bazel sync

victoriaroseb:
# Checklist

details.

# Description of this change

We are excluding other automatically created JetBrains folders that are not relevant depending on the IDE we are currently working in, allowing multiple IDEs to coexist in a Bazel project.


base/src/com/google/idea/blaze/base/sync/data/BlazeDataStorage.java
```diff
@@ -25,13 +25,15 @@
...
 import java.io.File;
+import java.util.HashMap;
 import java.util.Optional;
 
 /** Defines where we store our blaze project data. */
 public class BlazeDataStorage {
   public static final String WORKSPACE_MODULE_NAME = ".workspace";
   public static final String BLAZE_DATA_SUBDIRECTORY = ".blaze";
   public static final String PROJECT_DATA_SUBDIRECTORY = getProjectDataSubdirectory();
+  public static final HashMap<String, String> ALL_PROJECT_SUBDIRECTORIES = getAllProjectSubdirectories();
```
mai93, line 36: Can we use a set of directories names here instead of the map?

mai93:
@victoriaroseb Can you please check the test failures? For example: https://buildkite.com/bazel/intellij-ue-plugin/builds/7967

victoriaroseb:
> @victoriaroseb Can you please check the test failures? For example: https://buildkite.com/bazel/intellij-ue-plugin/builds/7967

Fixed thanks!

base/src/com/google/idea/blaze/base/sync/data/BlazeDataStorage.java
```diff
@@ -25,13 +25,17 @@
 import com.intellij.openapi.application.PathManager;
 import com.intellij.openapi.project.Project;
 import java.io.File;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Set;
```
mai93, line 31: Please remove this import and the`import java.util.HashMap;` as they are not needed.

base/src/com/google/idea/blaze/base/sync/projectstructure/ContentEntryEditor.java
```diff
@@ -68,6 +70,18 @@ public static void createContentEntries(
         contentEntry.addExcludeFolder(UrlUtil.fileToIdeaUrl(excludeFolder));
       }
 
+      File directory = new File(workspaceRoot.toString());
```
mai93, line 73: I looked more into this and I think it is not needed to add this part inside the loop over `rootDirectories` because IIUC it only needs the `workspaceRoot`. I think it will be better if we removed this from here and instead modified [`ImportRoots$excludeProjectDataSubDirectory()`](https://github.com/bazelbuild/intellij/blob/master/base/src/com/google/idea/blaze/base/sync/projectview/ImportRoots.java#L141:L143) to be 
```
    private void excludeProjectDataSubDirectories() {
      excludeDirectoriesBuilder.addAll(
          BlazeDataStorage.ALL_PROJECT_SUBDIRECTORIES.stream()
              .map(d -> new WorkspacePath(d))
              .collect(toImmutableList()));
    }
```

I think this should be enough to add all the possible project subdirectories to the excluded list. If you think the current solution will still be needed, please let me know. Also this may require modifying some of the tests.

raoarun:
@mai93 what is blocking review/merge?

mai93:
> @mai93 what is blocking review/merge?

Unaddressed review comment: https://github.com/bazelbuild/intellij/pull/4412#discussion_r1121214561

Issue #4433, opened 2023-02-13
JUnit 5 Support


blorente:
### Description of the feature request:

Bazel 6 supports JUnit 5 for Java tests. Furthermore, some third party rulesets (e.g. [`contrib_rules_jvm`](https://github.com/bazel-contrib/rules_jvm) already offer facilities to allow it.

As far as I can tell, the plugin does not yet offer support for these tests.
I've given the implementation a shot in this branch: https://github.com/blorente/intellij/tree/blorente/junit5 (leveraging the existing testbed: [commit](https://github.com/blorente/intellij/commit/ab2ade87085d516fd3c9131a859600ee36893416), but it looks like it will need at least a few days more of investment.

### What underlying problem are you trying to solve with this feature?

Currently, the plugin accidentally supports some of its features. As far as I can tell, I can:
- Right-click run a test, and most of the time it will be run correctly (but with the JUnit4 runner)
- Import JUnit 5 as a library, and get autocompletion and symbol resolution.

The plugin does not, however:
- Recognize a file as JUnit 5 for the purposes of gutter annotation:

JUnit 4:
<img width="363" alt="Screenshot 2023-02-13 at 13 33 23" src="https://user-images.githubusercontent.com/5861182/218471823-2c12aec2-0bc1-4bcd-9745-f80dcf7fbb5e.png">

JUnit 5 (missing the icons):
<img width="348" alt="Screenshot 2023-02-13 at 13 33 14" src="https://user-images.githubusercontent.com/5861182/218471981-0f34fa22-546c-4f0d-8b01-6110e42461b9.png">

- I suspect (but haven't validated yet) that it doesn't support a JUnit 5 runner either.
- And looking through the code, I found a lot of references to JUnit 4 and 3 and none to JUnit 5, which makes me believe that any JUnit 5 support is provided by IntelliJ itself and not the plugin.

### What operating system, Intellij IDE and programming languages are you using? Please provide specific versions.

I use an arm macOS. This reproduces in IntelliJ UE 2022.2 and 2022.3, on the plugin's `master` branch.

### Have you found anything relevant by searching the web?

- https://github.com/bazelbuild/intellij/issues/245 -> Closed for being old.
- https://github.com/bazelbuild/intellij/issues/3413 -> Outlines that some JUnit 5 features work.

### Any other information, logs, or outputs that you want to share?

I'm happy to work on this. The main reason for the ticket is to make sure nobody is looking into it before sinking significant time.

[...]

PR #4473, opened 2023-02-20
Implement JUnit5 support and parameterize tests

blorente:
# Checklist

  - Well, this has some refactoring, mostly of the test code to make it easier to parameterize.

details.

Issue number: #4433

# Description of this change

Luckily, JUnit 4 test filters are compatible with JUnit 5.

Therefore, the only things that this PR needs to do are:
- Change `JUnitTestHeuristic` to account for JUnit 5.
- Change the test filter generators to account for JUnit 5 as well.
- Change the testbed so that it also tests against JUnit 5 code:
  - Introduce `JUnitTestUtils`, a utility class to set up test repositories with JUnit 5 or 4 code.
  - Introduce `BlazeJUnitRunConfigurationProducerTestCase`, to remove parameterization boilerplate for the Run Configuration Producer tests.
  - Parameterize relevant integration tests (either with the base class mentioned above, or by hand), to test a set of JUnit versions (currently, 4 and 5). This mostly affects the set up of the code under tests, so we can reuse a lot of the test logic.
  - Introduce JUnit 5 to the java example project.

- Implement fastbuild support for `java_junit5_test`:
  - Add the `java_junit5_test` as an allowed target kind for FastBuild.
  - Capture the `main_class` attribute in the FastBuild aspect proto.
  - Prefer `main_class` when building run configurations over the default bazel runner (`java_junit5_test` uses a custom test runner).


raoarun:
@mai93 please review.

blorente:
Rebased on top of `master` and implemented https://github.com/bazelbuild/intellij/issues/5550 in the last two commits.

kermitnirmit:
Hi - I'm super interesting in this because I use junit5 and my parameterized tests don't look like this: 
image
They are all on the top level instead. I would like this hierarchy.

java/tests/unittests/com/google/idea/blaze/java/run/producers/BlazeJUnitTestFilterFlagsTest.java
```diff
@@ -63,6 +71,25 @@ public void testParameterizedIgnoredForSingleClass() {
         .isEqualTo("com.google.idea.ClassName#");
   }
 
+  @Test
+  public void testParameterizedIgnoredForSingleClassJUnit5() {
+    assertThat(
```
mai93, line 76: what is the difference between this test and `testSingleJUnit5ClassFilter`?

java/tests/integrationtests/com/google/idea/blaze/java/run/producers/MultipleJavaClassesTestContextProviderTest.java
```diff
@@ -71,63 +77,32 @@ public final void addSourceFolder() {
...
   @After
   public final void removeSourceFolder() {
+    VirtualFile pkgRoot = workspace.createDirectory(pkgRootPath);
     ApplicationManager.getApplication()
         .runWriteAction(
             () -> {
               final ModifiableRootModel model =
                   ModuleRootManager.getInstance(testFixture.getModule()).getModifiableModel();
               ContentEntry contentEntry = model.getContentEntries()[0];
-              contentEntry.clearSourceFolders();
```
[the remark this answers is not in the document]
blorente: Good question, it took me a while to find the answer 😓 .
I've left it in a comment, but the reason is:

```
// There is a global content entry, `//src`, which is necessary to infer targets from test classes.
// This is set outside of these test (somewhere in the superclass) once per run, not once per test.
// If we just use contentEntry.clearSourceDirs() here, it won't be re-created, and tests that assume it exists will fail.
// Therefore, we only delete the content entries that this test has created.
```

java/tests/unittests/com/google/idea/blaze/java/run/producers/BlazeJUnitTestFilterFlagsTest.java
blorente: You're absolutely right, there's no difference. That means the test above it (the same, but for junit4) is broken as well. Will fix them both and make them meaningful.

blorente:
Addressed the comments, sorry for the delay but it was hard to find time, as per usual.

agluszak (approved):

PR #4517, opened 2023-02-28
Adds not stale label on copybara-service[bot] PR's

sgowroji:
# Description of this change
This Github action will add "not stale" label to all the copybara-service[bot] PR's. So that they are exempted from stale Github action workflow.

### DIFF
--- a/.github/workflows/notstale.yml
+++ b/.github/workflows/notstale.yml
@@ -1,0 +1,43 @@
+# This workflow adds not stale label to only copybara[bot] PRs to avoid stale workflow 
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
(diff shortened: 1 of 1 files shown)


jastice (approved):

.github/workflows/notstale.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  label-not-stale-prs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR author
+        id: pr-author-check
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
mai93, line 20: @keertk I think we will have the same problem providing the github-token here, or is this different?

Issue #4546, opened 2023-03-04
java_binary rules labeled as manual disables the Debug icon


rogerhu:
### Description of the bug:

Suppose I have this Java rule:

```
java_binary(
    name = "shaded_jar",
    main_class = "com.squareup.MyApp",
    tags = ["manual"],
    runtime_deps = ["//myapp/src/main/java:lib"],
)
```

Because it is added with a `manual` tag, the IDE information (e.g.`intellij-info.txt`) will not be generated for this rule (confirmed by looking inside the `bazel-out` directory). There is a specific tag that excludes it [here](https://github.com/bazelbuild/intellij/blob/94b6618bffdc91ddb0b6f773fb99e1cb5f928cdc/base/src/com/google/idea/blaze/base/dependencies/BlazeQueryDirectoryToTargetProvider.java#L72), which prevents the kind from derived. We can see this phenomenon in the UI:

<img width="542" alt="image" src="https://user-images.githubusercontent.com/326857/222921403-cf1aa3f7-e8ea-4820-83fb-094780f5c1e0.png">

As a result, the Debug icon gets disabled. Without the metadata intellij-info.txt created, the Bazel kind cannot be determined. As a result, the run configuration is viewed as not debuggable [here](https://github.com/bazelbuild/intellij/blob/bbbc838e13238e2b8c8fbf12e7ceb7f1a8c4adf1/java/src/com/google/idea/blaze/java/run/BlazeJavaDebuggerRunner.java#L54).

A few options I can see:

1. Do nothing. Remove the manual tag and run resync and you can workaround this issue.
2. Stop performing Debug icon check (e.g. disable returning false [here](https://github.com/bazelbuild/intellij/blob/bbbc838e13238e2b8c8fbf12e7ceb7f1a8c4adf1/java/src/com/google/idea/blaze/java/run/BlazeJavaDebuggerRunner.java#L54))
3. Run a query on the command `bazel query --output=label_kind //myapp:shaded_jar` to figure out if it can derive things regardless of the manual tag. Changes to the [ProjectTargetFinder](https://github.com/bazelbuild/intellij/blob/94b6618bffdc91ddb0b6f773fb99e1cb5f928cdc/base/src/com/google/idea/blaze/base/run/targetfinder/ProjectTargetFinder.java#L41-L43) method is likely to invoke the Bazel command should this information not be retrieved from `intellij-info.txt`.
[...]

PR #4547, opened 2023-03-04
Perform query lookup for the specific target in case it's not part of the synced data

rogerhu:
Related to: https://github.com/bazelbuild/intellij/issues/4546

# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/4546

# Description of this change

If the kind cannot be determined from the sync process, we invoke the Bazel process to query the kind being specified. There are similar classes that do similar things but this code gets invoked asynchronously as a Future and we only want to query the target specifically.

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/targetfinder/ProjectTargetFinder.java
+++ b/base/src/com/google/idea/blaze/base/run/targetfinder/ProjectTargetFinder.java
@@ -16,14 +16,24 @@
 package com.google.idea.blaze.base.run.targetfinder;
 
 import com.google.common.util.concurrent.Futures;
+import com.google.idea.blaze.base.async.process.ExternalTask;
+import com.google.idea.blaze.base.async.process.LineProcessingOutputStream;
+import com.google.idea.blaze.base.bazel.BuildSystem;
+import com.google.idea.blaze.base.command.BlazeCommand;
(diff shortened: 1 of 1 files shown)


cheister:
+1

mai93:
IIUC `Finder` interfaces like `TargetFinder` and `SourceToTargetFinder` must be fast, that's why they are implemented to just lookup the `TargetMap`. Also I think there is a pattern to avoid running bazel commands, like `bazel query` in this case, implicitly like this to not affect the performance.

rogerhu:
> IIUC `Finder` interfaces like `TargetFinder` and `SourceToTargetFinder` must be fast, that's why they are implemented to just lookup the `TargetMap`. Also I think there is a pattern to avoid running bazel commands, like `bazel query` in this case, implicitly like this to not affect the performance.

If that's a requirement (or is it?), would removing the manual tag (https://github.com/bazelbuild/intellij/blob/94b6618bffdc91ddb0b6f773fb99e1cb5f928cdc/base/src/com/google/idea/blaze/base/dependencies/BlazeQueryDirectoryToTargetProvider.java#L72) as part of generating the `TargetMap` be a better alternative?

mai93:
@rogerhu can we close this PR as https://github.com/bazelbuild/intellij/pull/5085 is merged now?

rogerhu:
Yes!

Issue #4496, opened 2023-02-23
Formatting buildfiles from IntelliJ can yield incorrect results


Duhemm:
### Description of the bug:

When formatting a file, buildifier is called without passing the `--path` argument. Because buildifier will receive the content of the file to format via stdin, buildifier cannot determine to which package the file being formatted belongs.

Unfortunately, this can lead to unexpected results during formatting. For instance, if `StripLabelLeadingSlashes` and `ShortenAbsoluteLabelsToRelative` are enabled, then the absolute label of a target at the root of the workspace such as `//:label` will be rewritten to the relative label `:label`, which has a different meaning if the build file being formatted is not at the root of the workspace.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Install buildifier
2. Clone https://github.com/Duhemm/buildifier-repro and import in IntelliJ
3. Edit `my-buildifier` to make sure it points to your install of `buildifier`
4. Open `subpkg/BUILD`
5. Run "Reformat code"
6. Notice the `actual` label changed from `actual = "//:example",` to `actual = ":example",`.

### Which Intellij IDE are you using? Please provide the specific version.

IntelliJ IDEA 2022.3.1 (Community Edition) Build #IC-223.8214.52

### What Bazel plugin version are you using?

2022.12.13.0.1-api-version-223

Duhemm:
I have a fix available in https://github.com/bazelbuild/intellij/compare/master...Duhemm:intellij:buildifier-path?expand=1.

I'd be happy to submit a PR if the maintainers are okay with it.

mai93:
Thank you for reporting this and for the fix, I noticed that the problem still happens on file save (with your fix patched). I think we need to pass the `--path` also [here](https://github.com/Duhemm/intellij/blob/1c1128a7372047b19354518697b9011b140c599d/base/src/com/google/idea/blaze/base/buildmodifier/BuildFileFormatter.java#L106) which is called by the onSaveHandler. I'm also looking into refactoring this part so that buildifier invocations are done from a single place to not run in similar cases like this.

Duhemm:
Thank you for taking a look @mai93. You're right, we also need to pass `--path` when saving. I've updated the patch and opened a pull request in #4550 .

PR #4550, opened 2023-03-06
Pass `--path` to buildifier when possible

Duhemm:
# Checklist

details.

Issue number: #4496 

# Description of this change

Previously, when formatting a file, buildifier would be called without passing the `--path` argument. Because buildifier will receive the content of the file to format via stdin, buildifier cannot determine to which package the file being formatted belongs.

Unfortunately, this can lead to unexpected results during formatting. For instance, if `StripLabelLeadingSlashes` and
`ShortenAbsoluteLabelsToRelative` are enabled, then the absolute label of a target at the root of the workspace such as `//:label` will be rewritten to the relative label `:label`, which has a different meaning if the build file being formatted is not at the root of the workspace.

We fix this issue by passing `--path=<path relative to workspace root>` to buildifier when formatting a file.

### DIFF
--- a/base/src/com/google/idea/blaze/base/buildmodifier/BuildFileFormatOnSaveHandler.java
+++ b/base/src/com/google/idea/blaze/base/buildmodifier/BuildFileFormatOnSaveHandler.java
@@ -61,7 +61,7 @@ public class BuildFileFormatOnSaveHandler extends FileDocumentManagerAdapter {
       return;
     }
     // DO not use formatter here, instead we format the entire file.
-    BlazeFileType type = ((BuildFile) psiFile).getBlazeFileType();
+    BuildFile buildFile = (BuildFile) psiFile;
     ListenableFuture<Void> future =
         FileBasedFormattingSynchronizer.applyReplacements(
             psiFile,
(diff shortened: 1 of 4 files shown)


base/src/com/google/idea/blaze/base/buildmodifier/BuildifierFormattingService.java
```diff
@@ -44,12 +48,20 @@ public final class BuildifierFormattingService extends AsyncDocumentFormattingSe
 
   static final Supplier<Optional<String>> binaryPath = Suppliers.memoize(() -> getBinary());
 
+  static ImmutableList<String> getCommandLineArgs(String binary, BuildFile buildFile) {
+    ImmutableList.Builder<String> cmd = ImmutableList.builder();
+    cmd.add(binary);
+    BlazeFileType type = buildFile.getBlazeFileType();
+    return cmd.add(fileTypeArg(type)).addAll(pathArg(buildFile)).build();
+  }
```
mai93, line 56: nit: I think it will be better if we moved this method (and the private methods it uses) to `BuildFileFormatter.java` because it already contains other method like `getReplacements` which are used by classes in this package.

base/src/com/google/idea/blaze/base/buildmodifier/BuildFileFormatOnSaveHandler.java
```diff
@@ -61,7 +61,7 @@ private static void formatBuildFile(Project project, PsiFile psiFile) {
       return;
     }
     // DO not use formatter here, instead we format the entire file.
-    BlazeFileType type = ((BuildFile) psiFile).getBlazeFileType();
+    BuildFile buildFile = (BuildFile) psiFile;
```
mai93, line 64: Can you please remove the unused import `import com.google.idea.blaze.base.lang.buildfile.psi.BuildFile.BlazeFileType;`?

base/src/com/google/idea/blaze/base/buildmodifier/BuildifierFormattingService.java
Duhemm: Done, thanks for the suggestion 👍

mai93 (approved):

Issue #1922, opened 2020-06-22
Incremental deployment failure with 'bazel-android-intro' codelab project


oliviernotteghem:
Incremental mobile deployment succeeds from command line, ie : 
`bazel mobile-install //mediarecorder/java/com/example/android/mediarecorder:mediarecorder --start_app --incremental`

However, when ran from Android Studio, several devices are specified when creating dex folder used for incremental deployment, causing command to fail : 
` ... '-s', 'emulator-5554:tcp:5037', ... `

Tried different physical / emulators device, and changing SDK version (28, 29, 30) : same issue

```
ERROR: /Users/oliviern/Uber/bazel-android-intro/solution/mediarecorder/java/com/example/android/mediarecorder/BUILD:13:15: Installing //mediarecorder/java/com/example/android/mediarecorder:mediarecorder failed (Exit 1)
Error:
adb command: ['/Users/oliviern/Library/Android/sdk/platform-tools/adb', '-s', 'emulator-5554', '-s', 'emulator-5554:tcp:5037', '-s', 'emulator-5554', '-s', 'emulator-5554:tcp:5037', 'shell', 'mkdir -p /data/local/tmp/incrementaldeployment/com.example.android.mediarecorder/dex']
return code: 1
stdout: 
stderr: adb: unknown host service
```

mai93:
Hello @oliviernotteghem do you still see this issue with the latest version of the plugin?

pswaminathan:
Hi @mai93, I am seeing this same issue (different project, but same behavior) with the latest plugin version.

Morfly:
Hi @pswaminathan. Could you click `Edit Configurations...` in Idea/AndroidStudio -> select your target -> disable `use mobile-install` checkmark and try again. It did the trick for me.

pswaminathan:
@Morfly unchecking it doesn't work either. I then get this:

```
Command: bazel build --tool_tag=ijwb:AndroidStudio --output_groups=+android_deploy_info "" --curses=no --color=yes --progress_in_terminal_title=no --fat_apk_cpu=arm64-v8a,armeabi-v7a --build_event_binary_file=/var/folders/71/0r_ylvb13vv2my1_676sx1gm0000gn/T/intellij-bep-a80e09e0-223d-4caa-ad57-b02a6ac7539d --nobuild_event_binary_file_path_conversion -- //src/seaborn/app:app

Loading: 
Loading: 0 packages loaded
[...]

PR #4669, opened 2023-03-31
Fix mobile install step

idanakav:
# Checklist

details.

#1922 

# Description of this change
Without an active tunnel config, the existing logic is adding ":tcp:port" to the device serial passed to adb, where port is the adb server port configured in the IDE.

The original intention was to use the right adb server port configured in the IDE, and unless running an adb connect, or different command that will modify the device serial we should simply use the -P option for that.


mai93:
merged in 32e2646a805fd0b82a11e6f7ee283f11118177e5

Issue #4664, opened 2023-03-30
Bazel CLion plugin has non-optional dependencies on Google Test, Catch and Boost Test plugins


odisseus:
### Description of the bug:

These are bundled plugins that come with CLion, but they cause a lot of overhead in some cases, so users which don't use the particular style of tests may want to disable some or all of these plugins. Unfortunately, this is not possible without disabling the Bazel plugin as well.

The Bazel plugin for Android studio has a similar problem with the Google Test plugin.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Install Clion
2. Install Bazel plugin
3. Try to disable either of the three plugins in the title.

### Which Intellij IDE are you using? Please provide the specific version.

Clion, Android Studio

### What programming languages and tools are you using? Please provide specific versions.

C++, Android

nikhil-marathe-skydio:
This seems like a duplicate of #4481

Issue #4481, opened 2023-02-22
Make CLion Boost.Test and CLion Catch optional dependencies


nikhil-marathe-skydio:
### Description of the feature request:

Using `2023.01.24.0.1-api-version-223` and earlier of the plugin.
I would like to disable the CLion Boost.Test and CLion Catch plugins because they add more indexing time on top of the already high indexing time for my project. However when I try to do so from the Plugins settings, I get a dialog saying this plugin depends on those so I would also have to disable the Bazel plugin.

My project doesn't use either of these test frameworks, but vendored third-party libraries do, so the plugins trigger their indexing.

I don't know why these plugins are required dependencies.

Thanks for creating and maintaining this plugin!

![image](https://user-images.githubusercontent.com/84412986/220692634-f5ceb60c-d2d1-4663-a5aa-a9f9456210e3.png)

### What underlying problem are you trying to solve with this feature?

Indexing on large projects is already slow. I'm trying to get some speedup by disabling unneeded plugins.

### What operating system, Intellij IDE and programming languages are you using? Please provide specific versions.

Ubuntu Linux 18.04, CLion 2022.3, C++

### Have you found anything relevant by searching the web?

No

martis42:
Would be really great if we could trim those dependencies

nikhil-marathe-skydio:
> Seems like the only reference to these plugins is here so it's reasonable to make them optional
> 
> https://github.com/bazelbuild/intellij/blob/f892b55deaa9425153ea8b81f9d67ee5610566c8/clwb/src/com/google/idea/blaze/clwb/run/producers/NonBlazeProducerSuppressor.java

I got this plugin setup for development, but I've never developed Intellij plugins before. If someone could explain what I need to do (apart from removing the lines here), I could take a stab at putting up a PR.

PR #4704, opened 2023-04-05
Make the dependencies on CLion's bundled test plugins optional.

odisseus:
To build the plugin:
```
bazel build //clwb:clwb_bazel_zip --define=ij_product=clion-2023.1
```

# Checklist

details.

Issue number: #4664, #4481
# Description of this change

### DIFF
--- a/clwb/src/META-INF/clwb.xml
+++ b/clwb/src/META-INF/clwb.xml
@@ -17,9 +17,6 @@
   <vendor>Google</vendor>
 
   <depends>com.intellij.modules.clion</depends>
-  <depends>org.jetbrains.plugins.clion.test.boost</depends>
-  <depends>org.jetbrains.plugins.clion.test.catch</depends>
-  <depends>org.jetbrains.plugins.clion.test.google</depends>
 
   <extensions defaultExtensionNs="com.intellij">
     <consoleFilterProvider implementation="com.google.idea.blaze.clwb.run.BlazeCppPathConsoleFilter$Provider"/>
@@ -28,7 +25,6 @@
 
     <editorNotificationProvider implementation="com.google.idea.blaze.plugin.CMakeNotificationFilter" />
     <moduleType id="BLAZE_CPP_MODULE" implementationClass="com.google.idea.blaze.clwb.BlazeCppModuleType"/>
-    <postStartupActivity implementation="com.google.idea.blaze.clwb.run.producers.NonBlazeProducerSuppressor"/>
     <actionConfigurationCustomizer implementation="com.google.idea.blaze.plugin.ClwbHideMakeActions"/>
   </extensions>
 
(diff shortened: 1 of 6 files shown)


clwb/src/META-INF/clwb.xml
```diff
@@ -28,7 +25,6 @@
 
     <editorNotificationProvider implementation="com.google.idea.blaze.plugin.CMakeNotificationFilter" />
     <moduleType id="BLAZE_CPP_MODULE" implementationClass="com.google.idea.blaze.clwb.BlazeCppModuleType"/>
-    <postStartupActivity implementation="com.google.idea.blaze.clwb.run.producers.NonBlazeProducerSuppressor"/>
```
lukaszwawrzyk, line 31: It seems like the class is not deleted.

clwb/src/com/google/idea/blaze/clwb/run/producers/RunConfigurationProducerSuppressor.java
```diff
@@ -0,0 +1,63 @@
...
+  @Override
+  public void runActivity(@NotNull Project project) {
+    if (Blaze.isBlazeProject(project)) {
+      RunConfigurationProducerService producerService =
+              RunConfigurationProducerService.getInstance(project);
+      producerService.addIgnoredProducer(producerType());
+    }
+  }
+
+  public static class Boost extends RunConfigurationProducerSuppressor {
```
lukaszwawrzyk, line 43: To be safe, I'd put each of these class in a separate file (to avoid potential errors during class loading when one of the plugins is disabled).

odisseus: I have tested this by loading a C++ Bazel project with all three plugins disabled, and I didn't see any errors.

In fact, this whole situation is a bit strange, because the only reason why the Bazel plugin needs to have these testing plugins in the classpath is to know the names of the classes that it tries to disable. Since it is only the name of the class that matters, we could could have hardcoded it as a String, but unfortunately `RunConfigurationProducerService` [requires](https://github.com/JetBrains/intellij-community/blob/a39956e5e1712c9bec5b470b2362fef32e711119/platform/lang-api/src/com/intellij/execution/RunConfigurationProducerService.java#L56) a Class reference (only to call getName() on it immediately).

clwb/src/META-INF/clwb.xml
odisseus: Fixed

clwb/src/com/google/idea/blaze/clwb/run/producers/RunConfigurationProducerSuppressor.java
lukaszwawrzyk: Maybe we should go this way and just remove dependencies on these plugins completely?

PR #5147, opened 2023-07-25
Override Reload File debugger action

brian-mcnamara:
# Checklist

details.

# Description of this change

Currently, IJ has a "Compile and Reload file" option available via right clicking the contents of a java file. This is currently broken with Bazel since the action needs to be overridden to delegate to bazel build. This work aims to implement a similar functionality via overriding the action, delegating to bazel build, and finding and extracting the compiled classes to pass hotswap.


java/src/com/google/idea/blaze/java/run/hotswap/BlazeReloadFileAction.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Map;
+import java.util.concurrent.Future;
+import java.util.jar.JarInputStream;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
+
+public class BlazeReloadFileAction extends AnAction {
+
+    public static final String ACTION_ID = "Debugger.ReloadFile";
```
dkashyn-sfdc, line 51: looks like quite generic name, will it be scoped to Bazel in some way?

brian-mcnamara: This is the JB action id that we are overriding.
https://github.com/JetBrains/intellij-community/blob/2278d05ac40658a82fd3780a4145a292898b70b2/resources/src/idea/JavaActions.xml#L514

mai93:
I'm not sure if it is possible to reuse the logic of [BlazeHotSwapActionManager](https://github.com/bazelbuild/intellij/blob/ae1e62a403514a43bb2cc1ea42b6aaff1bc402fe/java/src/com/google/idea/blaze/java/run/hotswap/BlazeHotSwapManager.java) to reload the certain file of interest in [BlazeReloadFileAction.java](https://github.com/bazelbuild/intellij/pull/5147/files#diff-5396dc4c99f399f654f4910a1fdbc516baee1cbf1d5799fbd1d9d28806f38555)

brian-mcnamara:
> I'm not sure if it is possible to reuse the logic of [BlazeHotSwapActionManager](https://github.com/bazelbuild/intellij/blob/ae1e62a403514a43bb2cc1ea42b6aaff1bc402fe/java/src/com/google/idea/blaze/java/run/hotswap/BlazeHotSwapManager.java) to reload the certain file of interest in [BlazeReloadFileAction.java](https://github.com/bazelbuild/intellij/pull/5147/files#diff-5396dc4c99f399f654f4910a1fdbc516baee1cbf1d5799fbd1d9d28806f38555)

I did look at this code initially, unfortunately not without a lot of refactoring (which I could do, just decided it would not be a lot of overlapping code to consider it).

The current manager has a tight coupling to the ClassFileManifest which does not work for our case (where bazel run is run outside the IDE and we are attaching to the instance). Because of this (and other fun our app launcher works), the ClassFileManifest does not work. Because of this, the route I took with this PR is to only compile and hotswap a single file. Since we know the file, we do not need to use the ClassFileManifest. 

Let me know if I missed something or if you believe refactoring would still be useful and Ill take a closer look :) .

brian-mcnamara:
@mai93 checking back on this. Any thoughts or concerns with this getting merged?

mai93:
Sorry for the delay! I do not think I will have the time to review this, @tpasternak offered to take a look next week.

java/src/com/google/idea/blaze/java/run/hotswap/BlazeReloadFileAction.java
dkashyn-sfdc: I think then we might need to override `CompileFile` and `CompileProject` actions as well. They doesn't work with Bazel currently.

brian-mcnamara: We do not need that as part of the PR.

dkashyn-sfdc: Agree, let's start with this and think about the rest of them after this one merged.

java/src/com/google/idea/blaze/java/run/hotswap/BlazeReloadFileAction.java
```diff
@@ -0,0 +1,232 @@
...
+                        tempOutputDir.deleteOnExit();
+                    } catch (IOException e) {
+                        LOGGER.error("Failed creating temp directory for hotswap", e);
+                        return;
+                    }
+
+                    ProgressManager.getInstance().run(new Task.Backgroundable(project, "Preparing to hotswap", true) {
+                        @Override
+                        public void run(@NotNull ProgressIndicator progressIndicator) {
+                            // We want to be aware of entries like darwin_arm64-fastbuild/bin/some-api/lib-api.jar and not interested in darwin_arm64-fastbuild/external-api-0.0.654.jar
```
agluszak, line 101: Why not? Also, does it accomodate for bzlmod-related changes to path structure? Did you test it with bzlmod?

java/src/com/google/idea/blaze/java/run/hotswap/BlazeReloadFileAction.java
```diff
@@ -0,0 +1,232 @@
...
+                                LOGGER.warn(String.format("There was nothing found for HotSwap for package %s in outputs %s", jarDirectory, artifactsForLogging.get()));
+                            }
+                        }
+                    });
+                } else {
+                    LOGGER.debug("Run hotswap after compile set to 'never'");
+                }
+
+            } catch (Exception e) {
+                LOGGER.error("Exception occurred during building file to hotswap", e);
```
agluszak, line 127: Action cancellation is not handled correctly. Cancelling an ongoing hotswap results in this internal IDE error being thrown:

```
Exception occurred during building file to hotswap

java.util.concurrent.CancellationException: Task was cancelled.
	at com.google.common.util.concurrent.AbstractFuture.cancellationExceptionWithCause(AbstractFuture.java:1543)
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:586)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:567)
	at com.google.common.util.concurrent.FluentFuture$TrustedFuture.get(FluentFuture.java:91)
	at com.google.idea.blaze.java.run.hotswap.BlazeReloadFileAction.lambda$actionPerformed$1(BlazeReloadFileAction.java:76)
	at com.intellij.openapi.application.impl.ApplicationImpl$2.run(ApplicationImpl.java:249)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at com.intellij.util.concurrency.ContextCallable.call(ContextCallable.java:32)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at com.intellij.util.concurrency.ContextRunnable.run(ContextRunnable.java:27)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:702)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:699)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1.run(Executors.java:699)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

agluszak (changes requested):
Doesn't work for me.

I tested it on Linux by running the `getPendingExternalDeps_followJavaDeps_allBuilt` test from this repo. I placed a breakpoint inside it and clicked `Compile and reload file`.

The `building targets` progress indicator seems to be stuck forever. Looks like we're deadlocking Bazel.

image

image

Build waits for the test to finish, but it was started because I requested a rebuild from inside the test...

java/src/com/google/idea/blaze/java/run/hotswap/BlazeReloadFileAction.java
dkashyn-sfdc: Because libraries are not sources and parts of our project has `platfrom_stuff/bin/` which corresponds to `bazel-bin` symlink

dkashyn-sfdc:
> Doesn't work for me.
> 
> I tested it on Linux by running the `getPendingExternalDeps_followJavaDeps_allBuilt` test from this repo. I placed a breakpoint inside it and clicked `Compile and reload file`.
> 
> The `building targets` progress indicator seems to be stuck forever. Looks like we're deadlocking Bazel.
> 
> image
> 
> image
> 
> Build waits for the test to finish, but it was started because I requested a rebuild from inside the test...

Does it look like https://github.com/bazelbuild/bazel/issues/2337 ?

agluszak:
> Does it look like [bazelbuild/bazel#2337](https://github.com/bazelbuild/bazel/issues/2337) ?

I'm not sure what you mean

java/src/com/google/idea/blaze/java/run/hotswap/BlazeReloadFileAction.java
agluszak: Then updating the comment to reflect the intent, not the effect, will be very much appreciated. But anyway I'm not sure if that's a future-proof way of achieving what you want. + I'd like to bump my question about bzlmod

brian-mcnamara:
> > Does it look like [bazelbuild/bazel#2337](https://github.com/bazelbuild/bazel/issues/2337) ?
> 
> I'm not sure what you mean. I don't run anything in the command line manually. What I did was (I guess?) your expected workflow with this feature.

@agluszak was taking a look at this yesterday, as far as I could tell the original hotswap implementation is also not supported when running tests via bazel. Seems like `bazel run` does not lock the server, but `bazel test` does. That said, I did not intend to replace the original hotswap mechanism, simply provide another option when not using launching the bazel target through the debug profile. It was intended for remote debugging. Either way, I'll update the logic to disable this action when the debug session is from a bazel target. In addition ill address the cancelation exception you discovered.

java/src/com/google/idea/blaze/java/run/hotswap/BlazeReloadFileAction.java
agluszak: @dkashyn-sfdc @brian-mcnamara bump

brian-mcnamara: Sorry for the delay, Dmitry is out and I'm not exactly sure whats needed. I know Dmitry added it to help with debugging, I was unsure about it as its quite verbose.

brian-mcnamara:
Pushed a commit to add extra handling for the case of unsynced files. Was noticing the action was still available even when the file was unsynced. I also found that there was an exception for slow tasks in the UI thread, I have to override `getActionUpdateThread` to allow the BGT thread to be used, which the original action was using.

tpasternak:
@agluszak @brian-mcnamara I'm closing this PR, feel free to reopen if you still want to work on it

PR #5162, opened 2023-07-28
Make the default projectview empty during autoimport and add some com…

tpasternak:
…ments there

The difference between autoimport and regular import is that, during autoimport the user has no chance to edit the project view file before the import actually happens. Hence, if a repository is large, the current default "directory" value (".") caused the whole repository to be traversed.

This means the user sometimes has to wait a long time only to reduce the import scope. It would be better to import nothing by default and let the users choose what they want to import.

# Checklist

details.

# Description of this change

### DIFF
--- a/base/src/com/google/idea/blaze/base/project/AutoImportProjectOpenProcessor.java
+++ b/base/src/com/google/idea/blaze/base/project/AutoImportProjectOpenProcessor.java
@@ -196,22 +196,21 @@ public class AutoImportProjectOpenProcessor extends ProjectOpenProcessor {
 
   private ProjectView defaultEmptyProjectView() {
     Builder projectViewBuilder = ProjectView.builder();
-
-    projectViewBuilder.add(
-        ListSection
-            .builder(DirectorySection.KEY)
-            .add(DirectoryEntry.include(new WorkspacePath(".")))
-            .build()
-    );
-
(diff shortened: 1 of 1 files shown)


blorente:
I like the idea of not forcing `directories: "."` on auto-import, but I wonder what the default value is if we leave it empty.

I'm on my phone so can't check, but I think it will just import nothing (and thus show no files in the project view). If this is the case, I don't think this is ever what we want to do. Some alternatives that come to mind:

- Open the project view immediately after auto-importing, so that the first thing users see is the documentation.
- Prompt the user for their desired directory to import. With this change, we're essentially saying that we'll need manual intervention 100% of the time, so might as well do it in a nice form
- Guess the size of the project (maybe counting the number of files), and only default to empty if the project is above a certain size (10k files?).

Again, I can't check what the actual experience is with this change until Monday, so happy to hear other opinions.

liucijus (approved):

liucijus:
LGTM. @tpasternak do you have any numbers to share how long it takes with the current default view, which includes root dir, but does not sync any targets?

blorente (approved):
> Project view file is automatically opened immediately after empty project import

If this is the case, then this seems like an alright change for big projects. I'd still love to see a size-based filter, but it's probably work.

Issue #5217, opened 2023-08-08
Unnecessary computations in the sync process


idanakav:
### Description of the bug:

The issue will outline a few low hanging fruits in the sync process that can help optimizing the incremental sync times in the plugin.

1. `WorkingSet` computation can be skipped if expand to working set is disabled, computed 3 times per incremental sync.
Avoiding this computation can reduce sync times by roughly 3x git diff durations.

2. `BazelInfo` is computed 5 times during a incremental sync, with no caching involved.
Avoiding this computation can reduce sync times by roughly 5x bazel info durations.

For the Bazel info case, the idea is to cache the result and invalidate once needed, for example in a bazel sync error, alternatively, if that would be too tricky or error prone, we can proceed with computing it only once per sync.

### What Bazel plugin version are you using?

latest

PR #5219, opened 2023-08-08
Sync process optimizations

idanakav:
# Checklist

Issue number: #5217

More details can be found in the issue and on each commit.

### DIFF
--- a/base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
+++ b/base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2023 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.idea.blaze.base.command.info;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.idea.blaze.base.async.executor.BlazeExecutor;
+import com.google.idea.blaze.base.bazel.BuildSystem;
+import com.google.idea.blaze.base.scope.BlazeContext;
+import com.google.idea.blaze.base.settings.Blaze;
+import com.google.idea.blaze.base.settings.BlazeImportSettings;
+import com.google.idea.blaze.base.settings.BlazeImportSettingsManager;
+import com.google.idea.blaze.base.sync.SyncListener;
+import com.google.idea.blaze.base.sync.SyncMode;
+import com.google.idea.blaze.base.sync.SyncResult;
+import com.google.idea.common.experiments.BoolExperiment;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * <p> Providing a {@link com.google.idea.blaze.base.command.info.BlazeInfo}, the result will be cached and invalidated on any sync failure. </p>
+ *
+ * @see BlazeInfoRunner
+ */
+public class BlazeInfoProvider {
+    private static final Logger logger = Logger.getInstance(BlazeInfoProvider.class);
+    private static final BoolExperiment enabled =
+            new BoolExperiment("blaze.info.provider.enabled", false);
+
+    private final Project project;
+    private volatile BlazeInfo blazeInfo;
+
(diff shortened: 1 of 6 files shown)


base/src/com/google/idea/blaze/base/sync/ProjectStateSyncTask.java
```diff
@@ -189,6 +182,29 @@ private SyncProjectState getProjectState(BlazeContext context, BlazeSyncParams p
         .build();
   }
 
+  private ListenableFuture<BlazeInfo> createBazelInfoFuture(
+          BlazeContext context,
+          List<String> syncFlags,
+          SyncMode syncMode) {
+    boolean useBazelInfoRunner = !BlazeInfoProvider.isEnabled() || syncMode == SyncMode.FULL;
```
blorente, line 189: Maybe we could just pass the sync mode to the `BlazeInfoProvider` and have it invalidate itself (or pretend that it has invalidated) on a full sync? I quite like the idea of "everything goes through the provider".

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
```diff
@@ -0,0 +1,136 @@
...
+            ).get();
+            return blazeInfo;
+        } catch (InterruptedException | ExecutionException e) {
+            logger.warn("Unable to run blaze info", e);
+            return null;
+        }
+    }
+
+    public void invalidate() {
+        logger.debug("invalidating");
```
blorente, line 119: leftover print?

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public static BlazeInfoProvider getInstance(Project project) {
+        return project.getService(BlazeInfoProvider.class);
+    }
+
+    public static Boolean isEnabled() {
+        return enabled.getValue();
+    }
+
+    public ListenableFuture<String> getBlazeInfo(
```
blorente, line 63: Should this be public? Taking a key seems like an implementation detail.

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
```diff
@@ -0,0 +1,136 @@
+/*
```
blorente, line 1: I'm not sure we can cache this between runs. Command-line flags affect the outcome (e.g. `--server_javabase`), which can be in any of multiple locations and bazelrc files. Caching for a single run seems okay, since a lot of the flags that change the info output are startup flags and would invalidate most of the sync anyway.

blorente (commented):
Thanks for the contribution!

The main ideas look good, but I don't think we can cache the `bazel info` between runs. Command-line flags affect the outcome (e.g. --server_javabase), which can be in any of multiple locations and bazelrc files. Caching for a single run seems okay, since a lot of the flags that change the info output are startup flags and would invalidate most of the sync anyway.

So, if you could change the logic in `BlazeInfoProvider` to invalidate every run, I think it's a great PR!

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
idanakav: Maybe we can remove this method altogether, the idea was to offer same functionality as [BlazeInfoRunner](https://github.com/bazelbuild/intellij/blob/17b0bb8147bd11642c11cbce325ac9f301bca402/base/src/com/google/idea/blaze/base/command/info/BlazeInfoRunner.java#L27), but currently this method is not being used.

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
idanakav: It can be handy for debugging purposes.

base/src/com/google/idea/blaze/base/sync/ProjectStateSyncTask.java
idanakav: The idea was to keep `BlazeInfoProvider` useful to other use cases apart from the sync, but if we are not planning to extend it, we can make it more coupled to the sync and pass the sync mode as well.

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
idanakav: I've updated the PR to invalidate after each sync.

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
idanakav: Removed.

base/src/com/google/idea/blaze/base/command/info/BlazeInfoProvider.java
```diff
@@ -0,0 +1,124 @@
...
+        public void afterSync(
+                Project project,
+                BlazeContext context,
+                SyncMode syncMode,
+                SyncResult syncResult,
+                ImmutableSet<Integer> buildIds) {
+            if(!isEnabled()) {
+                return;
+            }
+            BlazeInfoProvider.getInstance(project).invalidate();
```
blorente, line 121: Should we check for null here? In case we de-register the provider accidentally and forget to de-register the invalidator.

blorente (approved):
LGTM! A minor nit, but otherwise thanks for the contribution!

blorente (approved):

blorente:
Tested it locally in IJ Ultimate and saw no issues.

PR #5227, opened 2023-08-10
Add managed project to import wizard project view field

brian-mcnamara:
# Checklist

details.

# Description of this change

This change adds either a default or adds to the history the projects .managed.bazelproject file if it exists. This helps users who are importing the project though the wizard to find this file


brian-mcnamara:
Switched over to adding new radio buttons in the wizard. I'm not 100% on the descriptions, so let me know if they should be updated
Screenshot 2023-08-14 at 10 25 53 AM

brian-mcnamara:
Thanks @blorente 
Interesting to know, we are currently checking in a base .managed.projectview file to provide some documentation and examples as our users are new to bazel and this provides a path forward to quickly onboard. While we do mainly rely on the auto-import behavior, there is a subset of users (due to our migration strategy for our existing build system to bazel) that have auto-import disabled to prevent switch over during the migration. Because of that, the path forward for them is to use the import wizard.

As for understanding the differences, thats a fair point, and I do agree updates to clearly this and the other options makes sense. I could add the path to the file as part of the option, but let me think about other options too.

blorente:
@tpasternak I agree with your points, moving the radio buttons to the bottom and explaining which file it is is probably sufficient.

brian-mcnamara:
Updated UI

Screenshot 2023-08-31 at 11 38 10 AM

PR #5341, opened 2023-09-13
Trying to workaround "unsynced" case for newly added files.

dkashyn-sfdc:
# Checklist

details.

# Description of this change
https://github.com/bazelbuild/intellij/issues/5347

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/autosync/ProjectTargetManagerImpl.java
+++ b/base/src/com/google/idea/blaze/base/sync/autosync/ProjectTargetManagerImpl.java
@@ -34,6 +34,8 @@ import com.google.idea.blaze.base.targetmaps.SourceToTargetMap;
 import com.intellij.ide.projectView.ProjectView;
 import com.intellij.openapi.project.Project;
 import java.io.File;
+import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


mai93:
This can be fixed by running "Partial Sync" on the newly added file, can you explain why this fix is needed?

dkashyn-sfdc:
Because when you refactoring/adding files, `unsynced` by default is confusing. 

Of course `Partial Sync` can help but:
 - it takes 1 minute for our project if you are lucky
 - with existing IJ project structure this operation is excessive so we can use sync status of surrounding files
 - it is so opaque and unapparent to our users that manual sync is required for every move/add operation

blorente:
Sorry for the late review. The more I think about it, the more I am against this sort of change, because it will expose lies to users. 

I agree that we should have a lightweight way of updating the SourceToTargetMap for just one target that essentially expands the globs in `srcs`. Does `Bazel` -> `Sync` -> `Sync Directories` work for this case? It should be much faster, since it doesn't build anything with Bazel.

dkashyn-sfdc:
@blorente, sorry for being a little bit blunt here, but we are having issues even with the first "Bazel Sync" need, with our engineers. This operation is so unnatural for IJ. 

"Sync" after any refactor or file changes might be okay, for Google with your huge project and associated tooling, but for mere mortals outside of Google, this is so unnatural that people are really struggling with this remedy.

Even more: a few of hundreds of our engs are fluent in all 5 types of syncs that Bazel Plugin has.

dkashyn-sfdc:
Well, this was never 100%-complete solution since if the package is empty there is nothing to triangulate against.

Also if we are talking about tests there won't be consensus in packages for those.

dkashyn-sfdc:
We decided not to proceed with this PR and to keep it to our internal version only :( I wish BSP were here faster!

PR #5458, opened 2023-10-11
Fix manual targets sync for target explicitly specified

rogerhu:
# Checklist

This PR is a continuation of https://github.com/bazelbuild/intellij/issues/4546. One issue is that for manual targets specified, they are also excluded without this change (see https://github.com/bazelbuild/intellij/issues/4546#issuecomment-1741969833).

This fix will only work currently if the [shard_sync: true](https://ij.bazel.build/docs/project-views.html#shard_sync). Otherwise the query will not be performed. I pushed a [second commit](https://github.com/bazelbuild/intellij/pull/5458/commits/680862626c3fc88f8be0a1437450e73ed33dd118) to make this change explicit. Otherwise, two flags must be set for this part to work.

Issue number: https://github.com/bazelbuild/intellij/issues/4546

# Description of this change

When `allow_manual_targets_sync: true` is set, this change will allow manual targets to be part of the target query. Because expansion of build targets currently only occur when the sharding approach is enabled, this change would also switch on the `shard_sync: true` option to enable this query to occur. The alternative would be to create another option to expand the query, so I wanted to get feedback about it.

I've also modified unit tests to perform the check for both on/off, asserting the presence of the manual attr query.

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/SyncProjectTargetsHelper.java
+++ b/base/src/com/google/idea/blaze/base/sync/SyncProjectTargetsHelper.java
@@ -89,7 +89,7 @@ public final class SyncProjectTargetsHelper {
     return viewSet.getScalarValue(AutomaticallyDeriveTargetsSection.KEY).orElse(false);
   }
 
-  private static boolean shouldSyncManualTargets(ProjectViewSet viewSet) {
+  public static boolean shouldSyncManualTargets(ProjectViewSet viewSet) {
     return viewSet.getScalarValue(SyncManualTargetsSection.KEY).orElse(false);
   }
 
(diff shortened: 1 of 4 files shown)


blorente (approved):

rogerhu:
> @rogerhu @tpasternak Can we have this documented somewhere next to https://ij.bazel.build/docs/project-views.html#derive_targets_from_directories ? I've sharing the issue for this PR but I think docs supposed to reflect such properties.

Yep! I have to file a ticket apparently since the Bazel IntelliJ docs are not open source. =)

https://bazelbuild.slack.com/archives/C025SBYFC4E/p1697741079375969

rogerhu:
Created a ticket here: https://github.com/bazelbuild/intellij/issues/5590

PR #1786, opened 2020-04-22
Don't build unnecessary jars.

uri-canva:
# Checklist

- [x] I have filed an issue about this change and discussed potential changes with the maintainers.
- [ ] I have received the approval from the maintainers to make this change.
- [x] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: https://github.com/bazelbuild/intellij/issues/1785

# Description of this change

Don't build the class jar if source jars are available, as we can use that in conjunction with the interface jar to index the code instead.

Issue #1785, opened 2020-04-22
Syncing java targets builds resources.


uri-canva:
When syncing java targets the class jars are built, which requires building all resources. It's not clear why that would be necessary since the resources don't seem to be used for code indexing.

This causes syncing to take longer than necessary.

This might be due to there not being an output for a jar without resources in https://docs.bazel.build/versions/3.0.0/skylark/lib/java_output.html.

uri-canva:
I have noticed in https://github.com/bazelbuild/intellij/blob/c279bd605818d660a37e5a118c5a30761e821fb5/base/src/com/google/idea/blaze/base/ideinfo/LibraryArtifact.java#L88 we end up not using the class jar at all if a source jar is available. That makes sense: if we don't have a source jar, we can refer to the decompiled class jar, but if we have the source jar we can match that with the interface jar.

It seems wasteful to build the class jar just to have it ignored later on, so I have changed the aspect to not build it if source jars are available. See https://github.com/bazelbuild/intellij/pull/1786.

uri-canva:
ping @brendandouglas

brendandouglas:
Thanks for looking into this.

I have some concerns with the change which I'd need to spend some time investigating before committing to it:

1. is there any guarantee that rules which declare source jar outputs at analysis time actually output them at build time?  Similarly, what if a rule produces no ijars, only class jars and source jars?

2. there are situations where we will decide to drop source jars unconditionally, especially when working from home with additional security restrictions or an average network connection -- in those cases, build time (when building remotely) is not an issue, and we want to minimize the size of the jars we're copying to the local machine. We'd also prefer class jars over interface jars when available in that case. Switching this behavior at the aspect level is much more complex than doing it post-build at the plugin level (output groups are our best means of passing 'flags' to the aspect, and we already have too many of those)

brendandouglas:
There's also the question of how much time we're actually saving by skipping class jars -- do you have some data on this?

uri-canva:
> There's also the question of how much time we're actually saving by skipping class jars -- do you have some data on this?

I don't have hard data, but some of our java targets include closure compiled javascript or rust binaries as resources, building those resources takes over twice as long as building the java code in the whole repo, with those types of compilation being so time consuming.

So this patch is very important for us.

uri-canva:
> is there any guarantee that rules which declare source jar outputs at analysis time actually output them at build time? Similarly, what if a rule produces no ijars, only class jars and source jars?

[...]

PR #5515, opened 2023-10-20
Don't build unnecessary jars.

nkoroste:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/1785

# Description of this change

reopened https://github.com/bazelbuild/intellij/pull/1786 - Don't build the class jar if source jars are available, as we can use that in conjunction with the interface jar to index the code instead.

For our app it reduces `~30%` of the Jars the IDE will index (from `14484` to `10072`)

As per PR review request I put this feature behind `optimize_building_jars` experiment which is enabled by default. In case an issue is found with this feature this experiment can be disabled by disabled it in the IntelliJ experiments file as follows:
```
echo optimize_building_jars=0 >> ~/.intellij-experiments
```

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -605,6 +605,11 @@ def _collect_generated_files(java):
         return [(java.annotation_processing.class_jar, java.annotation_processing.source_jar)]
     return []
 
+def _should_optimize_building_jars(ctx):
+    if ctx.attr.optimize_building_jars == "enabled":
+        return True
+    return False
+
 def collect_java_info(target, ctx, semantics, ide_info, ide_info_file, output_groups):
     """Updates Java-specific output groups, returns false if not a Java target."""
     java = get_java_provider(target)
@@ -625,7 +630,7 @@ def collect_java_info(target, ctx, semantics, ide_info, ide_info_file, output_gr
     sources = sources_from_target(ctx)
     jars = [library_artifact(output, ctx.rule.kind) for output in java_outputs]
     class_jars = [output.class_jar for output in java_outputs if output and output.class_jar]
-    output_jars = [jar for output in java_outputs for jar in jars_from_output(output, ctx.attr.optimize_building_jars)]
+    output_jars = [jar for output in java_outputs for jar in jars_from_output(output, _should_optimize_building_jars(ctx))]
     resolve_files = output_jars
     compile_files = class_jars
 
@@ -768,18 +773,18 @@ def _build_filtered_gen_jar(ctx, target, java_outputs, gen_java_sources, srcjars
     """Filters the passed jar to contain only classes from the given manifest."""
     jar_artifacts = []
     source_jar_artifacts = []
-    should_should_optimize_building_jars = ctx.attr.optimize_building_jars
+    should_optimize_building_jars = _should_optimize_building_jars(ctx)
     for jar in java_outputs:
         if jar.ijar:
             jar_artifacts.append(jar.ijar)
-        elif jar.class_jar and not should_should_optimize_building_jars:
+        elif jar.class_jar and not should_optimize_building_jars:
             jar_artifacts.append(jar.class_jar)
         if hasattr(jar, "source_jars") and jar.source_jars:
             source_jar_artifacts.extend(_list_or_depset_to_list(jar.source_jars))
         elif hasattr(jar, "source_jar") and jar.source_jar:
             source_jar_artifacts.append(jar.source_jar)
 
-    if should_should_optimize_building_jars:
+    if should_optimize_building_jars:
         if len(source_jar_artifacts) == 0 or len(jar_artifacts) == 0:
             jar_artifacts.extend([jar.class_jar for jar in java_outputs.jars if jar.class_jar])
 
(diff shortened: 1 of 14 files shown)


agluszak:
Hi! 

> For our app it reduces ~30% of the Jars the IDE will index (from 14484 to 10072)
Would you be willing to run some benchmarks on open source projects and provide the results here?

aspect/intellij_info_impl.bzl
```diff
@@ -188,9 +188,11 @@ def jars_from_output(output):
     """Collect jars for intellij-resolve-files from Java output."""
     if output == None:
         return []
+    source_jars = get_source_jars(output)
+    jars = source_jars
```
blorente, line 192: What is this assignment for?

blorente (approved):
The implementation looks good to me, and I think it's unlikely to interfere with indexing. However, as @agluszak said, given that this changes a lot of the core resolution, it would be interesting to run the same benchmark in other projects (such as this repository, for instance) before merging, to make sure it's not a regression.

aspect/intellij_info_impl.bzl
nkoroste: Not needed, you're right. 
```suggestion
```

nkoroste:
> Hi!
> 
> > For our app it reduces ~30% of the Jars the IDE will index (from 14484 to 10072)
> 
> Would you be willing to run some benchmarks on open source projects and provide the results here?

and 

> The implementation looks good to me, and I think it's unlikely to interfere with indexing. However, as @agluszak said, given that this changes a lot of the core resolution, it would be interesting to run the same benchmark in other projects (such as this repository, for instance) before merging, to make sure it's not a regression.

I can provide a rough count of how many Jars would be ignored however I don't know how interesting that number is, instead what would be more interesting is to measure the impact this may have on indexing and code-completion time but unfortunately Intellij  IDE itself nor this plugin provide an easy way to get those numbers

mai93:
Can we try to incrementally roll it out?, I feel like we have been adding a lot of feature flags and not all of them are documented so not a lot of users get to try the features they protect.

mai93:
We can use [`FeatureRolloutExperiment`](https://github.com/bazelbuild/intellij/blob/master/common/experiments/src/com/google/idea/common/experiments/FeatureRolloutExperiment.java) to define the experiment flag and set the roll out percentage in [experiment.properties](https://github.com/bazelbuild/intellij/blob/master/common/experiments/src/com/google/idea/common/experiments/experiment.properties)

mai93:
Follow up on how we can select between the old and new behaviour in the aspect, we can use [--aspects_parameters](https://bazel.build/reference/command-line-reference#flag--aspects_parameters) to pass a boolean value to the aspect based on the experiment value (i.e. enabled or not) and select the one of the 2 approaches in the aspect based on this boolean. We need to add a boolean attribute in the aspect definition, can be called `optimize_building_jars` and pass its value from the command line as `--aspects_parameters=optimize_building_jars=true`.

aspect/testing/tests/src/com/google/idea/blaze/aspect/java/dependencies/DependenciesTest.java
```diff
@@ -103,20 +100,17 @@ public void testJavaLibraryWithExports() throws Exception {
...
             testRelative("libfoo_exporter-src.jar"),
-            testRelative("libexport_consumer.jar"),
             testRelative("libexport_consumer-hjar.jar"),
             testRelative("libexport_consumer-src.jar"));
     assertThat(getOutputGroupFiles(testFixture, "intellij-compile-java"))
         .containsAtLeast(
             testRelative("libfoo.jar"),
             testRelative("libfoo_exporter.jar"),
-            testRelative("libexport_consumer.jar"));
+            testRelative("libexport_consumer.jar"));z
```
nkoroste, line 113:
```suggestion
            testRelative("libexport_consumer.jar"));
```

nkoroste:
PTAL @mai93 @tpasternak

aspect/testing/tests/src/com/google/idea/blaze/aspect/java/javabinary/JavaBinaryTest.java
```diff
@@ -58,10 +58,8 @@ public void testJavaBinary() throws Exception {
             testRelative(intellijInfoFileName("foo")));
     assertThat(getOutputGroupFiles(testFixture, "intellij-resolve-java"))
         .containsAtLeast(
-            testRelative("libfoolib.jar"),
```
mai93, line 61: why these jars are no longer expected, is optimization enabled for this test? same for `DependenciesTest`

nkoroste: Reverted these changes, they were traces to when optimization was enabled by default.

mai93:
it looks good to me, I can merge it if none else wants to review cc @tpasternak @agluszak

aschott-looker:
This option isn't supported on the version of Bazel we use. How do you disable it for a project?

mai93:
what feature is not supported in the Bazel version you use?

@nkoroste mentioned that this experiment can be disabled by disabled it in the IntelliJ experiments file as follows:

`echo optimize_building_jars=0 >> ~/.intellij-experiments`

aschott-looker:
Any Bazel version older than when [this commit](https://github.com/bazelbuild/bazel/commit/37710728712e00e8bd7145662668d72591204146) landed will fail:

```
ERROR: --aspects_parameters=optimize_building_jars=enabled :: Unrecognized option: --aspects_parameters=optimize_building_jars=enabled
```

tpasternak:
@mai93 , @ujohnny do we still support Bazel 5?

tpasternak:
Bazel 5 will be EOL in September. https://blog.bazel.build/2020/11/10/long-term-support-release.html

@aschott-looker are there any chances you could upgrade to bazel 6?

hisener:
https://bazel.build/release says it's EOL in Jan 2025.

aschott-looker:
@tpasternak We definitely want to upgrade to a more recent version of Bazel but it's not feasible in the short term. So I was hoping for a quick workaround.

Thanks for creating the issues above!

tpasternak:
the hotfix is on its way #6653

PR #5526, opened 2023-10-24
fix: Mark package with BUILD.bazel file in completion

stevebarrau:
# Checklist

Discussed offline with @blorente. This looks like a regression.

# Description of this change

Mark package with BUILD.bazel file in completion

### The base branch for this PR, before change

In this repo under example/go/with_proto

Behavior: proto is a directory, it should be a package

image

### After the change

In this repo under example/go/with_proto

Behavior: proto is a package, as expected

image


blorente:
Tested on IJ UE 2023.2 and CLion 2023.2, and it works.
Happy to merge as-is, but leaving open for a couple of days in case @mai93 would like to give it a look.

blorente (approved):

mai93 (approved):

PR #5573, opened 2023-10-31
fix: Do not complete ignored directories

stevebarrau:
# Checklist

Discussed offline with @blorente.

# Description of this change

Do not complete ignored directories in a label in build files.


blorente (approved):
LGTM, we've tested internally and it makes results more accurate. Will merge in a couple of days.

Issue #5664, opened 2023-11-09
Exception during IJ startup with Bazel plugin when Bazel project opened by default


dkashyn-sfdc:
### Description of the bug:

```
[backend] java.lang.NullPointerException: Cannot invoke "com.google.idea.blaze.base.toolwindow.ConsoleView.getContent()" because the return value of "java.util.Map.get(Object)" is null
	at com.google.idea.blaze.base.toolwindow.TasksTreeConsoleView.treeSelectionChanged(TasksTreeConsoleView.java:95)
	at com.google.idea.common.ui.properties.Property.triggerChangeListeners(Property.java:97)
	at com.google.idea.common.ui.properties.Property.setValue(Property.java:80)
	at com.google.idea.blaze.base.toolwindow.TasksTreeView$TaskNodeAutoExpandingListener.treeNodesInserted(TasksTreeView.java:387)
	at com.intellij.util.ui.tree.TreeModelListenerList.treeNodesInserted(TreeModelListenerList.java:85)
	at com.intellij.ui.tree.AsyncTreeModel.treeNodesInserted(AsyncTreeModel.java:370)
	at com.intellij.ui.tree.AsyncTreeModel$CmdGetChildren.applyNodeToUiTree(AsyncTreeModel.java:746)
	at com.intellij.ui.tree.AsyncTreeModel$Command.applyToUiTree(AsyncTreeModel.java:499)
	at com.intellij.ui.tree.AsyncTreeModel.lambda$applyToUiTree$13(AsyncTreeModel.java:302)
	at com.intellij.util.concurrency.Invoker$Task.run(Invoker.java:381)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$1(CoreProgressManager.java:192)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$executeProcessUnderProgress$12(CoreProgressManager.java:610)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:685)
	at com.intellij.openapi.progress.impl.CoreProgressManager.computeUnderProgress(CoreProgressManager.java:641)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:609)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:78)
[...]

PR #5665, opened 2023-11-09
Work around NPE for toolwindow

dkashyn-sfdc:
# Checklist

details.

Issue number: [`5664`](https://github.com/bazelbuild/intellij/issues/5664)

# Description of this change


ujohnny:
So the workaround might lead to the bug that the most recent console might not be selected, but I think it's ok to have this behavior rather than have the exception. I was trying to find a root cause why the event comes before the console added, but the behavior is a bit obscure.

blorente:
I agree that it might cause a not-most-recent console to be selected, but also this would have been the behaviour if the NPE didn't raise, so I think it's a fine workaround.

blorente (approved):

dkashyn-sfdc:
In my case no user action was even required. I was able to repro issue both in local: during working with IJWB plugin itself and starting 2023.3 (it was not the thing before) and with Gateway mode.

In both cases we are launching IJ with project specified in CLI and by the time user opens UI there are such exception, without any user action.

Issue #5677, opened 2023-11-10
Starlark float literals are not supported


timothyg-stripe:
### Description of the bug:

Starlark has supported floating point numbers [since 2020](https://github.com/bazelbuild/starlark/pull/119), and Bazel [supported](https://github.com/bazelbuild/bazel/commit/21a8cf553c1cb315cf202ae492a5170f6adc8fb5) it starting in 4.0.0. Yet, the plugin does not yet support floating point literals (see screenshot below).

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

![image](https://github.com/bazelbuild/intellij/assets/140534739/850b6e59-5c40-4a1d-8421-9247e978f7a1)

### What programming languages and tools are you using? Please provide specific versions.

Starlark

### What Bazel plugin version are you using?

master

PR #5678, opened 2023-11-10
feat: Add support for Starlark float literals

timothyg-stripe:
# Checklist

details.

Issue number: #5677

# Description of this change

Add support for Starlark floating-point literals. Also add `float` as a built-in function, and consider float values primitives in the debugger.

Before:
not highlighted

After:
highlighted correctly

Much of the lexer implementation is ported from [starlark-go](https://github.com/google/starlark-go/blob/556fd59b42f68a2fb1f84957741b72811c714e51/syntax/scan.go#L909)

### DIFF
--- a/base/tests/unittests/com/google/idea/blaze/base/lang/buildfile/lexer/BlazeLexerTest.java
+++ b/base/tests/unittests/com/google/idea/blaze/base/lang/buildfile/lexer/BlazeLexerTest.java
@@ -63,28 +63,28 @@ public class BlazeLexerTest extends AbstractLexerTest {
     assertEquals("", names(tokens("")));
     assertEquals("", names(tokens("# foo")));
     assertEquals("INT INT INT INT", names(tokens("1 2 3 4")));
-    assertEquals("INT DOT INT", names(tokens("1.234")));
+    assertEquals("FLOAT", names(tokens("1.234")));
     assertEquals(
         "IDENTIFIER LPAREN IDENTIFIER COMMA IDENTIFIER RPAREN", names(tokens("foo(bar, wiz)")));
   }
 
   @Test
   public void testIntegersAndDot() throws Exception {
-    assertEquals("INT(1) DOT INT(2345)", values(tokens("1.2345")));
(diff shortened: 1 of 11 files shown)


base/tests/unittests/com/google/idea/blaze/base/lang/buildfile/lexer/BlazeLexerTest.java
```diff
@@ -63,28 +63,28 @@ public void testBasics4() throws Exception {
...
+    assertEquals("invalid float constant: 1.23E", lastError);
 
+    assertEquals("FLOAT(1.23E-10)", values(tokens("1.23E-10")));
     assertEquals("DOT INT(123)", values(tokens(". 123")));
-    assertEquals("DOT INT(123)", values(tokens(".123")));
+    assertEquals("FLOAT(0.123)", values(tokens(".123")));
     assertEquals("DOT IDENTIFIER(abc)", values(tokens(".abc")));
 
-    assertEquals("IDENTIFIER(foo) DOT INT(123)", values(tokens("foo.123")));
+    assertEquals("IDENTIFIER(foo) FLOAT(0.123)", values(tokens("foo.123")));
```
blorente, line 87: I'm not sure the spec covers this, but I think an identifier that ends in numbers should still parse the same way:

```suggestion
    assertEquals("IDENTIFIER(foo) FLOAT(0.123)", values(tokens("foo.123")));
    assertEquals("IDENTIFIER(foo456) FLOAT(0.123)", values(tokens("foo567.123")));
```

blorente (approved):
The changes in general look good, thanks for working on them! Added a corner case I'd like to test. If that passes, I think we're good to go.

base/tests/unittests/com/google/idea/blaze/base/lang/buildfile/lexer/BlazeLexerTest.java
timothyg-stripe: The spec says:

> Each token is formed from the longest sequence of characters that would form a valid token of each kind.

(i.e., the "maximal munch") so it does cover this.

Test added, passes :)

PR #5689, opened 2023-11-14
Allow to have non-existent import entries in `projectview` that pointing to …

dkashyn-sfdc:
…files that are optional and can be missing for some of the users.

# Checklist

details.

Issue number: `5688`

# Description of this change


base/tests/unittests/com/google/idea/blaze/base/projectview/parser/ProjectViewParserTest.java
```diff
@@ -303,11 +306,26 @@ public void testCanParseWithMissingCarriageReturnAtEndOfSection() {
 
   @Test
   public void testImportMissingFileResultsInIssue() {
+    Registry.get("bazel.projectview.optional.imports").setValue(false);
```
dkashyn-sfdc, line 309: I was unable to make it work with `Registry.is('bla', false)` for tests.

base/tests/utils/unit/com/google/idea/blaze/base/scope/ErrorCollector.java
```diff
@@ -55,6 +55,10 @@ public void assertHasErrors() {
     assertThat(issues.stream().anyMatch(i -> i.getCategory() == Category.ERROR)).isTrue();
   }
 
+  public void assertHasNoErrors() {
```
dkashyn-sfdc, line 58: in some cases there are issues but those are not errors.

base/src/com/google/idea/blaze/base/projectview/section/sections/ImportSection.java
```diff
@@ -50,7 +53,14 @@ protected WorkspacePath parseItem(
       if (parser.isRecursive()) {
         File projectViewFile = parseContext.getWorkspacePathResolver().resolveToFile(workspacePath);
         if (projectViewFile != null) {
-          parser.parseProjectView(projectViewFile);
+          boolean projectViewImportsMandatory = !Registry.is("bazel.projectview.optional.imports");
+          if (projectViewFile.exists() || projectViewImportsMandatory) {
```
ujohnny, line 57: I'd probably flip the argument for (a) readability though this is imo (b) less file operations as the value will be `true` by default.

ujohnny:
codewise lgtm, haven't checked how it works IRL. @tpasternak could you please check this? 

related though: I'd be happy to see warnings (primarily highlighting) in editor on unresolved imports when feature is enabled, though I believe it falls outside of the scope of proposed change.

dkashyn-sfdc:
sigh...

ujohnny (changes requested):
We had a discussion about this feature and the common point was that primarily in all programming languages imports require imported entity to exist. Though for example python allows to handle import errors, so can we keep `import` statement as is and just introduce `try_import` which will allow imported entity to be absent?

dkashyn-sfdc:
I don't think that `.projectview` is a programing language to be that strict :)

dkashyn-sfdc:
@ujohnny please note that this is protected by registry value and not default behavior.

Our use case is following:
1. we do have templates 
2. customers use those templates to create their projects
3. template changes over time and we delete/remove imports
4. files not there anymore and projects generated as part of `2` and oblivious to `3` still expecting files to be there when they are physically not

In our case we want to have freedom to remove any imported file without much deprecation process.
Every import that we have should be `try-import` based on what you are suggesting.

If we go via `try-import` way then we don't need that registry key.

Could you please consider approving this PR and if someone needs `try-import` they can add it in the future.

dkashyn-sfdc:
With your suggestion we need to ask EVERY engineer to replace `import` in their existing projects with `try-import` and this is what we wanted to avoid by enabling optional imports with warning without involvement of EVERY engineer who have their project already configured.

ujohnny:
@dkashyn-sfdc you could leave implicit optional import support under the flag, but please then make `try_import` statement available for everyone.

dkashyn-sfdc:
@ujohnny could you please check now? I'll think of other tests if the approach is correct.

base/src/com/google/idea/blaze/base/projectview/ProjectViewSet.java
```diff
@@ -52,10 +52,12 @@ public <T> List<T> listItems(SectionKey<T, ListSection<T>> key) {
   }
 
   /** Returns all values from all scalar sections in the project views, in order */
-  public <T> List<T> listScalarItems(SectionKey<T, ScalarSection<T>> key) {
+  public <T> List<T> listScalarItems(SectionKey<T, ScalarSection<T>>... keys) {
```
dkashyn-sfdc, line 55: not sure whether this is a good idea so can redo if someone have strong opinions against this way

ujohnny (approved):
LGTM. @tpasternak would you like to check this?

base/src/com/google/idea/blaze/base/wizard2/ui/BlazeEditProjectViewControl.java
```diff
@@ -599,6 +600,9 @@ public void updateBuilder(BlazeNewProjectBuilder builder) {
               .add(
                   ScalarSection.builder(ImportSection.KEY)
                       .set(selectProjectViewOption.getSharedProjectView()))
+              .add(
```
mai93, line 603: why is this needed? It results in having both import and try_import sections in the generated local projectview file.

dkashyn-sfdc: It is ok to use both `import` and `try_import`. It might be not ok to have some duplication or excessive defaults...

Could you please share more context for your inquiry?

mai93: I do not know why it is needed to add a `try_import` in this function? If I understand correctly, this function creates a project view that imports the shared project view file selected in the import process. So `import` section should be enough and there is no need to `try_import` the same project view.

dkashyn-sfdc: @mai93 there might be "mandatory" and "optional" imports in any project view. One example of optional would be ```
try_import:
  ~/.some.local.projectview

```

(let's assume `~` is resolved to user home dir ;) )

Comment (mai93) [reply base/src/com/google/idea/blaze/base/wizard2/ui/BlazeEditProjectViewControl.java:603]: I see, I'm still not sure what the effect of having something like:

```
import hello/hello.bazelproject
try_import hello/hello.bazelproject
```

will this provide a different result from only
```
import hello/hello.bazelproject
```

Comment (dkashyn-sfdc) [reply base/src/com/google/idea/blaze/base/wizard2/ui/BlazeEditProjectViewControl.java:603]: It is rather for
```
import hello/fail.if.not.present.bazelproject
try_import hello/ok.if.not.present.bazelproject
```

That can result in:
1. both imported
2. only `hello/fail.if.not.present.bazelproject` imported and accepting the fact that that `try_import hello/ok.if.not.present.bazelproject` is not there
3. failing if `hello/fail.if.not.present.bazelproject` is missing


Optional imports will allow engineers to have pre-defined custom import files both present or missing. 

Now it is possible for some of the users to benefit from custom contents while other users not affected even if they have no local customizations. For both groups of users the same `.bazelproject` can be stored in VCS now.

Comment (mai93) [reply base/src/com/google/idea/blaze/base/wizard2/ui/BlazeEditProjectViewControl.java:603]: so there is no additional value for having:
```
import hello/hello.bazelproject
try_import hello/hello.bazelproject
```
right? If so, then the change in this line can be reverted because it adds `try_import` section for the same project view file: `selectProjectViewOption.getSharedProjectView()` that already was added with `import` in line 600.

dkashyn-sfdc: Reverting here https://github.com/bazelbuild/intellij/pull/5958 @mai93

PR #5774, opened 2023-11-25
Fix Git detection on Mac OS and Windows

ahmedyarub:
# Checklist

Issue number: `5773`

# Description of this change
Adds quoting around the git command's parameters to avoid errors in some terminal shells.


ahmedyarub:
> This change makes the quotation mark being passed to git, it doesn't work. Maybe it's a shell issue, can you share what's your default shell?
> 
> BTW I'm getting `fatal: no such branch: '"'`

On Powershell:
`git rev-parse "@\{u\}"`

On Windows CMD:
`git rev-parse "@{u}"`

On MacOS' ZSH and Bash:
both `git rev-parse "@{u}`" and `git rev-parse "@{u}"`

ahmedyarub:
both in CMD and Powershell I get:

> jshell> Runtime.getRuntime().exec(new String[]{"git", "rev-parse", "\"@{u}\""}).getErrorStream().transferTo(System.out)
> fatal: no upstream configured for branch 'ay/fix_test_ui_mapping'
> $1 ==> 66
Which should be OK

On ZSH and Bash on MacOS I get:
> fatal: no such branch: '"'

ahmedyarub:
Added fallback for Windows and tested successfully.

ahmedyarub:
> So I tried it once again, and it seems to work correctly. This would make sense if we called the command via cmd or powershell. We would have to escape `{`, `}` characters as they are used to reference variables with special characters. But if we spawn subprocesses directly via JVM's, the arguments should be tokenized correctly.
> 
> My impression is that something is broken in your machine, or JVM. If this is not true then we would have to escape much more things, not only this single git rev-parse argument. Could you please double check on any other machine or VM?

I'll do that today.

ahmedyarub:
You are right, the problem was that I didn't have an upstream configured.

PR #5789, opened 2023-11-29
Support kt_android and kt_android_local_test

Bencodes:
The Bazel Intellij/Aswb plugins internally iterate over the `RuleTypes` enum values when deciding what target kinds to recognize. Example [KotlinBlazeRules](https://github.com/bazelbuild/intellij/blob/google/kotlin/src/com/google/idea/blaze/kotlin/KotlinBlazeRules.java#L60-L64)

`rules_kotlin` internally has native rules backing the `kt_jvm_` targets which is why targets of these rule kinds are recognized today.

The `kt_android_` rules in rules_kotlin are using a macro/rule "sandwich" where the top level `kt_android_` targets are just macros that eventually create a `android_library`/`android_local_test` target that's recognized by Aswb/ijwb.

We have a branch of `rules_kotlin` that drops the macro/rule sandwiches in favor of a single `kt_android_` rule. This pull request simply enables support for that branch allowing aswb/ijwb to recognize those proper `kt_android_` rule kinds, allowing features like resource resolution, running tests from the gutter, and debugging tests.

### DIFF
--- a/aspect/build_dependencies.bzl
+++ b/aspect/build_dependencies.bzl
@@ -680,7 +680,7 @@ FOLLOW_JAVA_ATTRIBUTES_BY_RULE_KIND = [
     ("_aspect_proto_toolchain_for_javalite", []),
     ("_aspect_java_proto_toolchain", []),
     ("runtime", ["proto_lang_toolchain", "java_rpc_toolchain"]),
-    ("_toolchain", ["_java_grpc_library", "_java_lite_grpc_library", "kt_jvm_library_helper", "android_library"]),
+    ("_toolchain", ["_java_grpc_library", "_java_lite_grpc_library", "kt_jvm_library_helper", "android_library", "kt_android_library"]),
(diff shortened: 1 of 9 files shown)


mai93:
thanks @Bencodes! Since `kt_android_library` and `kt_android_local_test` are defined as macros in https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/android.bzl, it is not clear why they need to be added as rule kinds in the plugin. Can you provide an example to show how targets defined by these macros are currently handled vs what the desired behaviour?

Bencodes:
@mai93 added a proper description. The tl;dr is we have a branch that swaps out the macros for proper rules. Without this PR `kt_android_` targets are no longer recognized by the IDE.

mai93 (approved):

mai93:
Imported in https://github.com/bazelbuild/intellij/commit/6f40cdd9eaf3cfa971cdf56d3e4fd43f75ff3eab

PR #5816, opened 2023-12-06
feat: Enable debugging the gazelle binary

stevebarrau:
Allow debugging gazelle (use the `gazelle_binary` target and not the `gazelle` one).


blorente:
LGTM. We checked internally, and it allows you to debug a gazelle target with a custom plugin from your own repository, which is a huge boon to gazelle plugin developers.

Approving, but it's in such a core place that I'd like @mai93 's opinion.

blorente (approved):

PR #5873, opened 2023-12-19
Fix warnings when syncing a workspace containing a bazelignore file

srdo-humio:
The warnings look like this

WARNING: Pattern '//dist/...:all' was filtered out by ignored directory 'dist'

They occur because the plugin passes paths from bazelignore into "bazel query" as exclusions. Bazel warns about this, as excluding a path that Bazel already knows should be ignored is likely a user error.

The reason bazelignore paths are ending up in the query is that we want the IDE to mark the paths as excluded from the project. For that reason, the plugin parses the bazelignore file, and adds the paths to an exclusion list. That list happens to also be passed to "bazel query".

This commit keeps bazelignore paths in the general exclusion list, but generates a separate list of paths to use for "bazel query", which only contains exclusions not listed in bazelignore.

### DIFF
--- a/base/src/com/google/idea/blaze/base/dependencies/BlazeQueryDirectoryToTargetProvider.java
+++ b/base/src/com/google/idea/blaze/base/dependencies/BlazeQueryDirectoryToTargetProvider.java
@@ -40,6 +40,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.List;
+import java.util.Set;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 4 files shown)


mai93:
thanks @srdo-humio! the PR looks good to me but can you add tests to cover the new behaviour?

srdo-humio:
Yes, I assume you just mean tests that the query string is now excluding bazelignored files? I'll try to add a test for that.

srdo-humio:
Added a test. Let me know if you think the testing needs to be more extensive.

mai93 (approved):

mai93:
That looks good, thanks @srdo-humio!

PR #5885, opened 2023-12-20
feat: Allow environment variables in run configurations

blorente:
# Checklist

details.

# Description of this change

Allow users to set their own env vars per run-configuration. It is not respected everywhere (see the list below), but it works on our example projects.

Main features:
- Env vars should work when **run**ing and **debug**ging any binary.
- Env vars should work when **run**ning any test.
- Env vars should work when **debug**ging Go and Java tests. Env vars for C++ test debugging are implemented internally, but they're part of a larger patch and I'll need some more work to untangle that (or maybe they're already covered in https://github.com/bazelbuild/intellij/pull/5771)

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
+++ b/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
@@ -81,6 +81,7 @@ import com.intellij.openapi.project.Project;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -192,9 +193,12 @@ public final class BlazeCommandGenericRunConfigurationRunner
     private ProcessHandler getScopedProcessHandler(
         Project project, BlazeCommand blazeCommand, WorkspaceRoot workspaceRoot)
         throws ExecutionException {
+      GeneralCommandLine commandLine = new GeneralCommandLine(blazeCommand.toList());
+      @NotNull Map<String, String> envVars = handlerState.getUserEnvVarsState().getData().getEnvs();
+      commandLine.withEnvironment(envVars);
       return new ScopedBlazeProcessHandler(
           project,
-          new GeneralCommandLine(blazeCommand.toList()),
+          commandLine,
           workspaceRoot,
           new ScopedBlazeProcessHandler.ScopedProcessHandlerDelegate() {
             @Override
@@ -239,13 +243,14 @@ public final class BlazeCommandGenericRunConfigurationRunner
           });
       addConsoleFilters(consoleFilters.toArray(new Filter[0]));
 
+      @NotNull Map<String, String> envVars = handlerState.getUserEnvVarsState().getData().getEnvs();
       ListenableFuture<BlazeBuildOutputs> blazeBuildOutputsListenableFuture =
           BlazeExecutor.getInstance()
               .submit(
                   () ->
                       invoker
                           .getCommandRunner()
-                          .run(project, blazeCommandBuilder, buildResultHelper, context));
+                          .run(project, blazeCommandBuilder, buildResultHelper, context, envVars));
       Futures.addCallback(
           blazeBuildOutputsListenableFuture,
           new FutureCallback<BlazeBuildOutputs>() {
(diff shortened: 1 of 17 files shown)


blorente:
This looks like a smallish change, but it's in a deep part of the codebase. I've tested it locally on the workflows listed, but please do give it a shot. It can use as many eyes as we can give it.

golang/src/com/google/idea/blaze/golang/run/BlazeGoRunConfigurationRunner.java
```diff
@@ -141,7 +141,7 @@ private ImmutableList<String> getParameters(ExecutableInfo executable) {
       return ImmutableList.<String>builder()
           .addAll(state.getExeFlagsState().getFlagsForExternalProcesses())
           .addAll(state.getTestArgs())
-          .addAll(executable.args)
+          .addAll(state.getUserEnvVarsState().asBlazeTestEnvFlags())
```
mai93, line 144: I think `ExecutableInfo` already has a field `envVars`, will it contain different variables from `state.getUserEnvVarsState().asBlazeTestEnvFlags()`?

Why do we no longer need the `executable.args`? and also we can remove the `executable` parameter of this function if it is not needed.

java/src/com/google/idea/blaze/java/run/BlazeJavaRunProfileState.java
```diff
@@ -280,9 +285,12 @@ static BlazeCommand.Builder getBlazeCommandBuilder(
   private ProcessHandler getScopedProcessHandler(
       Project project, ImmutableList<String> command, WorkspaceRoot workspaceRoot)
       throws ExecutionException {
+    GeneralCommandLine commandLine = new GeneralCommandLine(command);
+    Map<String, String> envVars = getState(getConfiguration()).getUserEnvVarsState().getData().getEnvs();
```
mai93, line 289: same as before: I'm not sure if this is needed but in https://github.com/ujohnny/intellij/blob/master/clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java#L239 the ParentEnvironmentType is set. do we need to do something similar here?

mai93:
I manually tested it for the java example project, it was working for running and debugging java_binary but did not work for running java_test.

mai93 (commented):
I manually tested it for Java, but I think we need to also add unit tests.

golang/src/com/google/idea/blaze/golang/run/BlazeGoRunConfigurationRunner.java
blorente: The envVars from `ExecutableInfo` are just for the test runner itself, unrelated to the ones the user specifies. However, this comment uncovers a massive bug in the implementation, thanks for pointing it out. I'll fix it.

base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
```diff
@@ -311,32 +316,42 @@ protected ConsoleView createConsole() {
         context.addOutputSink(PrintOutput.class, new WritingOutputSink(consoleView));
       }
       addConsoleFilters(consoleFilters.toArray(new Filter[0]));
-      return !invoker.getCommandRunner().canUseCli()
-          ? getCommandRunnerProcessHandler(
+      @NotNull Map<String, String> envVars = handlerState.getUserEnvVarsState().getData().getEnvs();
+
+      if (invoker.getCommandRunner().canUseCli()) {
+        // BlazeCommandRunner is already supposed to handle the cases where we pass env vars to tests
```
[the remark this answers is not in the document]
blorente: Sorry for the confusion. I have clarified the comment, hopefully it helps explain the weirdness.

blorente (commented):
@mai93 Thanks for the comments! I've tried to address the inline ones there, but for the broader points:

> I manually tested it for Java, it is not working when running a single test method but works when running the whole test class.

I've found that it works on a single method just fine (as long as I add the env vars to the new config as well). I've added example run configurations to the `greetings_project` to demonstrate, but please let me know if you still see that.

>  I also think we need to also add unit tests

I haven't done this yet, but will look into where it makes sense to add them. Thanks!

blorente:
I have added a few unit tests, particularly for java tests. Unfortunately, I haven't seen any unit or integration tests for the biggest changes (namely, how the processes are actually spawned).
Please let me know if you'd like me to add more tests somewhere else.

mai93 (commented):
I left small comments and I tested it manually on Java, it is working fine. Thanks @blorente!

examples/go/with_proto/with_proto.bazelproject
```diff
@@ -11,5 +11,5 @@ gazelle_target: //:gazelle
 import_run_configurations:
   # Test that validates that macro expansion works, as per issue:
   #    https://github.com/bazelbuild/intellij/issues/4112#event-7958662669
-  # This configuration should be executed on both 'run' (the green arrow) and 'debug' (the bug) modes.
+  # This configuration should only be executed on both 'run' (the green arrow) mode.
```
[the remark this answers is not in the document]
blorente: I should have written a better comment, sorry about that.
The main issue is that, because of how we pass env vars to debuggers, the IntelliJ macros (e.g. `$ProjectName$`) are not expanded. So, the env vars get passed, but the macros don't get expanded.

I've added a new run configuration to test env vars, which works on both run and debug.

blorente:
@mai93 Thanks for the review! Comments addressed, please take a look when you have time.

mai93 (approved):

mai93:
Thanks @blorente! LGTM, leaving the merge for you if you want someone else to take a look as well.

blorente:
Given the amount of time this has been active internally I'm pretty confident of merging after your review, but I'm happy to work on fixing issues (or writing documentation about where env vars do and don't work).

Issue #5909, opened 2024-01-03
Spurious error logs in bazel query when using non-default convenience symlinks


srdo-humio:
### Description of the bug:

Bazel allows overriding the names of the default symlinks (e.g. "bazel-bin") via --symlink_prefix, or omitting them entirely via --experimental_convenience_symlinks=ignore.

When a workspace does this, the default symlinks with names like bazel-out
and bazel-bin don't exist. This causes Bazel to emit an error when running
"bazel query", since the plugin hardcodes these names as excluded paths.

Syncing still works, but the errors are noisy.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Add the following to .bazelrc in a sync'able workspace

```
common --symlink_prefix=dist/
```

Open IJ and press the sync button. You will see errors like

```
ERROR: Skipping '//bazel-bin/...:all': no targets found beneath 'bazel-bin'
```

### Which Intellij IDE are you using? Please provide the specific version.

Intellij Community 2023.3.2

### What programming languages and tools are you using? Please provide specific versions.

Scala

### What Bazel plugin version are you using?

2023.12.19.0.1-api-version-233

srdo-humio:
I can see a couple of ways we could address this:

We can exclude paths from bazel query that don't exist.

We could try to extract the values for --symlink_prefix and --experimental_convenience_symlinks from Bazel, and make the plugin adjust the directories included by default by BazelBuildSystemProvider.buildArtifactDirectories

I think excluding paths that don't exist should be harmless, so the linked PR does the former. Doing the latter is more difficult, and I think also less robust, since the plugin will need to know about and react to all flags that affect the convenience symlinks.

mai93:
Thanks @srdo-humio! https://github.com/bazelbuild/intellij/pull/5910 is merged

PR #5910, opened 2024-01-03
Fix bazel sync errors when using non-default convenience symlinks

srdo-humio:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/5909

# Description of this change

Bazel allows overriding the names of the default symlinks via --symlink_prefix, or omitting them entirely via --experimental_convenience_symlinks=ignore.

When a workspace does this, the default symlinks with names like bazel-out and bazel-bin don't exist. This causes Bazel to emit an error when running "bazel query", since the plugin hardcodes these names as excluded paths.

This commit addresses this by only including paths in "bazel query" that actually exist. Workspaces that change the symlink names can then exclude the new names explicitly via bazelignore.

### DIFF
--- a/base/src/com/google/idea/blaze/base/dependencies/BlazeQueryDirectoryToTargetProvider.java
+++ b/base/src/com/google/idea/blaze/base/dependencies/BlazeQueryDirectoryToTargetProvider.java
@@ -39,6 +39,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.nio.file.Files;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


srdo-humio:
I can see a couple of ways we could address this:

* We can exclude paths from `bazel query` that don't exist. 

* We could try to extract the values for --symlink_prefix and --experimental_convenience_symlinks from Bazel, and make the plugin adjust the directories included by default by `BazelBuildSystemProvider.buildArtifactDirectories`

I think excluding paths that don't exist should be harmless, so this PR does the former. Doing the latter is more difficult, and I think also less robust, since the plugin will need to know about and react to all flags that affect the convenience symlinks.

Issue #5300, opened 2023-09-01
CLion can't use include_prefix and strip_include_prefix


yvvan:
### Description of the bug:

CLion needs information in order to properly support `include_prefix` and `strip_include_prefix`. If they are provided e.g. in form of flags CLion could support such paths even when symlinks don't exist for some reason (for example after some resync).

The related CLion issue - https://youtrack.jetbrains.com/issue/CPP-34318, the support will be added to the next 2023.2 CLion patch release.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Create targets which use `include_prefix` and remove the symlinks after Bazel sync. CLion won't be able to work with such included headers.

### Which Intellij IDE are you using? Please provide the specific version.

CLion 2023.2

### What programming languages and tools are you using? Please provide specific versions.

C++

### What Bazel plugin version are you using?

latest

PR #5969, opened 2024-01-19
Fix _virtual_includes handling for absolute strip_include_prefix paths

sfc-gh-abalik:
# Checklist

- [ ] I have filed an issue about this change and discussed potential changes with the maintainers.
- [ ] I have received the approval from the maintainers to make this change.
- [ ] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: `<please reference the issue number or url here>`

# Description of this change

PR #6007, opened 2024-02-01
Followup for Pass path associations as special compiler flags

LeFrosch:
# Checklist

details.

Issue number: #5300

# Description of this change
Followup for #5301. Pass all headers in targets with `include_prefix` with special "-ibazel" flag like that they can be handled by CLion. Also I tried to address all open comments in the pull request.

### DIFF
--- a/base/src/com/google/idea/blaze/base/model/primitives/Label.java
+++ b/base/src/com/google/idea/blaze/base/model/primitives/Label.java
@@ -107,7 +107,7 @@ public final class Label extends TargetExpression {
     }
     int slashesIndex = label.indexOf("//");
     logger.assertTrue(slashesIndex >= 0);
-    return label.substring(0, slashesIndex).replaceFirst("^@+", "");
+    return label.substring(1, slashesIndex);
   }
 
   /**
--- a/base/src/com/google/idea/blaze/base/model/primitives/Label.java
+++ b/base/src/com/google/idea/blaze/base/model/primitives/Label.java
@@ -107,7 +107,7 @@ public final class Label extends TargetExpression {
     }
     int slashesIndex = label.indexOf("//");
     logger.assertTrue(slashesIndex >= 0);
-    return label.substring(1, slashesIndex);
+    return label.substring(0, slashesIndex).replaceFirst("^@+", "");
   }
 
   /**
--- a/cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
+++ b/cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
@@ -22,7 +22,11 @@ import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.errorprone.annotations.Keep;
-import com.google.idea.blaze.base.ideinfo.*;
+import com.google.idea.blaze.base.ideinfo.ArtifactLocation;
+import com.google.idea.blaze.base.ideinfo.CIdeInfo;
+import com.google.idea.blaze.base.ideinfo.Dependency;
+import com.google.idea.blaze.base.ideinfo.TargetIdeInfo;
+import com.google.idea.blaze.base.ideinfo.TargetKey;
 import com.google.idea.blaze.base.model.BlazeProjectData;
 import com.google.idea.blaze.base.model.primitives.ExecutionRootPath;
 import com.google.idea.blaze.base.model.primitives.WorkspaceRoot;
@@ -70,7 +74,13 @@ import com.jetbrains.cidr.lang.workspace.compiler.TempFilesPool;
 import org.jetbrains.annotations.NotNull;
 
 import java.io.File;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Collections;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.stream.Stream;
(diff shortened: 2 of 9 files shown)


cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -165,6 +172,97 @@ public void run(ProgressIndicator indicator) {
...
+    }
+
+    try {
+      return Path.of(value);
+    } catch (InvalidPathException e) {
+      return null;
+    }
+  }
+
+  private List<String> collectIncludes(Path root, TargetKey targetKey, BlazeProjectData blazeProjectData) {
```
ujohnny, line 195: I'd pass indicator to this routine as well and performed a check `indicator.checkCanceled()` in both `for` loops.

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -165,6 +172,97 @@ public void run(ProgressIndicator indicator) {
...
+
+  private @Nullable Path pathOf(@Nullable String value) {
+    if (value == null || value.isEmpty()) {
+        return null;
+    }
+
+    try {
+      return Path.of(value);
+    } catch (InvalidPathException e) {
+      return null;
```
ujohnny, line 191: It should be logged, but please `warn`.

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
ujohnny, line 1: some indentations are wrong in this file in `if` blocks, please make sure you have style guide applied

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -257,6 +355,9 @@ private OCWorkspaceImpl.ModifiableModel calculateConfigurations(
                 .map(file -> "-I" + file.getAbsolutePath())
                 .collect(toImmutableList());
 
+        Path rootPath = workspaceRoot.directory().toPath();
+        List<String> includes = collectIncludesWithProgress(rootPath, targetKey, blazeProjectData, indicator);
```
ujohnny, line 359: these `includes` are quite specific ones. should we say something like `collectBazelStripIncludeHints`?

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -165,6 +172,97 @@ public void run(ProgressIndicator indicator) {
...
+    ArrayList<String> includes = new ArrayList<>();
+    for (ArtifactLocation header : cIdeInfo.getHeaders()) {
+      Path realPath = root.resolve(header.getExecutionRootRelativePath());
+
+      Path codePath = pathOf(header.getRelativePath());
+      if (codePath == null) {
+          continue;
+      }
+
+      // if absolut strip prefix is a repository-relative path
```
ujohnny, line 216: typo

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -165,6 +172,97 @@ public void run(ProgressIndicator indicator) {
...
+
+  private List<String> collectIncludes(Path root, TargetKey targetKey, BlazeProjectData blazeProjectData) {
+    TargetIdeInfo targetIdeInfo = blazeProjectData.getTargetMap().get(targetKey);
+    if (targetIdeInfo == null || targetIdeInfo.getcIdeInfo() == null) {
+      return Collections.emptyList();
+    }
+
+    CIdeInfo cIdeInfo = targetIdeInfo.getcIdeInfo();
+    Path includePrefix = pathOf(cIdeInfo.getIncludePrefix());
+    Path stripPrefix = pathOf(cIdeInfo.getStripIncludePrefix());
```
ujohnny, line 203: related to #5969, strip prefix might be `//foo/bar` and this code will break on Windows (yes we do not support it, though it works and maybe we should not introduce new issues).

ujohnny:
I remember when I intentionally added a check in [VirtualIncludesHandler.java](https://github.com/bazelbuild/intellij/blob/0630feef3f1b4e83a42dc9b57c64da9a51e2cb20/base/src/com/google/idea/blaze/base/sync/workspace/VirtualIncludesHandler.java#L88) to handle this. Something should be definitely done to it, either removed or message there should be updated.

ujohnny (changes requested):

base/src/com/google/idea/blaze/base/model/primitives/Label.java
```diff
@@ -107,7 +107,7 @@ public String externalWorkspaceName() {
     }
     int slashesIndex = label.indexOf("//");
     logger.assertTrue(slashesIndex >= 0);
-    return label.substring(1, slashesIndex);
+    return label.substring(0, slashesIndex).replaceFirst("^@+", "");
```
ujohnny, line 110: I believe there's something wrong here, this line probably should not be changed since in bazel external workspace name should has only one `@` symbol

ujohnny (changes requested):

base/src/com/google/idea/blaze/base/model/primitives/Label.java
ujohnny: I see that it's the way bazel 7 started to report external workspace targets. Investigating.

ujohnny: So my conclusion here is the following:

1. Please drop this change from this patchset.
2. Create new issue describing issue about bazel 7 -- this code works fine with bazel 6. 
3. Please check any other occurrences of `@` strip and adjust them. 
4. Should we introduce some special function that expects exactly 1 or exactly 2 `@` signs?

ujohnny (changes requested):
Left a comment in this thread
https://github.com/bazelbuild/intellij/pull/6007/files#r1514537317

ujohnny (approved):
lgtm

Issue #4431, opened 2023-02-12
A hyperlink in terminal is wrong for a  target which has dot in name


stle85:
### Description of the bug:

The links with dot in name are highlighted wrong in Terminal.
This links can be used to quick navigate to target by left button click on it.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Clone repository 
https://github.com/stle85/bazel-idea-hyperlink-issue

Run build in CLion terminal  `bazel build //...` 

Link on target in terminal has wrong link(or was highlighted wrong).

### Which Intellij IDE are you using? Please provide the specific version.

CLion 2022.3.2

### What programming languages and tools are you using? Please provide specific versions.

C++, clang

### What Bazel plugin version are you using?

2023.01.24.0.1-api-version-223

PR #6021, opened 2024-02-05
Changes to BlazeTargetFilter regex to match more target names

LeFrosch:
# Checklist

details.

Issue number: #4431

# Description of this change
Changed the regex used by BlazeTargetFilter to include more valid target names. I used [this](https://bazel.build/concepts/labels#target-names) documentation page to lookup the valid characters.

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/filter/BlazeTargetFilter.java
+++ b/base/src/com/google/idea/blaze/base/run/filter/BlazeTargetFilter.java
@@ -40,9 +40,9 @@ import javax.annotation.Nullable;
 /** Parse blaze targets in streamed output. */
 public class BlazeTargetFilter implements Filter {
 
-  // See Bazel's LabelValidator class. Whitespace character intentionally not included here.
-  private static final String PACKAGE_NAME_CHARS = "a-zA-Z0-9/\\-\\._$()";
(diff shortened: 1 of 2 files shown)


mmendonca2:
I think this issue created a problem where a top level exclusion "-someFolder" is now excluding that folder name from the entire project, no matter if it's nested in an explicitly included target. So now "includedTarget/someFolder" is excluded as well. It's problematic when explicitly excluding top level projects from .bazelproject whose name can show up in other targets as a package folder, as it breaks their builds.

LeFrosch:
Hi @mmendonca2, as far as I can tell the `BlazeTargetFilter` is only registered using a `consoleFilterProvider` and not used any where else. Therefore, I think it is only used for highlighting links in the console. 
If this is not the case and you are sure this is related to these changes please let me know. Otherwise, if this is a new issue please create an issue.

ujohnny:
I guess it is caused by #6094, @tpasternak haven't you mentioned something about this scenario to me several weeks ago?

LeFrosch:
@ujohnny @tpasternak I can confirm, the exclusion of subdirectories @mmendonca2 mentioned seems to be caused by #6094.

Also, setting `use_exclusion_patterns: false` looks like a valid workaround.

PR #6164, opened 2024-02-27
fix: Enviroment Variables are not correctly inherited from system on …

### LINKED ISSUE #6163: Env vars are lost in bazel run configurations
### Description of the bug:

1. Run any code printing environment variables using bazel's run configuration
2. Run a shell script using shell run configuration and print the same variables

### Which category does this issue belong to?

_No response_

_No response_

_No response_

_No response_

_No response_

_No response_

_No response_

tpasternak:
…MacOS

Use CONSOLE ParentEnvironmentType instead of SYSTEM

CONSOLE is the default value initialized in GeneralCommandLineso it was used before. Now, if we want to inherit the system envs we use SYSTEM, so in both cases it differs from what was there before. This PR just brings the old behavior in case a user wants to inherit enviroment from the system.

https://github.com/JetBrains/intellij-community/blob/9838665ba816fc8ce06e4071aa5a906a35fbb22e/platform/platform-util-io/src/com/intellij/execution/configurations/GeneralCommandLine.java#L85C1-L86C1

fixes #6163

# Checklist

details.

# Description of this change

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
+++ b/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
@@ -199,7 +199,7 @@ public final class BlazeCommandGenericRunConfigurationRunner
       commandLine.withEnvironment(envVarState.getEnvs());
       commandLine.withParentEnvironmentType(
               envVarState.isPassParentEnvs()
-                      ? GeneralCommandLine.ParentEnvironmentType.SYSTEM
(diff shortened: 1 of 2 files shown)


blorente (approved):
Thanks for the swift change!

blorente:
I would like it to be SYSTEM if the user ticks the "inherit" box, but I agree that right now there's a bug and the this gets people unblocked. I'm happy to iterate on making it work properly later.

blorente:
Ideally, we'd toggle between SYSTEM and CONSOLE, depending on whether the toggle was clicked or not.

blorente:
Let's merge and hot fix, I'll iterate in the morning

PR #6210, opened 2024-03-04
fix(sync): Disable query-sync by default

blorente:
# Checklist

details.

# Description of this change

Recently, upstream has started rolling out a new way of syncing a project (https://github.com/bazelbuild/intellij/pull/5646). This is still in beta, and we've seen some issues reported internally.

We disable it for now, until they stabilize the feature. Users that want to try it can still do so, from their settings menu.

### DIFF
--- a/base/src/com/google/idea/blaze/base/qsync/QuerySync.java
+++ b/base/src/com/google/idea/blaze/base/qsync/QuerySync.java
@@ -54,10 +54,7 @@ public class QuerySync {
   private QuerySync() {}
 
   public static boolean useByDefault() {
-    // We disable query sync by default until the feature is stable the feature.
-    // Note that users can still enable the feature via their settings menu (Cmd+, -> Bazel -> Query Sync)
-    //    return ENABLED.isEnabled();
-    return false;
+    return ENABLED.isEnabled();
   }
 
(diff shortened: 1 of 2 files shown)


mai93:
Thanks @blorente. What is strange is that in the feature is guarded with rollout experiment [here]( https://github.com/bazelbuild/intellij/blob/8e0658655c155ffbdd3ddf02dceb5a78d43f8fed/base/src/com/google/idea/blaze/base/qsync/QuerySync.java#L29) and `query.sync` is not set in [the experiments file]( https://github.com/bazelbuild/intellij/blob/master/common/experiments/src/com/google/idea/common/experiments/experiment.properties). I am not sure about the default behaviour but maybe we can add it there like `query.sync=0`.

blorente:
@mai93 I didn't know we had an experiments file :D 
I've changed the value, let me know how it looks now.
I get the value `disabled` from [here](https://github.com/bazelbuild/intellij/blob/73ba462c91dc5f414eaed15d1cfd7256463556fd/common/experiments/src/com/google/idea/common/experiments/FeatureRolloutExperiment.java#L46).

mai93 (approved):

mai93:
Thanks @blorente!  I think we need to import it internally to be disabled for ASwB as well.

PR #6218, opened 2024-03-05
Cherry-pick "Prevent the bazel plugin from being used for non-Android C++ development in Android Studio."

mai93:
original commit: https://github.com/bazelbuild/intellij/commit/28823de2b987213e3a5eba47c5a77345d83b4b22


mai93:
I wanted to get @ujohnny review on this

cpp/src/com/google/idea/blaze/cpp/CppSupportChecker.java
```diff
@@ -0,0 +1,43 @@
...
+
+/** Checks whether or not the IDE supports a given CPP configuration. */
+public interface CppSupportChecker {
+
+  ExtensionPointName<CppSupportChecker> EP_NAME =
+      new ExtensionPointName<>("com.google.idea.blaze.cpp.CppSupportChecker");
+
+  Logger logger = Logger.getInstance(CppSupportChecker.class);
+
+  static boolean isSupportedCppConfiguration(List<String> compilerSwitches, Path workspaceRoot) {
```
ujohnny, line 32: Have 2 comments to this implementation

1. Basically it's a `return A || B || C || ...` what looks slightly loose to me. I guess if at least 1 checker disallows then it should return false. So it should be `A && B && C`. 
2. Performance note: to get `true` for clwb we copy list of parameters here and do some extra logic to get compiler flags in workspace/qsync code. And it happens for every resolve configuration. I would consider changing the interface from the current one into something like

```
boolean requiresAdditionalArgumentsCheck(); 
boolean isSupportedCppConfiguration(...);
```

or maybe the check can be incapsulated with the something like

```
isSupportedCppConfiguration(CidrCompilerSwitches switches)
```

ujohnny (changes requested):

cpp/src/com/google/idea/blaze/cpp/CppSupportChecker.java
mai93: @mathewinwood as the author of the original commit. Can you address these comments?

mathewinwood: > Basically it's a return `A || B || C || ... ` 
True, but I don't think it would make any difference since currently there's at most 1 implementation per IDE configuration.

>  for clwb we copy list of parameters here and do some extra logic to get compiler flags 

Fair enough, I can change this.

ujohnny: > True, but I don't think it would make any difference since currently there's at most 1 implementation per IDE configuration.

This leaves a lot of space to abuse, it's so easy to add a new extension which just overrides NDK checker. I'd prefer to see `AND` there. 

> Fair enough, I can change this.

Thank you

mathewinwood: Ok, I see. I can update it to be an AND .

ujohnny:
@mai93 I'm still waiting for 1 more commit to arrive. https://github.com/bazelbuild/intellij/pull/6218#discussion_r1516446732

cpp/src/com/google/idea/blaze/cpp/CppSupportChecker.java
ujohnny: Not sure if we on the same page here, though I'm expecting the change that changes OR to AND. Is my expectation correct?

mathewinwood: That change is still in progress (sorry for the delay).

mai93:
> @mai93 I'm still waiting for 1 more commit to arrive. [#6218 (comment)](https://github.com/bazelbuild/intellij/pull/6218#discussion_r1516446732)

this is included now

ujohnny (approved):
lgtm except merge commits

Issue #6260, opened 2024-03-12
Python debugger shows a wrong file when stopped at a breakpoint


odisseus:
### Which category does this issue belong to?

Intellij, CLion

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Clone a Python project that has tests, for example https://github.com/christianfds/article-simple-bazel
2. Import it into IntelliJ with the following project view file:
```
directories:
  .

derive_targets_from_directories: true

targets:

additional_languages:
  python
```
3. Open a file with tests, e. g. `libs/complex_calculations/tests/test.py`
4. Follow the IntelliJ prompts to set up a Python venv and install the project requirements.
5. Run any test case in the file to make sure that the project is set up correctly.
6. Place a breakpoint inside the test case, and launch it with debugging.

**Expected behaviour:** The execution is suspended, and the line with the breakpoint is highlighted

**Actual behaviour:** A new file tab is opened, which is located somewhere in the Bazel execroot, e. g. `/private/var/tmp/_bazel_foo/4cd9340ff2eef52aa6659fe37648ff08/execroot/simple-bazel/bazel-out/darwin-fastbuild/bin/libs/complex_calculations/test.runfiles/simple-bazel/libs/complex_calculations/tests/test.py`. This is actually a symlink pointing to the original source file. The execution is suspended and the line that corresponds to the breakpoint is highlighted, but this happens in the wrong tab.

### Which Intellij IDE are you using? Please provide the specific version.

Idea IC-233.14475.28, also reproducible in Clion

### What programming languages and tools are you using? Please provide specific versions.

Python 3

### What Bazel plugin version are you using?

Locally built, commit d24616983c4027c090709c07febebc2188ec7f3c

andersdyhrnorlov:
What needs to happen before this fix can be applied, I've been waiting for this to be fixed.

tpasternak:
@odisseus are you interested in having this done?

PR #6283, opened 2024-03-18
Path translations for the Python debugger command line

odisseus:
# Checklist

details.

Issue number: #6260 

# Description of this change

When there is some Python code to be debugged, the plugin first builds the corresponding target, which (as far as I understand) involves copying some Python source files into some directory inside `bazel-bin`. The plugin also creates a script which invokes those files, and passes that script to pydevd (which is bundled with the python-ce plugin). 

The problem is, pydevd knows nothing about the Bazel workspace and its source root; from its point of view, the source root is somewhere in the Bazel output directory. When asked to stop at a breakpoint inside some source file, pydevd accepts its _relative_ path within the source root, and thus the breakpoint is places in the right location; however, when reporting back to IntelliJ, pydevd uses the _absolute_ path of the file, which of course doesn't match the path to the original source file.

Fortunately, pydevd allows to define path translations by setting the environment variable called `PATHS_FROM_ECLIPSE_TO_PYTHON`. When the environment is like `PATHS_FROM_ECLIPSE_TO_PYTHON=[["X", "Y"]]`, pydevd will report stopping at X whenever it actually stops at Y. (This feature is actually intended for debugging code that runs on a remote computer, but is is also suitable for our purposes).

This is an attempt to set up the right path mapping for the particular target that's being debugged. The logic can be summarised as follows:

1. Find the executable script which has been generated for our Bazel target.
2. Parse the path to the executable script in order to determine the workspace name.
3. Find the relative path from the workspace root to the package that contains our target.
4. Use the workspace name, the package's relative path, and the target name to guess the actual source root of the files that are processed by the debugger. 

In this PR, I have attempted to minimise the changes on the existing codebase of the plugin. I realise that this PR needs work before it can be merged into master, but I hope to get some early feedback, especially from people which are more familiar with the Python debugger.

I have tested my code manually on some Python tests, and it worked.

### DIFF
--- a/python/src/META-INF/blaze-python.xml
+++ b/python/src/META-INF/blaze-python.xml
@@ -19,4 +19,7 @@
     <TargetKindProvider implementation="com.google.idea.blaze.python.PythonBlazeRules"/>
     <BlazePyExecutableOutputFinder implementation="com.google.idea.blaze.python.run.BlazePyExecutableOutputFinderImpl"/>
   </extensions>
+  <extensions defaultExtensionNs="com.google.idea.blaze">
+    <BlazePyDebugFlagsProvider implementation="com.google.idea.blaze.python.run.BlazePyDebugHelperImpl"/>
+  </extensions>
 </idea-plugin>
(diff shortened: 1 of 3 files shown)


jastice:
since it's not directly mergable, let's close. Python debugging issues will be addressed in new plugin

Issue #6086, opened 2024-02-15
Publish render jars builder


idanakav:
### Description of the feature request:

ASWB is [using](https://github.com/bazelbuild/intellij/blob/5d612affff49b51d025c4f416656e11fecfa3641/aswb/sdkcompat/as223/com/google/idea/blaze/android/projectsystem/RenderJarClassFileFinder.java#L70) render jars for previews in Android Studio which is not available in the public repo.

Build usage can be seen in [BazelRenderJarBuilder](https://github.com/bazelbuild/intellij/blob/5d612affff49b51d025c4f416656e11fecfa3641/base/src/com/google/idea/blaze/base/qsync/BazelRenderJarBuilder.java#L76) and the aspect [semantics](https://github.com/bazelbuild/intellij/blob/5d612affff49b51d025c4f416656e11fecfa3641/aspect/intellij_info.bzl#L51) which does not define `build_render_resolve_jar`.

The ask is to get more info about render jars and to see if it's possible to share the implementation.

### What operating system, Intellij IDE and programming languages are you using? Please provide specific versions.

Android Studio

mai93:
Thanks @idanakav. @sgowroji can you import this internally?

sgowroji:
Imported internal with Android Team. Thanks!

PR #6315, opened 2024-03-23
Compose Preview Support

### LINKED ISSUE #2765: Support Jetpack Compose previews in Android Studio
Currently when you import an Android Studio project that builds an app using Jetpack Compose UI, it will fail to display compose previews.

The expected behavior is that Android Studio should recognize when a Kotlin file contains a composable `@Preview` and open the compose designer plugin shown [here](https://developer.android.com/jetpack/compose/tooling#preview).

idanakav:
Resolves #2765

Add support for Compose Preview in Android Studio, main changes are:
- Introduce a new `ClassFileFinder`, optimized for faster class lookups 
- Support reading and routing Android assets, such as fonts from the plugin aspect and pass them to Android Studio to be used in the rendering

Tested with the latest stable version, Iguana (23.2), it scales well even for our larger apps, I also did a sanity test and got the preview working in the [compose sample](https://github.com/bazelbuild/rules_kotlin/tree/master/examples/jetpack_compose) in rules kotlin.

Known gaps:
- Missing tests
- No support for versions older than 23.2

Possible future improvements: 
- More optimizations  `BinaryTargetClassFileFinder`, perhaps storing a map of package name -> target keys during sync could help
- Fix additional functionalities, including preview run configurations

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -840,6 +840,7 @@ def _collect_android_ide_info(target, ctx, semantics, ide_info, ide_info_file, o
 
     resources = []
     res_folders = []
+    assets_folders = []
     resolve_files = jars_from_output(output_jar)
     if hasattr(ctx.rule.attr, "resource_files"):
         for artifact_path_fragments, res_files in get_res_artifacts(ctx.rule.attr.resource_files).items():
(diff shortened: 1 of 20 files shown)


idanakav:
Also related: #6086

sgowroji:
@idanakav Could you please take a look at the failing checks?

nkoroste:
any updates on this?

PR #6328, opened 2024-03-28
fix: Replace zip with default Bazel zipper

blorente:
# Checklist

details.

# Description of this change

`zip` might not be available in all machines. Specifically, it might not be available in Remote Build Execution environments. Instead, we use Bazel's built-in zipper.


blorente:
This needs more investigation, seems like the bazel zipper doesn't create files just like the rest of the java wants them to.

Issue #6370, opened 2024-04-07
Add Line Numbers Project View Sections


ahmedyarub:
### Description of the feature request:

Add line numbers to the text blocks and sections so that we can identify them from a PsiElement. I have already created a PR with the proposed changes and it is obvious that not in all cases we know the line number: when we are generating a new file, the line numbers are only known when the file is actually generated. In these cases I set the line number to -1 knowing that before the object is used it will go through the print() command. There, I would update the line number of the element.
If at the moment of the change (ex in [my target management feature](https://github.com/bazelbuild/intellij/pull/6364) I already know the line number of the directory entry to change), then there is no need to assign a dummy value first.

### What underlying problem are you trying to solve with this feature?

There is no way now to link a project view section such as `ListSection` to a position in the physical file. We can neither link a PsiElement to the sections. In other words: there we cannot change the file contents given a section (without regenerating the whole file), nor change a section in response to changes to the file (again, without parsing the whole file).
This already seems to be causing problems. If you look at `TextBlock` removal methods for example: which Python entry is being removed [here](https://github.com/bazelbuild/intellij/blob/master/python/src/com/google/idea/blaze/python/sync/BlazePythonSyncPlugin.java#L510) exactly? And what happens if there is more than one such entry?
These questions have been [raised here](https://github.com/bazelbuild/intellij/pull/6364#issuecomment-2039713120) and that's why I'm proposing this solution.

### What operating system, Intellij IDE and programming languages are you using? Please provide specific versions.

Windows 11, IntelliJ 2024.1, Go

### Have you found anything relevant by searching the web?

No

PR #6364, opened 2024-04-04
#6265 Target Management

ahmedyarub:
Add "Add Directory" button to the project view
Add directory toggle checkboxes to the project view

# Checklist

- [X] I have filed an issue about this change and discussed potential changes with the maintainers.
- [X] I have received the approval from the maintainers to make this change.
- [X] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: `6265`

# Description of this change
Untitled Project

Add directory and management buttons to the project-view file. Note that this is the first part of a series of features and enhancements that target the hardships of navigating large repos using Bazel.

PR #6371, opened 2024-04-07
#6370 Add line numbers to sections and text-blocks

ahmedyarub:
# Checklist

details.

Issue number: `6370`

# Description of this change
Add line numbers to the text blocks and sections so that we can identify them from a PsiElement. It is obvious that not in all cases we know the line number: when we are generating a new file, the line numbers are only known when the file is actually generated. In these cases I set the line number to -1 knowing that before the object is used it will go through the print() command. There, I would update the line number of the element.
If at the moment of the change (ex in [my target management feature](https://github.com/bazelbuild/intellij/pull/6364) I already know the line number of the directory entry to change), then there is no need to assign a dummy value first.

### DIFF
--- a/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
+++ b/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
@@ -16,7 +16,6 @@
 package com.google.idea.blaze.android.projectview;
 
 import static com.google.idea.blaze.base.projectview.parser.ProjectViewParser.TEMPORARY_LINE_NUMBER;
-import static com.google.idea.blaze.base.projectview.parser.ProjectViewParser.TEMPORARY_LINE_NUMBER;
 import static java.util.stream.Collectors.toList;
 
 import com.google.common.collect.ImmutableList;
@@ -83,11 +82,11 @@ public class AndroidSdkPlatformSection {
 
       if (sdks.isEmpty()) {
         builder
-            .add(TextBlockSection.of(TextBlock.of(-1, "# Please set to an android SDK platform")))
+            .add(TextBlockSection.of(TextBlock.of(TEMPORARY_LINE_NUMBER, "# Please set to an android SDK platform")))
             .add(
                 TextBlockSection.of(
                     TextBlock.of(
-                            -1,
+                            TEMPORARY_LINE_NUMBER,
                         "# You currently have no SDKs. Please use the SDK manager first.")))
             .add(ScalarSection.builder(KEY).set("(android sdk goes here)"));
       } else if (sdks.size() == 1) {
@@ -103,7 +102,7 @@ public class AndroidSdkPlatformSection {
                 .stream()
                 .map(androidSdk -> "# android_sdk_platform: " + androidSdk)
                 .collect(toList());
-        builder.add(TextBlockSection.of(new TextBlock(ImmutableList.copyOf(sdkOptions), -1)));
+        builder.add(TextBlockSection.of(new TextBlock(ImmutableList.copyOf(sdkOptions), TEMPORARY_LINE_NUMBER)));
       }
       return builder.build();
     }
--- a/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
+++ b/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
@@ -78,7 +78,7 @@ public class AndroidSdkPlatformSection {
       }
       List<Sdk> sdks = BlazeSdkProvider.getInstance().getAllAndroidSdks();
       ProjectView.Builder builder =
-          ProjectView.builder(topLevelProjectView).add(TextBlockSection.of(TextBlock.newLine(-1)));
+          ProjectView.builder(topLevelProjectView).add(TextBlockSection.of(TextBlock.newLine(TEMPORARY_LINE_NUMBER)));
 
       if (sdks.isEmpty()) {
         builder
--- a/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
+++ b/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
@@ -101,7 +101,7 @@ public class AndroidSdkPlatformSection {
                 .stream()
                 .map(androidSdk -> "# android_sdk_platform: " + androidSdk)
                 .collect(toList());
-        builder.add(TextBlockSection.of(new TextBlock(-1, ImmutableList.copyOf(sdkOptions))));
+        builder.add(TextBlockSection.of(new TextBlock(ImmutableList.copyOf(sdkOptions), -1)));
       }
       return builder.build();
     }
--- a/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
+++ b/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
@@ -77,14 +77,15 @@ public class AndroidSdkPlatformSection {
       }
       List<Sdk> sdks = BlazeSdkProvider.getInstance().getAllAndroidSdks();
       ProjectView.Builder builder =
-          ProjectView.builder(topLevelProjectView).add(TextBlockSection.of(TextBlock.newLine()));
+          ProjectView.builder(topLevelProjectView).add(TextBlockSection.of(TextBlock.newLine(-1)));
 
       if (sdks.isEmpty()) {
         builder
-            .add(TextBlockSection.of(TextBlock.of("# Please set to an android SDK platform")))
+            .add(TextBlockSection.of(TextBlock.of(-1, "# Please set to an android SDK platform")))
             .add(
                 TextBlockSection.of(
                     TextBlock.of(
+                            -1,
                         "# You currently have no SDKs. Please use the SDK manager first.")))
             .add(ScalarSection.builder(KEY).set("(android sdk goes here)"));
       } else if (sdks.size() == 1) {
@@ -94,13 +95,13 @@ public class AndroidSdkPlatformSection {
       } else {
         builder.add(
             TextBlockSection.of(
-                TextBlock.of("# Please uncomment an android-SDK platform. Available SDKs are:")));
+                TextBlock.of(-1, "# Please uncomment an android-SDK platform. Available SDKs are:")));
         List<String> sdkOptions =
             AndroidSdkFromProjectView.getAvailableSdkTargetHashes(sdks)
                 .stream()
                 .map(androidSdk -> "# android_sdk_platform: " + androidSdk)
                 .collect(toList());
-        builder.add(TextBlockSection.of(new TextBlock(ImmutableList.copyOf(sdkOptions))));
+        builder.add(TextBlockSection.of(new TextBlock(-1, ImmutableList.copyOf(sdkOptions))));
       }
       return builder.build();
     }
--- a/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
+++ b/aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
@@ -15,6 +15,8 @@
  */
 package com.google.idea.blaze.android.projectview;
 
+import static com.google.idea.blaze.base.projectview.parser.ProjectViewParser.TEMPORARY_LINE_NUMBER;
+import static com.google.idea.blaze.base.projectview.parser.ProjectViewParser.TEMPORARY_LINE_NUMBER;
 import static java.util.stream.Collectors.toList;
 
 import com.google.common.collect.ImmutableList;
@@ -95,7 +97,7 @@ public class AndroidSdkPlatformSection {
       } else {
         builder.add(
             TextBlockSection.of(
-                TextBlock.of(-1, "# Please uncomment an android-SDK platform. Available SDKs are:")));
+                TextBlock.of(TEMPORARY_LINE_NUMBER, "# Please uncomment an android-SDK platform. Available SDKs are:")));
         List<String> sdkOptions =
             AndroidSdkFromProjectView.getAvailableSdkTargetHashes(sdks)
                 .stream()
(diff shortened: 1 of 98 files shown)


base/src/com/google/idea/blaze/base/projectview/section/ListSection.java
```diff
@@ -88,16 +90,29 @@ public static <T> Builder<T> update(
...
 
+    public Builder<T> setFirstLineNumber(int firstLineNumber) {
+      this.firstLineNumber = firstLineNumber;
+      return this;
+    }
+
     @CanIgnoreReturnValue
     public final Builder<T> add(T item) {
-      items.add(new ItemOrTextBlock<>(item));
+      items.add(new ItemOrTextBlock<>(item, -1));
```
blorente, line 109: When do we want to call this method? It feels like the only valid states an item should have are "with a valid line number" and _maybe_ "with the previous line number + 1", I'm not sure when this is a valid state.

base/src/com/google/idea/blaze/base/projectview/section/sections/TextBlock.java
```diff
@@ -28,30 +30,37 @@
...
-  public static TextBlock of(String... lines) {
-    return new TextBlock(ImmutableList.<String>builder().add(lines).build());
+  public static TextBlock of(int firstLineIndex, String... lines) {
+    return new TextBlock(ImmutableList.<String>builder().add(lines).build(), firstLineIndex);
   }
 
   /** A text block that is a single newline */
-  public static TextBlock newLine() {
-    return new TextBlock(ImmutableList.of(""));
+  public static TextBlock newLine(int firstLineIndex) {
```
blorente, line 50: If this is always going to be -1 at creation time (I don't see a line in this PR where we pass a different value), could we just not pass it and assign -1 to it?

base/src/com/google/idea/blaze/base/projectview/ProjectView.java
```diff
@@ -129,6 +130,10 @@ public <T, SectionType extends Section<T>> SectionType getLast(SectionKey<T, Sec
       return null;
     }
 
+    public <T, SectionType extends Section<T>> Builder add(SectionBuilder<T, SectionType> builder, int firstLineIndex) {
+      return null;
```
blorente, line 134: Is returning `null` correct here? I feel like we should at least call some other implementation of `add` and return that, because that's what all the other `add` methods do.

I also don't see this method called anywhere, so maybe we can remove it.

base/src/com/google/idea/blaze/base/projectview/ProjectView.java
```diff
@@ -25,6 +25,7 @@
 import com.google.idea.blaze.base.projectview.section.Section;
 import com.google.idea.blaze.base.projectview.section.SectionBuilder;
 import com.google.idea.blaze.base.projectview.section.SectionKey;
+
```
blorente, line 28:
```suggestion
```
Leftover space.

blorente (commented):
If we're using -1 to mark "uninitialized", could we make it a constant somewhere? I have a short memory, I'm sure I'll forget what -1 means 6months from now when I come back to this code.

blorente:
Additionally, I'm a bit scared of printing having side effects on the data structure.
Could we instead traverse the data structure in [`ProjectView.Builder.build()`](https://github.com/bazelbuild/intellij/blob/9d5dbe240b1055cb6ca0d1720a65398649f5b760/base/src/com/google/idea/blaze/base/projectview/ProjectView.java#L178) and assign line numbers then? That way we make sure that:
- Printing doesn't modify the structure.
- We always have line numbers when we build a project view, whether we print it or not.

blorente:
And, because I forgot to start with this: Thank you for your contributions! Most maintainers here have limited time to implement cool features in the plugin, and it's delightful to see the plugin improving.

base/src/com/google/idea/blaze/base/projectview/section/ListSection.java
ahmedyarub: This one can only be used when you are creating a project view file from scratch. At that time, you don't know anything about line numbers. So where do we get the line numbers from? These add commands always have a print after them. If you look at the print commands you will see that they alter the line numbers. So this is more than a temporary state that is complemented by actually writing to the file.

base/src/com/google/idea/blaze/base/projectview/section/sections/TextBlock.java
ahmedyarub: We could, but I can totally imagine a future implementation where we would add a new line at an existing file. In that case, we would know the line number AND we would wnt to specify where to do that exactly. One example would be: a linter/formatter.

base/src/com/google/idea/blaze/base/projectview/ProjectView.java
ahmedyarub: Method removed.

ahmedyarub:
> Additionally, I'm a bit scared of printing having side effects on the data structure. Could we instead traverse the data structure in [`ProjectView.Builder.build()`](https://github.com/bazelbuild/intellij/blob/9d5dbe240b1055cb6ca0d1720a65398649f5b760/base/src/com/google/idea/blaze/base/projectview/ProjectView.java#L178) and assign line numbers then? That way we make sure that:
> 
> * Printing doesn't modify the structure.
> * We always have line numbers when we build a project view, whether we print it or not.

What you are mentioning is the ideal solution beyond any doubt. The problem is that each print statement decides on its own how many new lines to add. Moving the logic to the builder means consolidating all that "new-line logic" into the ProjectView class. Can we guarantee that future printers would be compatible with ProjectView? I mean we can at least add a comment to the `print()` method in the interface saying that line numbers should be changed when printing which would be easily visible for devs.

ahmedyarub:
> And, because I forgot to start with this: Thank you for your contributions! Most maintainers here have limited time to implement cool features in the plugin, and it's delightful to see the plugin improving.

I really appreciate your support and patience you and especially @tpasternak . I assure you that there will be much bigger changes to come.

ahmedyarub:
> > What you are mentioning is the ideal solution beyond any doubt. The problem is that each print statement decides on its own how many new lines to add. Moving the logic to the builder means consolidating all that "new-line logic" into the ProjectView class. Can we guarantee that future printers would be compatible with ProjectView? I mean we can at least add a comment to the print() method in the interface saying that line numbers should be changed when printing which would be easily visible for devs.
> 
> Hmm, that's a good point.
> 
> I'm looking at the code more closely now, and it seems that `print` is only ever called when we call `projectViewToString`. Instead of updating the line numbers when we print, could re-parse the string when we generate it, to update the structure? This is more inefficient, but honestly I don't see project view files going over a handful of thousands of lines anyway, and there are optimizations we can do.
> 
> Alternatively, we could have each `*Section` keep an in-memory, up-to-date string representation of itself, so that print is just returning that representation, and move all the update calculation to the `parse()` methods.
> 
> This _seems_ more incompatible for future printers, but I don't think it's the case:
> 
> If we're saying that "future printers might want to have different line numbers for the same section", the current code in this PR (and the current printing solution as a whole) is already incompatible with it. If we wanted this feature, we would have to build something else on top of it. I'm not opposed to it, but it seems like a separate change.
> 
> What do you think?

I will certainly A/B test implementing both of these suggestions and come back to you :)

ahmedyarub:
OK so I removed line number calculation from all the print() methods and then tested the plugin behavior thoroughly. I found that all the code-paths call the parseProjectView() after writing to the file multiple times so there is no need to calculate the line numbers when the file is being created or edited.

blorente (commented):
Thanks for making the changes!

Given the flow you've proposed, I'm not sure we need `TEMPORARY_LINE_NUMBER` anymore. In fact, I'm not sure we need to pass line numbers to the individual methods of the `Builder`s of different sections.

If I understand correctly, every creation and update to a `ProjectView` goes through a `Builder`. If we pass the initial line index to a `Builder`, it should have enough information to assign line numbers based on the operation and the sections it contains.

So, instead of having `Builder.add(T element, int lineIdx)`, we'd have `Builder.add(T element)` and the `Builder` would take care of calculating which line numbers `T` should occupy.

I think this encapsulates the behaviour of adding line numbers neatly into the Builders, so that the calling code doesn't have to keep that state.

What do you think?

ahmedyarub:
> public Builder(SectionKey<T, ListSection<T>> sectionKey, @Nullable ListSection<T> section) {

I tried doing that but found multiple instances where the builder is instantiated and used without having known line numbers. The line numbers are mostly defined when the print() method is called, and that's why we are depending on reparsing to replace the temporary line numbers with real one. I can't think of a way for example to make that change to AddDirectoryToProjectAction.doOKAction() and GenerateFromBuildFileSelectProjectViewOption.guessProjectViewFromLocation(). How do you suggest doing these changes in these methods?

base/src/com/google/idea/blaze/base/projectview/section/ListSection.java
```diff
@@ -88,16 +92,29 @@ public static <T> Builder<T> update(
   public static class Builder<T> extends SectionBuilder<T, ListSection<T>> {
     private final List<ItemOrTextBlock<T>> items = new ArrayList<>();
 
+    private int firstLineNumber = TEMPORARY_LINE_NUMBER;
+
     public Builder(SectionKey<T, ListSection<T>> sectionKey, @Nullable ListSection<T> section) {
```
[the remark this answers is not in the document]
ahmedyarub: I tried doing that but found multiple instances where the builder is instantiated and used without having known line numbers. The line numbers are mostly defined when the print() method is called, and that's why we are depending on reparsing to replace the temporary line numbers with real one. I can't think of a way for example to make that change to AddDirectoryToProjectAction.doOKAction() and GenerateFromBuildFileSelectProjectViewOption.guessProjectViewFromLocation(). How do you suggest doing these changes in these methods?

ahmedyarub:
> Thanks for making the changes!
> 
> Given the flow you've proposed, I'm not sure we need `TEMPORARY_LINE_NUMBER` anymore. In fact, I'm not sure we need to pass line numbers to the individual methods of the `Builder`s of different sections.
> 
> If I understand correctly, every creation and update to a `ProjectView` goes through a `Builder`. If we pass the initial line index to a `Builder`, it should have enough information to assign line numbers based on the operation and the sections it contains.
> 
> So, instead of having `Builder.add(T element, int lineIdx)`, we'd have `Builder.add(T element)` and the `Builder` would take care of calculating which line numbers `T` should occupy.
> 
> I think this encapsulates the behaviour of adding line numbers neatly into the Builders, so that the calling code doesn't have to keep that state.
> 
> What do you think?

I tried doing that but found multiple instances where the builder is instantiated and used without having known line numbers. The line numbers are mostly defined when the print() method is called, and that's why we are depending on reparsing to replace the temporary line numbers with real one. I can't think of a way for example to make that change to AddDirectoryToProjectAction.doOKAction() and GenerateFromBuildFileSelectProjectViewOption.guessProjectViewFromLocation(). How do you suggest doing these changes in these methods?

ahmedyarub:
@tpasternak @blorente is there anything else that I need to do here?

aswb/src/com/google/idea/blaze/android/projectview/AndroidSdkPlatformSection.java
```diff
@@ -77,14 +78,15 @@ public ProjectView addProjectViewDefaultValue(
       }
       List<Sdk> sdks = BlazeSdkProvider.getInstance().getAllAndroidSdks();
       ProjectView.Builder builder =
-          ProjectView.builder(topLevelProjectView).add(TextBlockSection.of(TextBlock.newLine()));
+          ProjectView.builder(topLevelProjectView).add(TextBlockSection.of(TextBlock.newLine(-1)));
```
blorente, line 81:
```suggestion
          ProjectView.builder(topLevelProjectView).add(TextBlockSection.of(TextBlock.newLine(TEMPORARY_LINE_NUMBER)));
```

base/tests/unittests/com/google/idea/blaze/base/projectview/parser/ProjectViewParserTest.java
```diff
@@ -451,23 +452,24 @@ public void testListParserAcceptsWhitespace() throws Exception {
...
+                            .add(TextBlock.of(2, "  "))
+                            .add(TextBlock.of(3, ""))
+                            .add(DirectoryEntry.include(new WorkspacePath("dir1")), 4)
+                            .add(TextBlock.of(5, "  ", "  "))
+                            .add(TextBlock.of(7, "# comment"))
+                            .add(DirectoryEntry.include(new WorkspacePath("dir2")), 8)
+                            .add(TextBlock.of(9, ""))
+                            .add(TextBlock.of(10, "  # commented out dir"))
+                            .add(TextBlock.of(11, "  ")))
+            .add(TextBlockSection.of(TextBlock.of(12, "# comment", "# comment")))
```
blorente, line 469: This test doesn't contain any list sections (I think). could you add a test that tests that lines are parsed correctly in all the sections you modified?

blorente (approved):
I finally got a little bit of time to re-review this.
It seems we can't get more invariants from the codebase as it is, so I think it's fine to merge this and iterate later on how we can remove `TEMPORARY_LINE_NUMBER`.

Approving with the caveat that I'd love to see more tests, I'll be happy to merge once the tests are there.

base/tests/unittests/com/google/idea/blaze/base/projectview/parser/ProjectViewParserTest.java
ahmedyarub: It does cover the `ListSection` since the declarations are `Builder<T> setFirstLineNumber(int firstLineNumber)`, `public static <T> Builder<T> builder(SectionKey<T, ListSection<T>> sectionKey)`, and `public static final SectionKey<DirectoryEntry, ListSection<DirectoryEntry>> KEY =  SectionKey.of("directories");` which ultimately calls `ListSection` and checks it. In fact, if you build my Target Manager diff which already has all the code here you will see that even if you add duplicate directories and sections the correct one would be toggled when you click on the checkbox.

ahmedyarub:
@blorente is there any other requirement for merging?

blorente:
Not as far as I can tell, thanks for the contribution, and apologies for the slow review turnarounds!

PR #6388, opened 2024-04-11
feat: create "Add package" EditorNotificationPanel if derive_targets_from_directories is on

tpasternak:
Based on #6359 by @dkashyn-sfdc
# Checklist

details.

# Description of this change

### DIFF
--- a/base/src/com/google/idea/blaze/base/dependencies/AddSourceToProjectHelper.java
+++ b/base/src/com/google/idea/blaze/base/dependencies/AddSourceToProjectHelper.java
@@ -44,6 +44,7 @@ import com.google.idea.blaze.base.settings.Blaze;
 import com.google.idea.blaze.base.settings.BlazeImportSettings.ProjectType;
 import com.google.idea.blaze.base.settings.ui.OpenProjectViewAction;
 import com.google.idea.blaze.base.sync.BlazeSyncManager;
+import com.google.idea.blaze.base.sync.BuildTargetFinder;
 import com.google.idea.blaze.base.sync.data.BlazeProjectDataManager;
 import com.google.idea.blaze.base.sync.projectview.ImportRoots;
 import com.google.idea.blaze.base.sync.workspace.WorkspacePathResolver;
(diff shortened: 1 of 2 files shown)


tpasternak:
The code contains some code duplications, but otherwise it would require either to put some `if` statements in the middle of the bodies, or to rework the whole file, which we would like to avoid @ujohnny @LeFrosch

base/src/com/google/idea/blaze/base/dependencies/AddSourceToProjectHelper.java
```diff
@@ -167,6 +199,74 @@ static void addSourceAndTargetsToProject(
...
+    if (edit == null) {
+      Messages.showErrorDialog(
+          "Could not modify project view. Check for errors in your project view and try again",
+          "Error");
+      return;
+    }
+    edit.apply();
+    // TODO(brendandouglas): support partially syncing a directory with the same query-based
+    // filtering
+    List<? extends TargetExpression> targetsToSync = targets;
```
ujohnny, line 240: This is always null

base/src/com/google/idea/blaze/base/dependencies/ExternalFileProjectManagementHelper.java
```diff
@@ -167,42 +168,56 @@ public EditorNotificationPanel createNotificationPanelForLegacySync(VirtualFile
...
+      boolean addTargets = !alreadyBuilt && !AddSourceToProjectHelper.autoDeriveTargets(project);
+      ListenableFuture<List<TargetInfo>> targetsFuture =
+              addTargets
+                      ? AddSourceToProjectHelper.getTargetsBuildingSource(context)
+                      : Futures.immediateFuture(ImmutableList.of());
+      if (targetsFuture == null) {
+        return null;
+      }
+      EditorNotificationPanel panel =
+              createPanel(
```
ujohnny, line 181: It seems to me that some regions have formatting issues. I believe the indent should be 4 spaces instead of 8.

ujohnny:
So initially there was a notification 

```
Do you want to add this file to your project sources?           Add file to project
```

And with this change it in some cases looks like

```
Do you want to add this file to your project sources?           Add package to project
```

I'd consider updating notification text for the new action into something like 
```
Do you want to add this package to your project sources?  Add package to project
```

Also I checked the code on a basic example with `directories: .` and click on the notification does not do anything. I guess this has to be checked before displaying the notification

[Screencast from 2024-04-18 16-33-45.webm](https://github.com/bazelbuild/intellij/assets/759709/a62b3d92-59e2-44bb-96e0-e7e799f72c3f)

ujohnny (changes requested):

ujohnny (approved):
lgtm, but agreed to extract duplicated constant to variable in a separate PR

PR #6393, opened 2024-04-18
Fix debugger attachment to running android process

idanakav:
The extension was never registered in the plugin xml and causing a silent crash when trying to attach a debugger to en existing android process


ThomasCJY:
@idanakav @mai93 @vsiva Android Debugger Attachment on bazel project has been broken since Iguana and this seems to be the correct fix. Can we prioritize merging the change and include it in the next release?

The `intellij-plugin-aspect-google`  failure seems to be flakiness, we probably can rebase the branch to trigger a rebuild

ThomasCJY:
Looks like [this cherry-pick](https://github.com/bazelbuild/intellij/pull/6803) merged nothing into google branch. Can we reopen this issue and merge the fix correctly? @mai93

ThomasCJY:
> The google branch is now deprecated and the Android Studio plugin is released from [AOSP](https://android.googlesource.com/platform/tools/adt/idea/+/refs/heads/mirror-goog-studio-main/aswb). I can initiate a change to fix this there.

SG. Thanks for the quick turnaround

mai93:
It seems that it is already there: https://android.googlesource.com/platform/tools/adt/idea/+/refs/heads/mirror-goog-studio-main/aswb/aswb/src/META-INF/aswb.xml#154

PR #6404, opened 2024-04-25
fix: disable bzlmod for"with_go_source" example

tpasternak:
# Checklist

details.

# Description of this change

### DIFF
--- a/examples/go/with_go_source/with_go_source.bazelproject
+++ b/examples/go/with_go_source/with_go_source.bazelproject
@@ -6,4 +6,7 @@ derive_targets_from_directories: true
 additional_languages:
   go
 
+build_flags:
+  --enable_bzlmod=false
+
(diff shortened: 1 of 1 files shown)


mai93:
is the plugin failing to support bzlmod for this project?

blorente:
I think this is not a plugin issue, but rather a project issue. As it stands the project is not bzlmod-compatible (running `bazel run //:gazelle` fails). Maybe we can just disable bzlmod in the `.bazelrc` instead of the project file and make it compatible later

blorente (approved):
Approving to unblock merging, but we definitely should either make the project bzlmod-compatible or move the flag to `.bazelrc`.

tpasternak:
But it's only because we have `--enable_bzlmod` in the root .bazelrc

tpasternak:
oh, wait, I'm wrong

PR #6420, opened 2024-05-06
Update CHANGELOG for v2024.05.07

satyanandak:


mai93:
- Plugin aspect now depends on `@rules_java`. If this breaks you, please update your `@rules_java` version to 5.3.5 or later.
- Internal cleanup

PR #6422, opened 2024-05-07
Add 'View project root' flag

tpasternak:
# Checklist

details.

# Description of this change

https://github.com/bazelbuild/intellij/assets/1066003/29720ae3-bc31-4b05-b9d6-da26d2e9b760

### DIFF
--- a/base/src/com/google/idea/blaze/base/projectview/section/sections/ViewProjectRootSection.java
+++ b/base/src/com/google/idea/blaze/base/projectview/section/sections/ViewProjectRootSection.java
@@ -19,7 +19,6 @@ import com.google.idea.blaze.base.projectview.section.ScalarSection;
 import com.google.idea.blaze.base.projectview.section.SectionKey;
 import com.google.idea.blaze.base.projectview.section.SectionParser;
 
-/** If set to true, automatically derives targets from the project directories. */
 public class ViewProjectRootSection {
   public static final SectionKey<Boolean, ScalarSection<Boolean>> KEY =
       SectionKey.of("view_project_root");
--- a/base/src/com/google/idea/blaze/base/projectview/section/sections/ViewProjectRootSection.java
+++ b/base/src/com/google/idea/blaze/base/projectview/section/sections/ViewProjectRootSection.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2024 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


base/src/com/google/idea/blaze/base/projectview/section/sections/ViewProjectRootSection.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.idea.blaze.base.projectview.section.sections;
+
+import com.google.idea.blaze.base.projectview.section.ScalarSection;
+import com.google.idea.blaze.base.projectview.section.SectionKey;
+import com.google.idea.blaze.base.projectview.section.SectionParser;
+
+/** If set to true, automatically derives targets from the project directories. */
```
blorente, line 22: I think this comment is out of date.

base/tests/unittests/com/google/idea/blaze/base/projectview/ProjectViewSetTest.java
```diff
@@ -120,6 +121,7 @@ public void testProjectViewSetSerializable() {
                         ScalarSection.builder(BuildConfigSection.KEY)
                             .set(new WorkspacePath("test")))
                     .add(ScalarSection.builder(UseExclusionPatternsSection.KEY).set(false))
+                    .add(ScalarSection.builder(ViewProjectRootSection.KEY).set(false))
```
blorente, line 124: Could you add a test where we actually set it in a project view? We don't have to assert anything on the actual behaviour, I just want to check that it parses.

blorente (approved):
Approved with very minor comments! Thanks a lot for this change.

tpasternak:
Oh, I can't just test it properly with the current test framework, because it converts `TestFileSystem`'s files to NIO files via paths without hestitation, so as a result the tests would crash because it's impossible to call `listFiles` on it. I removed the comment (as it's already commented in string) and I'll try to think about fixing the test framework later

https://github.com/bazelbuild/intellij/blob/master/base/tests/utils/integration/com/google/idea/blaze/base/BlazeIntegrationTestCase.java#L110

blorente:
That's fair, it's fine to leave the proper tests as a follow up. Do you think it would be hard to add a test that asserts that the section parses correctly? So a test that takes the string:
```
view_project_root: true
```
And asserts that the right Section is created.

tpasternak:
Ok, that's easy, could you take a look again?

blorente:
Yeah, looks good to me! Thanks for this change.

PR #6460, opened 2024-06-04
fix(output): Ignore sync errors in excluded directories

blorente:
# Checklist

details.

Issue number: `N/A`

# Description of this change
Even if a directory is excluded, Bazel can try to read it and issue
errors during sync. We should ignore Bazel errors that come from
excluded packages.

Implementation details:
- We create a new issue category, `IGNORE`.
- We modify the relevant `BlazeIssueParsers` to hold references to `workspaceRoot` and the `projectViewSet`
- When parsing an issue, if the issue happens in an excluded package or directory or label, we emit it as `IGNORE`.

### DIFF
--- a/base/src/com/google/idea/blaze/base/issueparser/BlazeIssueParser.java
+++ b/base/src/com/google/idea/blaze/base/issueparser/BlazeIssueParser.java
@@ -35,6 +35,8 @@ import com.google.idea.blaze.base.projectview.section.SectionKey;
 import com.google.idea.blaze.base.projectview.section.sections.TargetSection;
 import com.google.idea.blaze.base.run.filter.FileResolver;
 import com.google.idea.blaze.base.scope.output.IssueOutput;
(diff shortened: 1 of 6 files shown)


agluszak (commented):
Code-wise LGTM, but feature-wise I don't feel confident enough to approve it

blorente:
@agluszak fair enough, thanks for taking a look! It is a potentially confusing change to the UX, I appreciate the thoughtfulness. 

I'll ping @tpasternak and @mai93 to see if they want the feature, otherwise I'm happy to close the PR.

blorente:
Synced with @tpasternak online. We'll hold off on landing this because it might hide legitimate errors in third party dependencies.

Our use-case for this is pretty niche (we use it for projects with both Bazel and CMake builds in the same repo), but the right solution is probably to expose another extension point for issue parsers. I'm not going to do that in this PR, so closing this one.

PR #6462, opened 2024-06-04
fix: Rely on the bazel zipper instead of zip

blorente:
# Checklist

details.

N/A

# Description of this change

Not all environments have `zip` in their path. For instance, some Remote Execution environments like to have very lean workers. Furthermore, we shouldn't rely on all environments having the same version. We replace zip with the bazel jar handler.

We have been using this change internally to build the plugin against our build environment for a few months now.

### DIFF
--- a/intellij_platform_sdk/build_defs.bzl
+++ b/intellij_platform_sdk/build_defs.bzl
@@ -618,8 +618,26 @@ def no_mockito_extensions(name, jars, **kwargs):
             cmd = """
             cp "$<" "$@"
             chmod u+w "$@"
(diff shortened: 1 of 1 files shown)


agluszak (approved):
It's a pity that the bazel zipper requires so many additional steps, but I think it's worth it.

blorente:
Yeah, it wasn't my happiest PR to write 😅 . There are java tools like jarjar which can do this, but I erred on the side of not introducing another dependency.

tpasternak:
ok, now we have #6468

blorente:
I’ll get on this right now.

PR #6467, opened 2024-06-05
Add extension point to detect whether a module is to be orphaned or retained.

gchandu25:
# Checklist

details.

# Description of this change
Added an EP to give the feasibility to allow other modules. For example, when maven modules are added into project view bazel sync resets the view as they have different content root. An implementation could be added to allow maven modules to be shown in project tree view.


base/src/com/google/idea/blaze/base/sync/projectstructure/OrphanedModuleValidator.java
```diff
@@ -0,0 +1,14 @@
+package com.google.idea.blaze.base.sync.projectstructure;
+
+import com.intellij.openapi.extensions.ExtensionPointName;
+import com.intellij.openapi.module.Module;
+
+/**
+ * Finds if the module is an orphan module, to be not shown in project tree view.
+ */
+public interface OrphanedModuleValidator {
```
dkashyn-sfdc, line 9: I'm not sure about name... it is not `validator` but it is up to @tpasternak to decide what would be a good name here

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleEditorImpl.java
```diff
@@ -39,11 +39,12 @@
 import com.intellij.openapi.roots.impl.ModifiableModelCommitter;
 import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VirtualFile;
+
+import javax.annotation.Nullable;
```
dkashyn-sfdc, line 43: where is this used?

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleEditorImpl.java
```diff
@@ -119,8 +120,8 @@ public void commitWithGc(BlazeContext context) {
     List<Module> orphanModules = Lists.newArrayList();
     for (Module module : ModuleManager.getInstance(project).getModules()) {
       if (!modules.containsKey(module.getName())
-              && OrphanedModuleValidator.EP_NAME.getExtensionList().stream().anyMatch(ep -> ep.isOrphanModule(module))) {
-        orphanModules.add(module);
+              && (ModuleValidator.EP_NAME.getExtensionList().isEmpty() || ModuleValidator.EP_NAME.getExtensionList().stream().anyMatch(n -> !n.isModuleValid(module)))) {
```
dkashyn-sfdc, line 123: This is a wrong condition... if at least ANY "module consumer" is interested it cannot be `orphaned`, with your current logic if any consumer is not interested module will be `orphaned`

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleValidator.java
```diff
@@ -0,0 +1,14 @@
+package com.google.idea.blaze.base.sync.projectstructure;
+
+import com.intellij.openapi.extensions.ExtensionPointName;
+import com.intellij.openapi.module.Module;
+
+/**
+ * Extension point to customize the validity of a given module, used to show the module in project tree view.
+ */
+public interface ModuleValidator {
```
dkashyn-sfdc, line 9: I don't think `validity` is a proper term here...
I'd go with `ModuleManager` and instead of `isModuleValid` something like `isModuleSupported` or `isManaged`

gchandu25: There is already another ModuleManager...
I updated the method to 'hasOtherContentRoot' as thats really what is being checked to tag them as orphan or not.

gchandu25: I'm still open for suggestions on class names..

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleEditorImpl.java
```diff
@@ -120,7 +119,7 @@ public void commitWithGc(BlazeContext context) {
     List<Module> orphanModules = Lists.newArrayList();
     for (Module module : ModuleManager.getInstance(project).getModules()) {
       if (!modules.containsKey(module.getName())
-              && (ModuleValidator.EP_NAME.getExtensionList().isEmpty() || ModuleValidator.EP_NAME.getExtensionList().stream().anyMatch(n -> !n.isModuleValid(module)))) {
+              && (ModuleValidator.EP_NAME.getExtensionList().isEmpty() || ModuleValidator.EP_NAME.getExtensionList().stream().anyMatch(n -> !n.hasOtherContentRoot(module)))) {
```
dkashyn-sfdc, line 122: Instead of 
`ModuleValidator.EP_NAME.getExtensionList().stream().anyMatch(n -> !n.hasOtherContentRoot(module)))`
it should be `!ModuleValidator.EP_NAME.getExtensionList().stream().anyMatch(n -> n.hasOtherContentRoot(module)))` or so.

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleValidator.java
dkashyn-sfdc: You have package/namespace... Yet name is something to select carefully here. EP are generic so let's have something generic yet specific enough.

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleEditorImpl.java
```diff
@@ -118,8 +118,9 @@ public Module findModule(String moduleName) {
   public void commitWithGc(BlazeContext context) {
     List<Module> orphanModules = Lists.newArrayList();
     for (Module module : ModuleManager.getInstance(project).getModules()) {
-      if (!modules.containsKey(module.getName())) {
-        orphanModules.add(module);
+      if (!modules.containsKey(module.getName())
+              && (ModuleValidator.EP_NAME.getExtensionList().isEmpty() || !ModuleValidator.EP_NAME.getExtensionList().stream().anyMatch(n -> n.hasOtherContentRoot(module)))) {
```
tpasternak, line 122: Do we need `isEmpty` if we already have `!...anyMatch`?

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleValidator.java
tpasternak: Yeah, so I'd prefer to be as explicit as possible. So how about sth like `isOwnedByExternalPlugin` or sth like that? I agree that using the term "validity" might be misleading

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleEditorImpl.java
gchandu25: Got reordered , its here https://github.com/bazelbuild/intellij/blob/f17f006ac6c1d94711958038af383ca86a680fba/base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleEditorImpl.java#L114

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleValidator.java
gchandu25: I'm thinking something like this:
class name - `ExternalModuleProvider`, method - `shouldOrphanModule` or `isAnOrphanModule` (using word 'orphan' because orphan modules will not be shown in this context)

base/src/com/google/idea/blaze/base/sync/projectstructure/ModuleEditorImpl.java
gchandu25: huh..may be not. I added explicitly to handle default case, when there are no EP implementations.

base/src/com/google/idea/blaze/base/sync/projectstructure/OrphanedModuleValidator.java
gchandu25: updated to ExternalModuleProvider.

base/src/com/google/idea/blaze/base/sync/projectstructure/ExternalModuleProvider.java
```diff
@@ -0,0 +1,14 @@
...
+import com.intellij.openapi.module.Module;
+
+/**
+ * An EP provider for external modules.
+ */
+public interface ExternalModuleProvider {
+    ExtensionPointName<ExternalModuleProvider> EP_NAME =
+            ExtensionPointName.create("com.google.idea.blaze.base.sync.projectstructure.ExternalModuleProvider");
+
+    boolean isOwnedByExternalModule(Module module);
```
dkashyn-sfdc, line 13: it is not really `externalModule` that own this module but it is much better than `validation` so I'll approve (not that it matters :) )

dkashyn-sfdc (commented):
LGTM

base/src/com/google/idea/blaze/base/sync/projectstructure/ExternalModuleProvider.java
tpasternak: yeah, if it was an internal name I would change it afterwards, but this is an EP, so let's avoid changes once its merged. Could you please change that to `isOwnedByExternalPlugin` or `isOwnedByThirdPartyPlugin`?

tpasternak (dismissed):

base/src/com/google/idea/blaze/base/sync/projectstructure/ExternalModuleProvider.java
gchandu25: updated to 'isOwnedByExternalPlugin'. Please a take @tpasternak.

jastice (approved):

Issue #6474, opened 2024-06-07
Latest bazel plugin breaks on any C++ project that uses old (prior 6.4.0) bazel


erinacio:
### Description of the bug:

Syncing any project with bazel <=6.3.2 will fail with:

```
ERROR: Traceback (most recent call last):
	File "<redacted>/external/intellij_aspect/intellij_info_impl_bundled.bzl", line 3, column 37, in <toplevel>
		load("@rules_java//java:defs.bzl", "JavaInfo", "java_common")
Error: file '@rules_java//java:defs.bzl' does not contain symbol 'JavaInfo'
ERROR: at <redacted>/external/intellij_aspect/intellij_info_bundled.bzl:4:5: initialization of module 'intellij_info_impl_bundled.bzl' failed
```

### Which category does this issue belong to?

CLion

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Just create a minimal Bazel C++ project and sync, using bazel 6.3.2 or earlier version.

```bazel
# WORKSPACE
workspace(name = "bzlexp")
```

```bazel
# BUILD.bazel
cc_binary(
    name = "main",
    srcs = ["main.cc"],
)
```

```cpp
// main.cc
int main() { return 0; }
```

### Which Intellij IDE are you using? Please provide the specific version.

CLion 2024.1.2

### What programming languages and tools are you using? Please provide specific versions.

Not related

### What Bazel plugin version are you using?

2024.05.21.0.1-api-version-241

### Have you found anything relevant by searching the web?

https://github.com/bazelbuild/intellij/issues/6430 is closed because claimed that the bug is fixed, which is not.

Someone noticed about upgrading `rules_java` in that issue, but how can one upgrade `rules_java` that is just not imported?

### Any other information, logs, or outputs that you want to share?

No.

tpasternak:
Thank you for reporting this, we are aware of the problem, work in progress

LeFrosch:
In the meantime you can try adding this to your `WORKSPACE` file as a workaround:
```python
http_archive(
    name = "rules_java",
    urls = [
        "https://github.com/bazelbuild/rules_java/releases/download/7.6.1/rules_java-7.6.1.tar.gz",
    ],
    sha256 = "f8ae9ed3887df02f40de9f4f7ac3873e6dd7a471f9cddf63952538b94b59aeb3",
)
```

tpasternak:
[...]

PR #6476, opened 2024-06-10
Revert "Automatic code cleanup." - Remove dependency on rules_java

LeFrosch:
This reverts commit 697b9a9a9fe4d0b3c0672a694d8bbd3a5e5a5d4e.

# Checklist

details.

Issue number: #6474

# Description of this change
It seems to be the case that if a project:
- depends directly or transitively on rules_java < 5.3.5 
- or uses bazel <= 6.3.2 

The project can no longer be synced because of an error like this:
```
	File "<redacted>/external/intellij_aspect/intellij_info_impl_bundled.bzl", line 3, column 37, in <toplevel>
		load("@rules_java//java:defs.bzl", "JavaInfo", "java_common")
Error: file '@rules_java//java:defs.bzl' does not contain symbol 'JavaInfo'
```

I would suggest reverting this commit for now and to release a hotfix.

### DIFF
--- a/aspect/build_dependencies.bzl
+++ b/aspect/build_dependencies.bzl
@@ -5,7 +5,6 @@ load(
     "CPP_COMPILE_ACTION_NAME",
     "C_COMPILE_ACTION_NAME",
 )
-load("@rules_java//java:defs.bzl", "JavaInfo")
 
 ALWAYS_BUILD_RULES = "java_proto_library,java_lite_proto_library,java_mutable_proto_library,kt_proto_library_helper,_java_grpc_library,_java_lite_grpc_library,kt_grpc_library_helper,java_stubby_library,kt_stubby_library_helper,aar_import,java_import"
 
--- a/aspect/fast_build_info.bzl
+++ b/aspect/fast_build_info.bzl
@@ -1,6 +1,5 @@
 """An aspect to gather info needed by the FastBuildService."""
 
-load("@rules_java//java:defs.bzl", "JavaInfo", "java_common")
 load(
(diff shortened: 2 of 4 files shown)


mai93:
We added a note about this breaking change to the release notes https://github.com/bazelbuild/intellij/releases/tag/v2024.05.21-stable, do you think it is not enough?

ujohnny:
> We added a note about this breaking change to the release notes https://github.com/bazelbuild/intellij/releases/tag/v2024.05.21-stable, do you think it is not enough?

We've received several reports also to JetBrains YouTrack about this issue, I assume we have to take some action based on this feedback. It's partially confusing to require `rules_java` in C++ projects, for example old `rules_java` version is coming from `googletest` dependencies IIUC.

mai93:
@hvadehra will this load be mandatory in Bazel 8?

is there a way we can refactor the aspect code to not have this load for C++ workspaces? Until we figure this out we can rollback this commit

hvadehra:
> @hvadehra will this load be mandatory in Bazel 8?

The plan currently is Bazel 8 will not contain these native rules/symbols, so yes, the loads will become necessary then.

Rollback SGTM for now.

> is there a way we can refactor the aspect code to not have this load

`load()` statements unfortunately can't be conditional. If we have different plugin releases by Bazel versions, we can to handle this in the packaging/release step for the plugin. Otherwise, ISTM the plugin may need to use different root bzl files (i.e. one with loads and one without) for the aspect(s) that will be conditioned on the Bazel version.

mai93 (approved):

PR #6463, opened 2024-06-04
fix(coverage): Account for SHA in lcov DA parsing

blorente:
# Checklist

- [ ] I have filed an issue about this change and discussed potential changes with the maintainers.
- [X] I have received the approval from the maintainers to make this change.
- [X] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

N/A

# Description of this change

Some coverage formats include a SHA in their report. In those cases, the parser failed and no coverage was reported. This PR fixes that.

PR #6419, opened 2024-05-06
Fix for wrong compiler executable when using custom toolchains

LeFrosch:
# Checklist

- [x] I have filed an issue about this change and discussed potential changes with the maintainers.
- [x] I have received the approval from the maintainers to make this change.
- [x] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: #6418

# Description of this change
Fixes #6418 by using `cc_common.get_tool_for_action` to find the right c and cpp compiler executable.

PR #6484, opened 2024-06-17
Update changelog for v2024.06.18 release

satyanandak:


tpasternak:
btw how about passing these entries through chat gpt? it would look like that

### IntelliJ Bazel Plugin Release Notes

#### Fixes
- **View Project Root Issue**: Resolved missing directories when `view_project_root` is enabled. ([#6493](#6493))
- **Read Access Exception**: Fixed "Read access is allowed from inside read-action" exception occurring when opening a file. ([#6489](#6489))
- **Compiler Executable**: Corrected the issue with the wrong compiler executable when using custom toolchains. ([#6419](#6419))
- **Coverage Reports**: Improved coverage report parsing by accounting for SHA in `lcov` DA. ([#6463](#6463))
- **File Saving**: Marked `saveAllFiles` as a `WriteAction` to ensure proper file saving actions. ([#6459](#6459))
- **Zip Handling**: Updated to rely on the Bazel zipper instead of the standard zip utility. ([#6462](#6462))

#### Enhancements
- **Orphaned Modules**: Added an extension point to detect whether a module is to be orphaned or retained. ([#6467](#6467))

mai93 (approved):

Issue #6451, opened 2024-05-29
Windows Bazel integration in CLion


bashtavenko:
### Description of the bug:

Please see full description in [JetBrains](https://youtrack.jetbrains.com/issue/CPP-38861#focus=Comments-27-9839704.0-0).Ba

### Which category does this issue belong to?

CLion

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Please see above

### Which Intellij IDE are you using? Please provide the specific version.

CLion

### What programming languages and tools are you using? Please provide specific versions.

C++

### What Bazel plugin version are you using?

2024.05.07.0.1-api-version-241

### Have you found anything relevant by searching the web?

Not much

### Any other information, logs, or outputs that you want to share?

To summarize the issue that I linked above.

On CLion Sync Project, there is an error

```
Warning:Unable to check compiler version for "C:\users\sb\_bazel_sb\wi3ucxsc\external\bazel_tools~cc_configure_extension~local_config_cc\vc_installation_error_x64.bat".
stderr: " 
The target you are compiling requires Visual C++ build tools. 
Bazel couldn't find a valid Visual C++ build tools installation on your machine. 
 
Visual C++ build tools seems to be installed at C:\Program Files\LLVM\bin\clang-cl.exe 
But Bazel can't find the following tools: 
    VCVARSALL.BAT, cl.exe, link.exe, lib.exe, ml64.exe 
for x64 target architecture 
 
Please check your installation following https://bazel.build/docs/windows#using 
 
"
stdout: ""
Check if running the compiler with --version works on the cmdline.
```

As a result, CLion fails to process any c++ files.  

I assume that Bazel works on both MSVC and CLang on Windows, is that right?

tpasternak:
@ujohnny @LeFrosch could you please take a look at this problem?

LeFrosch:
As far as I know we do not support MSVC #4265. I am not sure if we are planning to add support for MSVC? @ujohnny

PR #6500, opened 2024-06-24
CLion MSVC Support

LeFrosch:
# Checklist

details.

Issue number: #6451

# Description of this change
Fixes windows specific issues and adds support for MSVC. Since, the `MSVCCompilerToVersionCacheService` is only available since 231 all earlier versions are not supported.

Limitations:
- Debugging only using LLDB
- Path mapping is not correct when debugging


cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -184,6 +189,35 @@ private ImmutableMap<TargetKey, BlazeCTargetInfoService.TargetInfo> calculatePer
     return infoMap.build();
   }
 
+  private CompilerSpecificSwitchBuilder selectSwitchBuilder(
```
ujohnny, line 192: static?

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -184,6 +189,35 @@ private ImmutableMap<TargetKey, BlazeCTargetInfoService.TargetInfo> calculatePer
...
+    }
+    if (CompilerVersionUtil.isMSVC(version)) {
+      return new MSVCSwitchBuilder();
+    }
+
+    // default to gcc
+    return new GCCSwitchBuilder();
+  }
+
+  private CidrCompilerSwitches buildSwitchBuilder(
```
ujohnny, line 210: static?

ujohnny (changes requested):

ujohnny (approved):

Issue #6490, opened 2024-06-17
TrustAwareProjectCreator should not show when REMOTE_DEV_TRUST_PROJECTS is set


brian-mcnamara:
### Description of the bug:

For our remote development environment, we have been using REMOTE_DEV_TRUST_PROJECTS environment flag that is stated in remote-dev-server.sh (as part of IJ bundle). However we are now getting "Trust Project" dialogs at start due to https://github.com/bazelbuild/intellij/blob/master/base/src/com/google/idea/blaze/base/project/TrustAwareProjectCreator.java

This is unexpected since the REMOTE_DEV_TRUST_PROJECTS flag should trust projects

### Which category does this issue belong to?

Intellij

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

This dialog appears when a new bazel project is opened for the first time. 

### Which Intellij IDE are you using? Please provide the specific version.

241.17890.1

### What programming languages and tools are you using? Please provide specific versions.

N/A

### What Bazel plugin version are you using?

2024.06.04.0.2-api-version-241

### Have you found anything relevant by searching the web?

https://github.com/bazelbuild/intellij/blob/master/base/src/com/google/idea/blaze/base/project/TrustAwareProjectCreator.java

### Any other information, logs, or outputs that you want to share?

N/A

brian-mcnamara:
@mai93 FYI

PR #6521, opened 2024-06-26
trusted project: respect intellij settings controlling project trustness

ujohnny:
# Checklist

details.

# Description of this change

Do not show dialogs to trust if trust checks are disabled in the intellij platform. Project creation code is handled in a custom way in bazel plugin because any project is trusted by default in intellij. However, bazel plugin always creates a fake project in .cl/ij/aswb folder, so the plugin should handle trust checks by itself.

### DIFF
--- a/base/src/com/google/idea/blaze/base/project/TrustAwareProjectCreator.java
+++ b/base/src/com/google/idea/blaze/base/project/TrustAwareProjectCreator.java
@@ -48,6 +48,10 @@ public class TrustAwareProjectCreator implements ExtendableBazelProjectCreator {
   /** Returns true if the user has trusted the project. */
   @Override
   public boolean canCreateProject(@Nullable BuildSystemName buildSystemName) {
+    if (TrustedProjects.isTrustedCheckDisabled()) {
+      return true;
+    }
+
     var trustText = IdeBundle.message("untrusted.project.dialog.trust.button");
     var dontOpenText = IdeBundle.message("untrusted.project.open.dialog.cancel.button");
 


mai93 (approved):

mai93:
is this related to https://github.com/bazelbuild/intellij/issues/6490?

tpasternak:
> is this related to #6490?

yes it is

dkashyn-sfdc (commented):
Thanks!

brian-mcnamara:
Thanks! It may be worth checking @tpasternak if TrustedProjects should come out of Internal status, most of this function could be replaced by other methods in that class

tpasternak:
> Thanks! It may be worth checking @tpasternak if TrustedProjects should come out of Internal status, most of this function could be replaced by other methods in that class

Sorry I don't understand

brian-mcnamara:
> > Thanks! It may be worth checking @tpasternak if TrustedProjects should come out of Internal status, most of this function could be replaced by other methods in that class
> 
> Sorry I don't understand

Right now https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/ide/trustedProjects/TrustedProjects.kt#L14
is marked internal, however the logic in this repo could also use `getProjectTrustedState` as well. I'm just unsure if this project prefers not to use internal apis

mai93:
do we need to add a similar check to `BazelGuard` https://github.com/bazelbuild/intellij/blob/master/base/src/com/google/idea/blaze/base/execution/BazelGuard.java#L31

brian-mcnamara:
> do we need to add a similar check to `BazelGuard` https://github.com/bazelbuild/intellij/blob/master/base/src/com/google/idea/blaze/base/execution/BazelGuard.java#L31?

We do not have any extensions for that EP so we do not need a check there to fix this issue.

ujohnny:
> do we need to add a similar check to BazelGuard

no because it does not ask for trust. 

> is marked internal, however the logic in this repo could also use `getProjectTrustedState` as well. I'm just unsure if this project prefers not to use internal apis

We can't use `getProjectTrustedState` because we don't have a project yet. It's all because plugin creates a "fake" project in ijwb/clwb and all created projects are trusted by default. So because of this custom behavior in bazel plugin we have to ask explicitly in TrustAwareProjectCreator.

ujohnny:
I checked this with gateway and I do not observe trust dialogs, so I'm moving forward to merge

tpasternak:
@mai93 @ujohnny can we cut a hotfix release with this change?

mai93:
released in version 2024.06.04.0.3

PR #6562, opened 2024-07-10
Fixes after sync freeze on MacOS 

LeFrosch:
# Checklist

details.

Issue number: 

# Description of this change
CLion freezes every time after sync because the `XcodeCompilerSettings` are resolved under read lock. 

I think we can drop the read lock for the entire workspace update since as far as I can tell nothing actually requires a read lock. And the actual update of the workspace is performed under write lock.

Or we could just move the compiler settings resolution before the read lock is taken. But dropping the read lock might prevent freezes in future.


ujohnny:
I see that a while ago RA was pushed down in a045eabbfae7944b02d0aa5d7026f57d5e5d8853, but I'm not sure what part of update actually requires RA. Is it so that no runtime warnings for missing RA are reported? It could be that `OCWorkspace.getConfigurations` requires it, but it's not mentioned in the doc.

ujohnny (approved):
The purpose of removed RA is unclear to me and it's not required by the code called.

Issue #6600, opened 2024-07-29
Correct handling of debugging of `py_binary` targets when there is a transition


andponlin-canva:
### Description of the bug:

**Note:** Please see associated PR [here](https://github.com/bazelbuild/intellij/pull/6605).

I have marked all IDEs because I assume they will all experience the same problem when handling Python targets but am only testing with Intelli-J.

When using a pinned Python interpreter, the expected behavior is that any `py_binary` targets that use the pinned interpreter will continue to be able to be debugged. The actual behavior is that the `py_binary` that uses the pinned interpreter is no longer able to use the debugger features in the IDE.

This problem is happening because the Rule type comes through from the aspect as `_transition_py_binary` and the logic in the plugin does not recognize this as a `py_binary`.

### Which category does this issue belong to?

Intellij, GoLand, CLion, PyCharm, Android Studio

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Please find attached an example project 
[`handle-rule-transitions-example-20240729.tgz`](https://github.com/user-attachments/files/16408222/handle-rule-transitions-example-20240729.tgz).

1. Unpack the tar-ball
2. Run the binary with `bazelisk run ":bin"` and observe the output text `Heizung`.
3. Open the workspace in the Intelli-J IDE with the Bazel plugin installed
4. Open `BUILD.bazel` and right click on the `bin` target.
5. Choose "Debug..."
6. Edit the configuration
7. Change "Bazel command:" to `run`
8. Click OK
9. Observe that the option to debug is showing as disabled

It should be the case that this `py_binary` is able to be debugged.

### Which Intellij IDE are you using? Please provide the specific version.

Intelli-J 2024.1

### What programming languages and tools are you using? Please provide specific versions.

Python 3.10

### What Bazel plugin version are you using?

Recent

### Any other information, logs, or outputs that you want to share?

Screenshot without the patch;

<img width="128" alt="handle-rule-transitions-example-before" src="https://github.com/user-attachments/assets/5906fc27-d8c9-4baa-9430-1f25c6c320d3">

Screenshot with the patch noting that the debug is now enabled;

<img width="128" alt="handle-rule-transitions-example-after" src="https://github.com/user-attachments/assets/b50acdc6-66e9-4b1f-979b-4cc285a4f534">

andponlin-canva:
Change merged - thanks.

tpasternak:
Thank you for the contribution!

[...]

PR #6605, opened 2024-07-31
handling for rules that involve transitions

andponlin-canva:
# Checklist

details.

Issue number: [6600](https://github.com/bazelbuild/intellij/issues/6600)

# Description of this change

When a Python rule is used such as `py_binary` and the rule is using a pinned Python interpreter, the `ruleName` is coming through as `_transition_py_binary` from the aspect but is still a `py_binary`. This change will still detect the transitioned rule name as `Kind.Provider.create("py_binary", LanguageClass.PYTHON, RuleType.BINARY)` from `PythonBlazeRules.java` so that debugging works for `py_binary`-s that are using the pinned interpreter.

See the issue discussion for further details and an example.

### DIFF
--- a/base/src/com/google/idea/blaze/base/model/primitives/Kind.java
+++ b/base/src/com/google/idea/blaze/base/model/primitives/Kind.java
@@ -45,6 +45,13 @@ import javax.annotation.Nullable;
 public abstract class Kind {
 
   /**
+   * When a rule has a transition applied to it then it will present with this
+   * prefix on it.
+   */
+
+  private final static String RULE_NAME_PREFIX_TRANSITION = "_transition_";
+
+  /**
    * Provides a set of recognized blaze rule names. Individual language-specific sub-plugins can use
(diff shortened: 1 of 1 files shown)


tpasternak:
This seems not to be enough, the targets with pinned seem not to contain `PyInfo`

https://github.com/bazelbuild/intellij/blob/6aa0ee48c0bf3460cdab290cf35377dca870a53d/aspect/intellij_info_impl.bzl#L362-L363

tpasternak:
The root cause is that, pinning the repo requires to load non-builtin rules_python. When you do `load('@python_3_11:defs.bzl", "py_binary")`, the py_binary rule exports  `@@rules_python//python/private/common:providers.bzl%PyInfo` instead of `PyInfo`

tpasternak:
Ok, but it still fixes debugging. Correct import is a separate issue

tpasternak (approved):

Issue #6599, opened 2024-07-29
Correct handling of `imports` rule attribute on `py...` rules


andponlin-canva:
### Description of the bug:

**Note:** See the associated [PR](https://github.com/bazelbuild/intellij/pull/6606).

I have marked all IDEs because I assume they will all experience the same problem when handling Python targets.

Some of the `rules_python` rules such as `py_library` and `py_binary` have the ability to supply an `imports` attribute. For example consider the following Bazel repo setup;

```
//
├── WORKSPACE
└── aaa
    └── bbb
        └── ccc
            ├── BUILD.bazel
            ├── ddd
            │   └── eee
            │       ├── doo.py
            │       └── fff
            │           └── poe.py
            └── zzz
                └── main.py
```

The `BUILD.bazel` may carry the following rule;

```
py_library(
    name = "lib",
    srcs = [
        "ddd/eee/doo.py",
        "ddd/eee/fff/poe.py",
    ],
...
)
```

At the point of use in `main.py`, some Python library functions in those files would be;

```
from aaa.bbb.ccc.ddd.eee.doo import print_fruit
from aaa.bbb.ccc.ddd.eee.fff.poe import print_tree
```

If the rule is changed to introduce the `imports` attribute...

```
py_library(
  name = "lib",
...
  imports = ["ddd"]
...
)
```

...then the associated Python module path is rooted at the same point as the `BUILD.bazel` file and includes the `imports`. At the point of use in `main.py`, the reference to the library functions would become;

```
from eee.doo import print_fruit
from eee.fff.poe import print_tree
```

The expected behavior of the Intelli-J plugin would be to identify the use of `imports` attribute in `py_...` rules and to adjust the module paths accordingly. The actual behavior seems to be to ignore the `imports` attribute.

Changes in can be made to support the `imports` feature. The data from the `imports` attribute needs to be fed through from the executed aspect via the proto and can then be post-processed in `AbstractPyImportResolverStrategy` to replicate the observed behavior of Bazel.

### Which category does this issue belong to?

Intellij, GoLand, CLion, PyCharm, Android Studio

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Please find attached a tar-ball [`support-py-imports-example-20240729.tgz`](https://github.com/user-attachments/files/16407837/support-py-imports-example-20240729.tgz).

1. Unpack the tar-ball
[...]

PR #6606, opened 2024-07-31
Implement py imports handling

andponlin-canva:
# Checklist

details.

Issue number: [6599](https://github.com/bazelbuild/intellij/issues/6599)

# Description of this change

Some `py_...` rules such as `py_binary` and `py_library` allow for an `imports` attribute. The `imports` attribute allows control over how the Python files are vended in terms of modules. This change will mean that the use of the `imports` attribute in Bazel targets is reflected in how the modules are auto-completed and recognized in the IDE.

### DIFF
--- a/python/src/com/google/idea/blaze/python/resolve/provider/BazelPyImportResolverStrategy.java
+++ b/python/src/com/google/idea/blaze/python/resolve/provider/BazelPyImportResolverStrategy.java
@@ -51,15 +51,6 @@ public class BazelPyImportResolverStrategy extends AbstractPyImportResolverStrat
     if (source.isGenerated() || !source.getRelativePath().endsWith(".py")) {
       return null;
     }
-    QualifiedName name = fromRelativePath(source.getRelativePath());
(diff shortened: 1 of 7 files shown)


python/src/com/google/idea/blaze/python/resolve/provider/BazelPyImportResolverStrategy.java
```diff
@@ -51,15 +51,6 @@ protected QualifiedName toImportString(ArtifactLocation source) {
     if (source.isGenerated() || !source.getRelativePath().endsWith(".py")) {
       return null;
     }
-    QualifiedName name = fromRelativePath(source.getRelativePath());
```
andponlin-canva, line 54: I have observed that in this situation, the associated `py_library` is already coming through with an `imports` and so this PR will handle this situation and there should no longer be a need to perform this work-around.

tpasternak (approved):
Thank you @andponlin-canva

Issue #6613, opened 2024-08-06
Add build folder action to plugin


kitbuilder587:
### Description of the feature request:

You can't build folder, you can only sync it. Building could be much faster. For example in our project sync of entire folder takes x2 time more than build. I've already added this feature to my local Bazel plugin. I've added button to compile directory below the partial sync. 
<img width="338" alt="image" src="https://github.com/user-attachments/assets/41b80dbd-088b-4f5d-a4f2-c170d628cb55">

It would be great if we can contribute it

### What underlying problem are you trying to solve with this feature?

Faster build to check compile errors without sync

tpasternak:
Closed with #6615

PR #6615, opened 2024-08-06
Added build folder action

kitbuilder587:
# Checklist

details.

Issue number: [`#6613`](https://github.com/bazelbuild/intellij/issues/6613)

# Description of this change
Added action to compile folder
image


base/src/com/google/idea/blaze/base/sync/sharding/BlazeBuildTargetSharder.java
```diff
@@ -111,6 +109,9 @@ private enum ShardingApproach {
 
   private static ShardingApproach getShardingApproach(
       SyncStrategy parallelStrategy, ProjectViewSet viewSet) {
+    if (parallelStrategy == SyncStrategy.SERIAL_NOT_EXPAND) {
```
tpasternak, line 112: So this overrides the `shard_sync` setting in the PV, how about just keeping the old behavior here?

base/src/com/google/idea/blaze/base/actions/CompileFolderAction.java
```diff
@@ -0,0 +1,73 @@
...
+    BlazePackage blazePackage = BuildFileUtils.getBuildFile(project, virtualFile);
+
+    boolean enabled = blazePackage != null;
+    presentation.setVisible(enabled || !ActionPlaces.isPopupPlace(e.getPlace()));
+    presentation.setEnabled(enabled);
+
+    ActionPresentationHelper.of(e)
+            .disableIf(virtualFile == null)
+            .disableIf(!virtualFile.isDirectory())
+            .setText(virtualFile.getName() + "/...:all")
```
tpasternak, line 64: No the text is "folder/...alll", can we change it to "Compile folder/...all"?

base/src/com/google/idea/blaze/base/sync/sharding/BlazeBuildTargetSharder.java
kitbuilder587: The issue here is that compile folder with shard_sync=true takes sooo much more time. For example, one of our project's folder compiles in 17s with shard_sync=true and in 1s with shard_sync=false. The main reason why we added this button was that sync time was very slow, and we wanted to speed up compile errors check. Ordinary user won't set shard_sync to false. We can add another setting to set shard_sync for build folder action, but I think in cases when OOM happens, you can just use Partial Sync instead

tpasternak: Oh, that's a good point. Btw what do you think about setting the default PV in `tools/intellij/.managed.bazelproject` with `shard_sync: true`?.

Example here https://github.com/bazelbuild/bazel/blob/master/tools/intellij/.managed.bazelproject

kitbuilder587: Personally, I think It's a good idea, but I don't have a lot of knowledge about how often people want to use `shard_sync=true`. In our project we use it quite rarely, but I can't expand my PV on all users of the plugin. 
Should I create another PR [here](https://github.com/bazelbuild/bazel/blob/master/tools/intellij/.managed.bazelproject) or we can stay on current realization or mb another ideas?

tpasternak: @LeFrosch what do you think about it?

kitbuilder587: Oh, actually if I am not mistaken, `shard-sync=true` by default. We can delete this property in enum

kitbuilder587: Deleted SERIAL_NOT_EXPAND. Sorry for disappointing, that's my bad

kitbuilder587: But anyway logically it would be better if `shard-sync` won't affect build. Often`shard-sync=true` is used for full sync of project if OOM happens. `shard-sync=true` will affect the build folder speed. (17s and 1s is big diff) I don't think that it will be used for the whole project often. With `shard-sync=true` build folder action is useless, because the diff between build and sync time isn't big. I think SERIAL_NOT_EXPAND anyway will be better

tpasternak (approved):

base/src/com/google/idea/blaze/base/sync/sharding/BlazeBuildTargetSharder.java
tpasternak: Okay, but I can't guarantee it will be permanent. If it turns out to be misleading, we’ll adjust it to align with the shard_sync setting.

PR #6698, opened 2024-08-30
feat(clion): Add working directory to CLion Run configurations

shs96c:
# Checklist

details.

Issue number: N/A

# Description of this change

In CMake CLion, one has the option of setting a working directory under which the binary will be run. This is useful for debugging single binaries (i.e. they need no runfiles) under different contexts.

This PR adds some of that capability to the Bazel plugin by creating a setting for run configurations, and adding it to the Cidr debug process handlers.

Limitations:

 * For now, this only applies to CLion debugging configurations. Other debuggers will have to be modified to use the setting, as well as non-debug configurations.
 * Previously, we defaulted to the runfiles dir as the pwd if there was one. The explicit working dir has precedence over it now.

242549858-02f0d9e3-aa6a-4e78-9bf3-90f25b296e60

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/state/WorkingDirectoryState.java
+++ b/base/src/com/google/idea/blaze/base/run/state/WorkingDirectoryState.java
@@ -1,0 +1,97 @@
+package com.google.idea.blaze.base.run.state;
+
+import com.google.common.base.Strings;
+import com.google.idea.blaze.base.ui.UiUtil;
+import com.intellij.openapi.fileChooser.FileChooserDescriptorFactory;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.ui.LabeledComponent;
+import com.intellij.openapi.ui.TextBrowseFolderListener;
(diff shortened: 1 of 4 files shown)


tpasternak:
Unfortunately the setting doesn't work with gdbserver wrapper. I think it could be fixed here https://github.com/bazelbuild/intellij/blob/9343e7e9dce962b5e5358bede4083e3a57a607bd/clwb/src/com/google/idea/blaze/clwb/run/BlazeGDBServerProvider.java#L100-L115

maybe we could just add `cd <path> && ` there

ujohnny:
Indeed since `gdbserver` is a default run type for linux I'd appreciate to see its support. Though I'm unsure if we should `cd` in `run_under` as I'd expect it to inherit cwd from bazel cwd

ujohnny (changes requested):
Thank you for the PR! Could you please update `BlazeGDBServerProvider` as @tpasternak suggested?

PR #6693, opened 2024-08-30
Fix reading `MatchedMatcher` data in `BlazePyTracebackParser`

mielnikk:
# Checklist

- [ ] I have filed an issue about this change and discussed potential changes with the maintainers.
- [x] I have received the approval from the maintainers to make this change.
- [x] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: #6652 

# Description of this change

Referencing groups in `MatchedMatcher` by `1` and `2` indexes was not correct, so I changed it to how it's done in Python plugin's `PyTracebackParser`.

PR #6745, opened 2024-09-09
Update changelog for v2024.09.10 release

satyanandak:


mai93:
### Bug Fixes
- https://github.com/bazelbuild/intellij/pull/6693
- https://github.com/bazelbuild/intellij/pull/6741

mai93 (approved):

tpasternak:
Hey, we are releasing external labels indexing and completions, thanks to @mtoader

tpasternak:
We didn't have such a tradition, however this contribution is so great, I would consider giving credits, WDYT? @mai93

mai93:
> We didn't have such a tradition, however this contribution is so great, I would consider giving credits, WDYT? @mai93

I agree, feel free to propose a suggestion on the comment. Thanks @mtoader!

tpasternak:
so maybe let's just add `credits: @mtoader (https://github.com/mtoader)`?

tpasternak:
idk how about just `(credits:  )`

mtoader:
That still doesn't work well without the bazel flag changing `~` to `+` (the one becoming default true in bazel 8). So it will be limited in exposure until i have the notification in place to have the users update said flag. 

Also the more important bit is https://github.com/bazelbuild/intellij/pull/6742 (since it enabled completion past the `@xxx//` in these cases :) i would mention this issue after we land this and add the notification bubble since that will be a "proper" user experience.

mai93:
thanks @mtoader for the clarification, @satyanandak can you remove the "Features" part before submitting?

PR #6756, opened 2024-09-12
Fix unrecognized options by bazel mod command

LeFrosch:
Build flags defined in the project view are forwarded to `bazel mod` which breaks sync, since `--define` is not a known option for the mod command.

Not sure if excluding build flags from the mod command is the right way to go. But fixes the sync.

### DIFF
--- a/base/src/com/google/idea/blaze/base/command/BlazeFlags.java
+++ b/base/src/com/google/idea/blaze/base/command/BlazeFlags.java
@@ -77,7 +77,9 @@ public final class BlazeFlags {
     for (BuildFlagsProvider buildFlagsProvider : BuildFlagsProvider.EP_NAME.getExtensions()) {
       buildFlagsProvider.addBuildFlags(project, projectViewSet, command, invocationContext, flags);
     }
-    flags.addAll(expandBuildFlags(projectViewSet.listItems(BuildFlagsSection.KEY)));
+    if (command != BlazeCommandName.MOD) {
+      flags.addAll(expandBuildFlags(projectViewSet.listItems(BuildFlagsSection.KEY)));
+    }
     if (invocationContext.type() == ContextType.Sync) {
       for (BuildFlagsProvider buildFlagsProvider : BuildFlagsProvider.EP_NAME.getExtensions()) {
         buildFlagsProvider.addSyncFlags(


LeFrosch:
CC @mtoader

mtoader:
The right thing here is more complicated unfortunately since what is really needed is something similar to how `.bazelrc` works. 

This is ok for now but it will break in different ways too, e.g. adding `--incompatible_use_plus_in_repo_names` in `build_flags` will make mod show different canonical names than what build would use.

LeFrosch:
Yes I thought that this might be the case. Tbh, not really sure how the bazel mod stuff works. But if it is okay with you I would like to merge this for now to fix sync.

mtoader:
> Yes I thought that this might be the case. Tbh, not really sure how the bazel mod stuff works. But if it is okay with you I would like to merge this for now to fix sync.

I'm ok with it. I have to add a notifier anyway to let people know that the plus flag should be set for proper completion to work anyway. I'll recommend them set it in the `.blazerc` file.

mtoader:
> > Yes I thought that this might be the case. Tbh, not really sure how the bazel mod stuff works. But if it is okay with you I would like to merge this for now to fix sync.
> 
> I'm ok with it. I have to add a notifier anyway to let people know that the plus flag should be set for proper completion to work anyway. I'll recommend them set it in the `.blazerc` file.

I also don't have write/review access here so I can't help further unfortunately

mtoader:
@agluszak @tpasternak I think the semantics of `build_flags` in the new plugin could be smarter. Maybe a `.blazerc` fragment? I would avoid this kind of issues ?

tpasternak:
Sorry, my responses might be delayed until the end of next week. Sure, providing bazelrc-like inheritance would be nice, however it might require extra maintenance to make sure we agree with bazel on that.

It could also be done for free just with --bazelrc CLI option, it would be even a low hanging fruit in the old plugin, to provide 'bazelrc` section that would create a bazelrc file on the fly. It could replace all other sections and perform better 

But speaking about this particular problem, it seems like the `mod` command simply doesn't inherit flags from build so we can't pass build_flags there.

mtoader:
> Sorry, my responses might be delayed until the end of next week. Sure, providing bazelrc-like inheritance would be nice, however it might require extra maintenance to make sure we agree with bazel on that.
> 
> It could also be done for free just with --bazelrc CLI option, it would be even a low hanging fruit in the old plugin, to provide 'bazelrc` section that would create a bazelrc file on the fly. It could replace all other sections and perform better
> 
> But speaking about this particular problem, it seems like the `mod` command simply doesn't inherit flags from build so we can't pass build_flags there.

Yea. That was my thinking also. I opened a ticket to think about this in the new plugin (now might be a good time to update section semantics if this is what it would be needed and since I'm adding blazerc support there it would be easier too).

PR #6759, opened 2024-09-16
CLion Virtual Include Fixes

LeFrosch:
Addresses problems with the virtual includes handler:

- Fixes absolute strip prefix
- Fallback to virtual includes for include prefix (no longer an error)
- Fallback to virtual includes if a source file is generated

Also removes the `-ibazel` support and adds integration tests.


anonrig:
Is there an ETA to this pull-request changes?

ujohnny:
> Is there an ETA to this pull-request changes?

I'll check it this week

ujohnny:
@LeFrosch I almost understand the whole change, but let's have a chat on Monday as the PR description is too concise imo. I've also noticed it does not work with bzlmod (but I believe the original code didn't work either), so maybe consider this as a next step (or improve here?)

ujohnny (changes requested):

ujohnny (approved):

PR #6761, opened 2024-09-17
Fix Security Vulnerability with the PR labeler and remove label github actions.

sgowroji:
Fix Security Vulnerability with the PR labeler and remove label github actions

As per the b/364700434, there can be a potential risk when using pull_request_target. This change converts the pull_request_target to pull_request event trigger.


mai93 (approved):

sgowroji:
Hello @mai93, I reverting this PR changes as it looks like we do need **pull_request_target** to be able to add labels. Since this action doesn't check out any code from the PR, and  security risk is very limited.

PR #6767, opened 2024-09-18
Fix view_project_root with overlapping directories

jdai8:
# Checklist

details.

# Description of this change
When `view_project_root: true` and there are overlapping directories, for example:

The repo contains these directories:
- foo
- foo/bar
- foo/baz

And .bazelproject is:
```
view_project_root: true

directories:
  foo
  foo/bar
```

Then, `foo/baz` will incorrectly be excluded. This is unexpected because it's contained under the `foo` directory.

This commit fixes the behavior, so that files are not excluded if they are under any root directory.


tpasternak (approved):

tpasternak:
Thank you @jdai8

PR #6841, opened 2024-10-03
fix: Query Sync: don't check for validity of excludes that are system…

tpasternak:
…-excludes (bazel-bin, .ijwb, etc.)

Demo for bazelbuild/bazel:

https://github.com/user-attachments/assets/89322a24-bf70-42a8-8636-163ace1326ed

# Checklist

details.

# Description of this change

### DIFF
--- a/querysync/java/com/google/idea/blaze/qsync/project/ProjectDefinition.java
+++ b/querysync/java/com/google/idea/blaze/qsync/project/ProjectDefinition.java
@@ -42,7 +42,8 @@ public abstract class ProjectDefinition {
           /* includes= */ ImmutableSet.of(),
           /* excludes= */ ImmutableSet.of(),
           /* languageClasses= */ ImmutableSet.of(),
-          /* testSources= */ ImmutableSet.of());
+          /* testSources= */ ImmutableSet.of(),
+          /* systemExcludes = */ ImmutableSet.of());
 
   /**
    * Project includes, also know as root directories. Taken from the users {@code .blazeproject}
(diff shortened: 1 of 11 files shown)


querysync/java/com/google/idea/blaze/qsync/project/ProjectDefinition.java
```diff
@@ -85,6 +90,11 @@ public QuerySpec deriveQuerySpec(Context<?> context, Path workspaceRoot) throws
       }
     }
     for (Path exclude : projectExcludes()) {
+     if(systemExcludes().contains(exclude.toString())){
+       // We don't have to check if these dirs are valid for queries, because they also don't fall into //... wildcard
+       continue;
```
mai93, line 95: but do we still need to exclude them via `result.excludePath(exclude);` or not?

tpasternak: I'm not sure, can investigate it later. The whole point of this PR is to avoid `isValidPathForQuery` for these paths which might be super expensive

tpasternak: So that's basically an optimization, it doesn't change the behavior

mai93: what happens if we added `result.excludePath(exclude);` instead of `continue`? if we get the same performance improvement then it seems like the right thing to do

tpasternak: I haven't tried but I think it could go to the query, as a substitution unconditionally, without validity check and hence could result in a crash

mai93: I see that makes sense, I can merge it as is to release for 2024.3 but we may come up with a better optimization later

mai93 (approved):

PR #6842, opened 2024-10-04
chore: copy some aspect files to the client side to prepare for templating

xuansontrinh:
# Description of this change

The purpose of this PR is to ready the plugin for templating some aspect files, e.g. `java_info.bzl`.
This is just an intermediate step and the next change may considerably change what is put here.


tpasternak:
The test is most likely false positive. Let's disable it for now, so @xuansontrinh can move forward and I (or someone else) will take care of the test

tpasternak (approved):

PR #6854, opened 2024-10-05
Backport .bazelrc support from Hirschgarten

agluszak:
Original author in Hirschgarten: @mtoader 

TODO:
- [ ] split off changes related to introduction of kotlin
- [ ] backport tests


mtoader:
Heads up: This will be a source of issues for now :) the flags database is minimal and incomplete and it will trigger all kinds of red things in .bazelrc files. I would disable the annotator in here until the known flags database is more coherent.

mtoader:
> Heads up: This will be a source of issues for now :) the flags database is minimal and incomplete and it will trigger all kinds of red things in .bazelrc files. I would disable the annotator in here until the known flags database is more coherent.

Later edit: I see that it is not in here. That's good

base/BUILD
```diff
@@ -18,10 +19,9 @@ load(
     "intellij_unit_test_suite",
 )
 
-java_library(
+kt_jvm_library(
```
tpasternak, line 22: Can we keep bazelrc as a separate library?

agluszak: why?

tpasternak: Base is quite big already, just to speed up compilation time. Also, compilation time of base would be even slower if we started conversion to kotlin

PR #6871, opened 2024-10-09
SyncListener for aspect templating

LeFrosch:
Moved the aspect templating code to a new SyncListener. This allows for tighter integration with the rest of the plugin and for proper error handling.


LeFrosch:
Committed the MODULE file by accident

tpasternak (approved):

Issue #6725, opened 2024-09-04
Support for Python Rules that are Code Generators


andponlin-canva:
### Description of the feature request:

Please see accompanying PR [6726](https://github.com/bazelbuild/intellij/pull/6726).

## Plugin recognizing code generators

When working with Python in Bazel, it is sometimes desirable to generate code using an Action configured from a Rule. In this situation, the Rule that configures the Action might return a `PyInfo` and to all intents and purposes behave as a `py_library`.

In the Intelli-J environment, the Rule would however not be recognized and the outputs of the Rule would not be seen as source code. This means that in logic that uses the generated code, these sources would simply be missing and would show as errors in the IDE.

It would be ideal if there were a means by which Rules could signal to the IDE that their outputs rather than their inputs (`srcs` attribute) supplied the sources for the Rule.

## Plugin handling source supplied as a directory

Code generators are often not able to specify the precise list of output files ahead of time. This might be because the list of files is contingent on the contents of an input such as schema. For this reason the outputs from a code-generators' Rule might be a directory. In this case, The IDE is currently not able to handle a directory of Python source; it is wanting to work with files only. It would be better if the Plugin were able to work with source directories as well.

### Which category does this issue belong to?

Intellij, GoLand, CLion, PyCharm

### What underlying problem are you trying to solve with this feature?

At the present time when a Python Bazel Project uses the outputs of a code-generator Rule, the sources from the code generator are not recognized in the IDE and show as errors.

### What operating system, Intellij IDE and programming languages are you using? Please provide specific versions.

2024.1

### Have you found anything relevant by searching the web?

--

### Any other information, logs, or outputs that you want to share?

--

andponlin-canva:
## Before

If I open the example found at `examples/python/simple_code_generator` from the [PR branch](https://github.com/bazelbuild/intellij/pull/6726) with the Intelli-J / Bazel plugin installed as the mainline build from `master` (from the Marketplace) then I see the following when I sync and open the `main.py` file;

<img width="360" alt="before" src="https://github.com/user-attachments/assets/44c2ebb9-9292-4969-86e3-b57b7530fa20">

The behavior of the IDE confirms the anticipation;

- `print_rivers` will be found because it is not generated source; it is checked-in source via the `:static_lib` Bazel Target.
- the IDE will fail to "find" `print_cities` symbol because it is generated by a Bazel Rule `:generated_files_lib` that outputs single files. 
[...]

PR #6884, opened 2024-10-11
fix: Disable python codegen support by default

tpasternak:
# Checklist

details.

# Description of this change

### DIFF
--- a/base/src/com/google/idea/blaze/base/projectview/section/sections/EnablePythonCodegenSupport.java
+++ b/base/src/com/google/idea/blaze/base/projectview/section/sections/EnablePythonCodegenSupport.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2024 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


tpasternak:
@agluszak

agluszak (approved):

FrankPortman:
I happened to see this PR when browsing this PR and this change looks potentially significant to me as I have macros/rules that generate `PyInfo`/files -> `py_library` in my codebase. Is there any documentation for this change?

andponlin-canva:
> I happened to see this PR when browsing this repo and this change looks potentially significant to me as I have macros/rules that generate `PyInfo`/files -> `py_library` in my codebase. Is there any documentation for this change?

Hello @FrankPortman ; I added some comments [here](https://github.com/bazelbuild/intellij/blob/master/docs/python/code-generators.md) which also includes an example.

FrankPortman:
Thanks - if my understanding is correct then this flag and tag are required to make IntelliJ properly detect and remove the "red squiggle" from codegenned files. But my basic usage of both macros and rules to code generate Python files has not had issues with IntelliJ as of now. Is this change a general safety net or only needed if you intend to output directories rather than marking individual files?

andponlin-canva:
Yes the flag and tag will be required to make this work fully. The intention is to remove the "red squiggle" (see [here](https://github.com/bazelbuild/intellij/issues/6725#issuecomment-2389649698)) and to allow code-complete on generated code.

andponlin-canva:
> But my basic usage of both macros and rules to code generate Python files has not had issues with IntelliJ as of now

@FrankPortman ; If I understand you correctly, you are facading your Rules' files-based (no directory) output with a `py_library` which is working for you in Intelli-J? This change presumes that you might be using the Rule directly and the Rule has the option of having directory-based outputs. This may be the two differences from the approach that you have.

FrankPortman:
Thank you - that does sound like the distinction.

PR #6909, opened 2024-10-18
Intellij integration test

LeFrosch:
Moved CLion integration test to base as `BazelHeavyIntegrationTest` to share integration test logic. Also added a first simple integration test for ijwb.


ijwb/tests/heavy_integrationtests/com/google/idea/blaze/ijwb/SimpleTest.java
```diff
@@ -0,0 +1,25 @@
...
+    errors.assertNoErrors();
+
+    checkConfiguration();
+  }
+
+  private void checkConfiguration() {
+    final var javaFile = findProjectPsiFile("src/simple/Main.java");
+    final var kotlinFile = findProjectPsiFile("src/simple/Main.kt");
+
+    // not sure what to test here
```
LeFrosch, line 23: @tpasternak do you what I could use to verify a successful project setup here?

tpasternak: I would say that for start it is ok just to check if the source root is set correctly (so if any of the symbols in kotlin/java file can be found)

PR #6911, opened 2024-10-18
Use execroot for compiler working directory

LeFrosch:
Changes the working directory of the compiler to the current execroot. Also adds an integration test for the custom llvm toolchain.


base/src/com/google/idea/blaze/base/scope/BlazeContext.java
```diff
@@ -26,6 +26,7 @@
 import com.google.idea.blaze.common.Output;
 import com.google.idea.blaze.common.PrintOutput;
 import com.google.idea.blaze.exception.BuildException;
+import com.intellij.openapi.application.ApplicationManager;
```
ujohnny, line 29: looks unrelated (and the one below)

clwb/tests/projects/llvm_toolchain/WORKSPACE
```diff
@@ -0,0 +1,21 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
ujohnny, line 1: let's move to bzlmod for this test because we use bazel 7+ anyway

ujohnny (changes requested):

clwb/tests/projects/llvm_toolchain/WORKSPACE
LeFrosch: I migrated the test, but I cannot delete the WORKSPACE file because `rules_bazel_integration_test` requires a WORKSPACE file.

PR #6912, opened 2024-10-18
Removed unused CppSupportChecker

LeFrosch:
Removes the unused CppSupportChecker. The only implementation ClwbSupportsCpp returns always true.


LeFrosch:
Nvm, used by aswb.

PR #5220, opened 2023-08-08
Add handling in aspect for implementation_deps

sfc-gh-abalik:
# Checklist

- [x] I have filed an issue about this change and discussed potential changes with the maintainers.
- [x] I have received the approval from the maintainers to make this change.
- [x] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: https://github.com/bazelbuild/intellij/issues/5149

# Description of this change
Update the aspect to
- Propagate along `implementation_deps`
- Collect includes/defines from a target's `implementation_deps`

These changes are needed to enable intellisense, go to definition, etc in CLion for files that belong to targets using `implementation_deps`.

Issue #5149, opened 2023-07-26
CLion doesn’t sync header files under `includes` dirs from `implementation_deps`


sfc-gh-abalik:
### Description of the bug:

- Given a `cc_library` (`lib1`) that includes another `cc_library` (`lib2`) via `implementation_deps`
- `lib2` exposes `lib2.h` via `includes`

```
cc_library(
    name = "lib2",
    srcs = ["lib2.cc"],
    hdrs = ["include/lib2.h"],
    includes = ["include"],
)

cc_library(
    name = "lib1",
    srcs = ["lib1.cc"],
    hdrs = ["lib1.h"],
    implementation_deps = [
        ":lib2",
    ],
)

cc_binary(
    name = "helloworld",
    srcs = ["main.cc"],
    deps = [
        ":lib1",
    ],
)
```

CLion will not find `lib2.h` when syncing files in `lib1`

```
# File: lib1.cc

#include "main/lib1.h"
#include "lib2.h"     // => red squiggly here with "'lib2.h' file not found"   

std::string get_string() {
  print_something();  // => function from lib2.h not found  
  return "Hello world";
}
```

As a result intellisense, go to definition, etc don’t work for `lib2.h`.

A couple workarounds that fix this
- Changing from `implementation_deps` to `deps`
- Changing the `#include` to `"include/lib2.h"` (e.g. not relying on the include dir)

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Here's a full workspace with the example above. Just open the `.bazelproject` and look at `lib1.cc`.
[repro.tar.gz](https://github.com/bazelbuild/intellij/files/12167960/repro.tar.gz)

### Which Intellij IDE are you using? Please provide the specific version.

CLion 2023.1.2

### What programming languages and tools are you using? Please provide specific versions.

C++

### What Bazel plugin version are you using?

Bazel for CLion 2023.07.04.0.1-api-version-231

### Have you found anything relevant by searching the web?

https://github.com/bazelbuild/intellij/issues/3352 seems similar, but for `cc_binary`

sfc-gh-abalik:
[...]

PR #6919, opened 2024-10-21
CLion support for cc_library implementation deps

LeFrosch:

Issue number: #5149

# Description of this change

Rebased PR #5220 and added an integration test for implementation_deps.


aspect/intellij_info_impl.bzl
```diff
@@ -513,16 +521,27 @@ def collect_cpp_info(target, ctx, semantics, ide_info, ide_info_file, output_gro
     target_copts = _do_starlark_string_expansion(ctx, "copt", target_copts, extra_targets)
 
     compilation_context = target[CcInfo].compilation_context
+    defines = compilation_context.defines.to_list()
+    includes = compilation_context.includes.to_list()
+    quote_includes = compilation_context.quote_includes.to_list()
+    system_includes = compilation_context.system_includes.to_list()
+
+    if hasattr(ctx.rule.attr, "implementation_deps"):
```
ujohnny, line 529: can't this information be obtained from `CcInfo` provider?

LeFrosch: I don't think there is a way since bazel treats impl deps context different from the main context. CcInfo only exposes the main context as far as I can tell. 

https://github.com/bazelbuild/bazel/blob/b85db78de3b25cb7f5c9cb0c7afe388f75e040e3/src/main/starlark/builtins_bzl/common/cc/cc_compilation_helper.bzl#L449

PR #6941, opened 2024-10-29
fix(qsync): Enable basic syntax highlighting in 'analysis disabled' mode

tpasternak:
## Before:
Screenshot 2024-10-29 at 17 36 26

## After
Screenshot 2024-10-29 at 16 44 22

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -412,7 +412,7 @@
                  key="bazel.read.external.workspace.data"/>
     <registryKey defaultValue="true"
                  description="Enable basic syntax highliting in Query Sync's 'Analysis Disabled' mode"
-                 key="bazel.qsync.enable.basic.highlighting.in.non.analysis.mode"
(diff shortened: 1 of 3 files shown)


base/src/META-INF/blaze-base.xml
```diff
@@ -409,6 +410,9 @@
     <registryKey defaultValue="true"
                  description="Causes the plugin to read external workspace data, enabling features such as code completion for external targets. This may incur the cost of an additional 'bazel mod' call, which could cause issues with older versions or setups."
                  key="bazel.read.external.workspace.data"/>
+    <registryKey defaultValue="true"
+                 description="Enable basic syntax highliting in Query Sync's 'Analysis Disabled' mode"
+                 key="bazel.qsync.enable.basic.highlighting.when.analysis.disabled"/>
```
dkashyn-sfdc, line 415: For flags trailing `.disabled` can be confusing... Usually it is attributed to flag default state.

tpasternak: right!

agluszak (approved):

Issue #6943, opened 2024-10-30
Query sync(and probably regular as well) is using working sets even when it is not supposed to


dkashyn-sfdc:
### Description of the bug:

![image](https://github.com/user-attachments/assets/050a024c-1dcd-48dc-b024-94c90528deb9)
is happening with QS mode even when Working Sets are not enabled.

With our repo it can take extra 2 mins to sync it perceived speed of QS is affected in a negative way because of this extra op

PR #6944, opened 2024-10-30
Use git info only when Expand to Working Set is enabled

dkashyn-sfdc:
# Checklist

details.

Issue number: (https://github.com/bazelbuild/intellij/issues/6943)

# Description of this change

### DIFF
--- a/base/src/com/google/idea/blaze/base/qsync/ProjectQuerierImpl.java
+++ b/base/src/com/google/idea/blaze/base/qsync/ProjectQuerierImpl.java
@@ -22,6 +22,7 @@ import com.google.idea.blaze.base.async.executor.BlazeExecutor;
 import com.google.idea.blaze.base.logging.utils.querysync.SyncQueryStats;
 import com.google.idea.blaze.base.logging.utils.querysync.SyncQueryStatsScope;
 import com.google.idea.blaze.base.scope.BlazeContext;
+import com.google.idea.blaze.base.settings.BlazeUserSettings;
 import com.google.idea.blaze.base.vcs.BlazeVcsHandlerProvider.BlazeVcsHandler;
 import com.google.idea.blaze.common.vcs.VcsState;
 import com.google.idea.blaze.exception.BuildException;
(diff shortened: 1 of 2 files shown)


base/src/com/google/idea/blaze/base/qsync/ProjectQuerierImpl.java
```diff
@@ -120,15 +121,17 @@ public PostQuerySyncData update(
...
-    SyncQueryStatsScope.fromContext(context)
-        .ifPresent(stats -> stats.setSyncMode(SyncQueryStats.SyncMode.DELTA));
-    logger.info(
-        String.format(
-            "Starting partial query update; upstream rev=%s; snapshot path=%s",
-            vcsState.map(s -> s.upstreamRevision).orElse("<unknown>"),
-            vcsState.flatMap(s -> s.workspaceSnapshotPath).map(Object::toString).orElse("<none>")));
-
+    Optional<VcsState> vcsState = Optional.empty();
+    if (BlazeUserSettings.getInstance().getExpandSyncToWorkingSet()) {
```
dkashyn-sfdc, line 125: Probably other option should be added for QS yet even QS-relevant delta calculation will be taxng in some cases.

Also Sync to Working Set is enabled by default, iirc. So default QS behavior won't change.

base/src/com/google/idea/blaze/base/sync/ProjectStateSyncTask.java
```diff
@@ -125,7 +126,7 @@ private SyncProjectState getProjectState(BlazeContext context, BlazeSyncParams p
         createBazelInfoFuture(context, syncFlags, params.syncMode());
 
     ListenableFuture<WorkingSet> workingSetFuture;
-    if (params.addWorkingSet() || params.syncMode() == SyncMode.FULL) {
+    if (params.addWorkingSet() || (BlazeUserSettings.getInstance().getExpandSyncToWorkingSet() && params.syncMode() == SyncMode.FULL)) {
```
dkashyn-sfdc, line 129: Don't know when this one is used but it is bad to ignore the setting anyway so we need to double check here.

tpasternak: Oh, params.addWorkingSet already checks for that, but I'll merge

tpasternak (approved):

base/src/com/google/idea/blaze/base/sync/ProjectStateSyncTask.java
dkashyn-sfdc: I want to avoid unconditional `params.syncMode() == SyncMode.FULL` this is quite recent change from September

PR #6971, opened 2024-11-06
Remove soft-fail marker from the CI tests

mai93:
it hides failures on the under-dev IDE versions and it is no longer needed since the compatibility changes for the new IDE versions are usually sent out in the same PR now

### DIFF
--- a/.bazelci/aspect.yml
+++ b/.bazelci/aspect.yml
@@ -41,8 +41,6 @@ tasks:
       - --notrim_test_configuration
     test_targets:
       - //aspect/testing/...
-    soft_fail:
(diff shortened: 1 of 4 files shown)


mai93:
@LeFrosch can you check the CLion failures with 2024.3?

tpasternak:
We need to wait until the next clion release - it will come with updated JS plugin

ujohnny:
From my point this was intentional, see #6856. I'm fine to keep under-dev builds failing, because it leaves us some space to work on it, but I guess right now we're just missed to update oldest and latest oss-stable refs.

PR #6977, opened 2024-11-07
2024.3 update for plugins

ujohnny:
# Description of this change

switch from 241 + 242 to 242 + 243


ujohnny:
waiting for CLion beta build publication, will update `MODULE.bazel` and then CI should be green

tpasternak (approved):

PR #7040, opened 2024-11-22
Modify details about sharing code between this repo and AOSP

mai93:


tpasternak (approved):

LeFrosch:
Sounds good to me. Let's hope we can keep up with picking commits :)

PR #7048, opened 2024-11-25
fix issues with aspect templates not available in query sync mode

xuansontrinh:
Before the change: debug on java binary is failing due to not having access to aspect templates
image

After the change: debug on java binary is successful
image

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/aspects/strategy/SyncAspectTemplateProvider.java
+++ b/base/src/com/google/idea/blaze/base/sync/aspects/strategy/SyncAspectTemplateProvider.java
@@ -56,6 +56,15 @@ public class SyncAspectTemplateProvider implements SyncListener {
     prepareProjectAspect(project);
   }
 
+  @Override
+  public void onQuerySyncStart(Project project, BlazeContext context) {
+      try {
+          prepareProjectAspect(project);
+      } catch (SyncFailedException e) {
+          throw new RuntimeException(e);
+      }
+  }
+
   private void prepareProjectAspect(Project project) throws SyncFailedException {
     var manager = BlazeProjectDataManager.getInstance(project);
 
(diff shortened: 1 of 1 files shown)


base/src/com/google/idea/blaze/base/sync/aspects/strategy/SyncAspectTemplateProvider.java
```diff
@@ -131,16 +140,18 @@ private void writeLanguageInfo(
...
+    var projectData = Optional.ofNullable(manager.getBlazeProjectData()); // It can be empty on intial sync. Fall back to no language support
     var activeLanguages = projectData.map(it -> it.getWorkspaceLanguageSettings().getActiveLanguages()).orElse(ImmutableSet.of());
-    var isJavaEnabled = activeLanguages.contains(LanguageClass.JAVA)
-            && blazeProjectData != null
-            && blazeProjectData.getExternalWorkspaceData().getByRepoName("rules_java") != null;
-    var isPythonEnabled = activeLanguages.contains(LanguageClass.PYTHON)
-            && blazeProjectData != null
-            && blazeProjectData.getExternalWorkspaceData().getByRepoName("rules_python") != null;
+//    var externalWorkspaceData = projectData.map(BlazeProjectData::getExternalWorkspaceData).orElse(null);
+    var isJavaEnabled = activeLanguages.contains(LanguageClass.JAVA);
```
tpasternak, line 147: how about keeping it for legacy sync, and disabling the check for query sync?

tpasternak: so instead of commenting out, we could add `|| isQuerySync`

tpasternak (changes requested):

base/src/com/google/idea/blaze/base/sync/aspects/strategy/SyncAspectTemplateProvider.java
xuansontrinh: sure, will do.

tpasternak:
Thank you @xuansontrinh

tpasternak (approved):

Issue #7043, opened 2024-11-25
Play button in executable code and in GTest tests displays `Nothing here` when using Nova engine


sfc-gh-mgalindo:
### Description of the bug:

When CLion Nova is enabled, the play button next to the code editor does nothing for main functions and gtest tests.

<img width="643" alt="Screenshot 2024-11-24 at 8 35 50 PM" src="https://github.com/user-attachments/assets/9c07ad32-ae41-42b8-a459-a7c6430c1089">

### Which category does this issue belong to?

CLion

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.
For `gtest`

- Checkout branch from https://github.com/bazelbuild/intellij/pull/7044
- Open the sample project under `clwb/tests/projects/simple` 
- Enable Nova in CLion
- Sync the bazel project
-  Open the file `main/hello-test.cc`
- Click in the play button next to `TEST` in the code editor and see no action is a available

For `main`
- Open the sample project under `clwb/tests/projects/simple` 
- Enable Nova in CLion
- Sync the bazel project
- Open the file `main/hello-world.cc`
- Click in the play button next to `main` in the code editor  and see no action is a available

### Which Intellij IDE are you using? Please provide the specific version.

CLion 2024.3 RC Build #JBC-243.21565.198, built on November 12, 2024

### What programming languages and tools are you using? Please provide specific versions.

C++

### What Bazel plugin version are you using?

2024.10.22-stable

Issue #6981, opened 2024-11-08
CLion unit test gutter icon shows "nothing here"


ivan-golub:
### Description of the bug:

Unable to run cc_tests (google tests or plain main function) from gutter icon. Same setup works with CMake
 
<img width="1390" alt="Screenshot 2024-11-08 at 2 28 48 PM" src="https://github.com/user-attachments/assets/b43dc4a9-569f-42b1-b0d0-0bd6f1b73fc2">

### Which category does this issue belong to?

CLion

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Import bazel project -> examples/cpp/simple_project/proj.bazelproject. 
2. Wait until loading and indexing completes
3. Navigate to lib/greeting_test.cc
4. Try to press gutter "play" button

### Which Intellij IDE are you using? Please provide the specific version.

2024.2.3

### What programming languages and tools are you using? Please provide specific versions.

C++, Bazel 7.4.0

### What Bazel plugin version are you using?

2024.10.22.0.1-api-version-242

LeFrosch:
Maybe related to: https://youtrack.jetbrains.com/issue/CPP-36051

connoranderson:
@LeFrosch I think you are correct. I'm seeing that with Nova enabled, the unit tests are not runnable. With Nova disabled, I can run / debug individual tests. It'd be nice to get Nova supported here as it supposedly is significantly faster than legacy, according to 2024.3 release notes.

ghost:
I am seeing this issue again on clion2025.2.1 with bazel-for-clion-2025-08.04.

<img width="472" height="440" alt="Image" src="https://github.com/user-attachments/assets/5434caec-fe29-4607-87b3-4016b55dfbea" />

ghost:
with nova disabled, the "test and run" options show up agian

<img width="500" height="240" alt="Image" src="https://github.com/user-attachments/assets/e31a8f31-b92e-4188-8041-0e95c0936ca4" />

LeFrosch:
Thank you for submitting this. I'll take a look as soon as I find the time.

LeFrosch:
@pinkfrog9 So I cannot reproduce the issue of the missing run configs. After I imported the ray project successfully, all the run configs where there (in Nova). However, I noticed an issue we have with make variables that initially kept me from properly importing the project. I am working on a workaround for this: #7896 

Did you had any problems importing the project when the run configs do not show up? 

[...]

PR #7062, opened 2024-11-26
Adds support for the google test run button in nova

LeFrosch:

Issue number: #6981 and #7043

# Description of this change

Adds support for the gutter run button for google tests. The button itself is provided by the backend and we only need to register a provider to actually run these tests.


intellij_platform_sdk/BUILD.clion243
```diff
@@ -59,6 +59,8 @@ java_import(
         "plugins/clion-test-google/lib/*.jar",
         "plugins/clion-test-catch/lib/*.jar",
         "plugins/clion-test-boost/lib/*.jar",
+        "plugins/clion-radler/lib/*.jar",
```
ujohnny, line 62: a possible solution: define this in a separate `java_import` statement, use that library in https://github.com/bazelbuild/intellij/pull/7062/files#diff-242eb0d3dc78241907aa160705183dba65b9f3cb747ee146887eaf081f7622f5R18 and let's keep using classic so far for development purposes

ujohnny (changes requested):

intellij_platform_sdk/BUILD.clion243
LeFrosch: I played around with multiple options, I think just using the suppressed plugin set options is the easiest for now. This allows us later to run test for classic and radler. Tho radler requires the path to the backend binary which I could not get working for now.

ujohnny (changes requested):
proposed patch set is shared privately

ujohnny (approved):

PR #7083, opened 2024-11-29
chore: set explicit 2024.3 version and enable qs in the project view

tpasternak:
# Checklist

details.

# Description of this change

### DIFF
--- a/ijwb/ijwb.bazelproject
+++ b/ijwb/ijwb.bazelproject
@@ -14,7 +14,7 @@ targets:
 workspace_type: intellij_plugin
 
 build_flags:
-  --define=ij_product=intellij-latest
+  --define=ij_product=intellij-2024.3
 
 test_sources:
   */tests/unittests*
@@ -22,3 +22,5 @@ test_sources:
 
 additional_languages:
   kotlin
+
+use_query_sync: true


ijwb/ijwb.bazelproject
```diff
@@ -14,11 +14,13 @@ targets:
 workspace_type: intellij_plugin
 
 build_flags:
-  --define=ij_product=intellij-latest
+  --define=ij_product=intellij-2024.3
```
LeFrosch, line 17: Maybe intllij-oss-under-dev?

LeFrosch (changes requested):

ijwb/ijwb.bazelproject
LeFrosch: Because, sooner or later this will be outdated.

LeFrosch:
Also I don't think these `ijwb` and `clwb` specific configurations make sense with qsync. Since, the main benefit of qsync is that you can just import what you need. Maybe we can just add on `.bazelproject` file for the repo?

ijwb/ijwb.bazelproject
dkashyn-sfdc: `intellij-latest` is outdated by itself and using some dated version which is far from being "latest"

tpasternak:
> Maybe we can just add on .bazelproject file for the repo
Yeah, you're right!

ijwb/ijwb.bazelproject
LeFrosch: Sure, I don't mind the change. If you think it makes it easier for external contributors go ahead.

LeFrosch (approved):

Issue #7087, opened 2024-11-30
Allow to enable and disable Query Sync for existing projects


dkashyn-sfdc:
### Description of the feature request:

Currently, the only way to enable or disable QS is to re-create the project. 
It would be beneficial to allow to change this more dynamically and to enable for already present projects.

### Which category does this issue belong to?

Intellij

PR #7088, opened 2024-11-30
Reload project view, and make sure that BazelImportSettings respect p…

dkashyn-sfdc:
…roject view choice before performing sync calls.

# Checklist

details.

Issue number: [`7087`](https://github.com/bazelbuild/intellij/issues/7087)

# Description of this change


dkashyn-sfdc:
Last 2 runs after `git pull` and it is obvious that our project will benefit from QS. 
image

However, there are some gaps so "toggle" mechanism is essential.

base/src/com/google/idea/blaze/base/settings/Blaze.java
```diff
@@ -72,6 +78,38 @@ public static ProjectType getProjectType(@Nullable Project project) {
     return blazeImportSettings.getProjectType();
   }
 
+  /**
+   * This variant allows us to enable and disable Query Sync for already imported project.
+   * com.google.idea.blaze.base.settings.Blaze#getProjectType(com.intellij.openapi.project.Project) is called quite often
+   * so we cannot reload project view from for every of such call.
+   * This is why we have this special case to make sure that Sync respects project view selection if there is any.
+   */
+  public static ProjectType getProjectTypeBeforeSync(@Nonnull Project project) {
```
ujohnny, line 87: I've been also looking for a way to jump between qs/as on the fly, thanks for submitting this PR! Can we add some javadoc here to reflect somthing like `returns the project type with project view file lookup`? and maybe make the name more matching the action it does rather than a point it code where it should be called?

dkashyn-sfdc: There are 15 places where `Blaze.getProjectType*(project) == ProjectType.QUERY_SYNC` check is performed. So I wanted to have a special case for sync calls only. This is why name reflects this.

I don't really know how frequent the rest of places are hitting this method and saw it called from action apperance update, which cannot afford to reload file for sure.

So instead of having "get type with refresh" I'd rather keep "get type for sync and make sure the proper sync mode is used". Also we are updating `BazelImportSettings` here and it is ok to update those on sync but for the rest of ops I'd like to avoid this.

dkashyn-sfdc: I can rename it to `getSelectedProjectTypeBeforeSync` or so. The fact that `BazelImportSetings` are updated is important yet subtle here. Before this change it was set once during project creation and not touched, for whatever reason.

Overall `Blaze.getProjectType*(project) == ProjectType.QUERY_SYNC` is someway out of control... It is peppered over the code quite randomly and I'm a little bit worried about it causing trouble in the future even for cases beyond sync.

dkashyn-sfdc: `getRefreshedProjectType` or `getNotCachedBeforeSync` is another option... But then it will be a dilemma where to call those :)

base/src/com/google/idea/blaze/base/sync/data/DelegatingBlazeProjectDataManager.java
```diff
@@ -61,6 +54,27 @@ public BlazeProjectData loadProject(BlazeImportSettings importSettings) {
 
   @Override
   public void saveProject(BlazeImportSettings importSettings, BlazeProjectData projectData) {
+    if (projectType != Blaze.getProjectType(project)) {
+      initBlazeProjectDataDelegate(project);
```
dkashyn-sfdc, line 58: Probably we need to call `save` here for the old model... But this is a thing only for Aspect one.

dkashyn-sfdc: With 50% of non-test impls not using `saveProject` this interface definitely needs some cleanup...
```    
// TODO(mathewi) Tidy up the interface to remove this unnecessary stuff.
```

base/src/com/google/idea/blaze/base/sync/actions/IncrementalSyncProjectAction.java
```diff
@@ -48,7 +48,7 @@ protected void runSync(Project project, AnActionEvent e) {
   }
 
   public static void doIncrementalSync(Class<?> klass, Project project, @Nullable AnActionEvent e) {
-    if (Blaze.getProjectType(project) == ProjectType.QUERY_SYNC) {
+    if (Blaze.getProjectTypeBeforeSync(project) == ProjectType.QUERY_SYNC) {
```
tpasternak, line 51: So when I switch from legacy sync to aspect sync, QS is able to detect it's the first sync and falls to non-incremental sync. It happens here https://github.com/bazelbuild/intellij/blob/5ad3c3d5bf3d61b5184c3cc21e230abe994e5c43/base/src/com/google/idea/blaze/base/sync/actions/IncrementalSyncProjectAction.java#L54-L57

------------------

On the other hand when I switch back from QS to AS, it results in Incremental Sync which might cause some unexpected issues (For example leftoveers after QS in project model)

dkashyn-sfdc:
```
  private static boolean allTargetsBuild(SyncMode mode) {
    return mode == SyncMode.FULL || mode == SyncMode.INCREMENTAL;
  }
```

With
```
    } else {
      BlazeSyncManager.getInstance(project)
          .incrementalProjectSync(/* reason= */ "IncrementalSyncProjectAction");
    }
```


 
```
  public void incrementalProjectSync(String reason) {
    BlazeSyncParams syncParams =
        BlazeSyncParams.builder()
            .setTitle("Sync")
            .setSyncMode(SyncMode.INCREMENTAL)
            .setSyncOrigin(reason)
            .setAddProjectViewTargets(true)
            .setAddWorkingSet(BlazeUserSettings.getInstance().getExpandSyncToWorkingSet())
            .build();
    requestProjectSync(syncParams);
  }
```
And "Full" is:
```
  public void fullProjectSync(String reason) {
    BlazeSyncParams syncParams =
        BlazeSyncParams.builder()
            .setTitle("Full Sync")
            .setSyncMode(SyncMode.FULL)
            .setSyncOrigin(reason)
            .setAddProjectViewTargets(true)
            .setAddWorkingSet(BlazeUserSettings.getInstance().getExpandSyncToWorkingSet())
            .build();
    requestProjectSync(syncParams);
  }
```

dkashyn-sfdc: TBH I've been using "INCREMENTAL" all this time and there is nothing really incremental about it :)

tpasternak: well, this means `INCREMENTAL` was pretty good, theoretically it should always work

tpasternak: but I would fall back to FULL, for sure external repo completions in starlark won't work

dkashyn-sfdc: After checking the code I'm not that sure about
> QS is able to detect it's the first sync and falls to non-incremental sync

since nothing resets `com.google.idea.blaze.base.qsync.QuerySyncManager#loadedProject` for QS to do FULL or `onStartup`

tpasternak: well, here it ends up with calling `onStartup`

```

it just seems safer to me

ujohnny (approved):

PR #7092, opened 2024-12-02
fix: remove strongInfoHolder flag, so users can choose which plugin t…

tpasternak:
…o use on import

# Checklist

details.

# Description of this change


base/src/com/google/idea/blaze/base/project/AutoImportProjectOpenProcessor.java
```diff
@@ -67,7 +67,7 @@ public class AutoImportProjectOpenProcessor extends ProjectOpenProcessor {
   public @NotNull
   @Nls
   String getName() {
-    return Blaze.defaultBuildSystemName() + " Project";
+    return "Bazel (OG)";
```
jastice, line 70: OG is kind of internal terminology, let's consider if there is a more clear name. On the other hand, none of "native", "Google", "old" are clear names either :/

jastice (approved):

PR #7094, opened 2024-12-02
Basic qsync support in CLion for linux

LeFrosch:
This PR provides basic support for qsync in CLion, but only for linux with gcc so far. Issues addressed:

- cc_info_file and cc_toolchain_info propagation in the aspect
- project model setup in CLion

Also adds a simple integration test that imports the project using query sync.

### DIFF
--- a/clwb/tests/integrationtests/com/google/idea/blaze/clwb/QuerySyncTest.java
+++ b/clwb/tests/integrationtests/com/google/idea/blaze/clwb/QuerySyncTest.java
@@ -1,0 +1,55 @@
+package com.google.idea.blaze.clwb;
+
+import static com.google.common.truth.Truth.assertThat;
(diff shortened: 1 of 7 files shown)


clwb/tests/integrationtests/com/google/idea/blaze/clwb/QuerySyncTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+  private void checkAnalysis() throws ExecutionException {
+    final var success = enableAnalysisFor(findProjectFile("main/hello-world.cc"));
+    assertThat(success).isTrue();
+  }
+
+  private void checkCompiler() {
+    final var compilerSettings = findFileCompilerSettings("main/hello-world.cc");
+
+    // TODO: query sync selects the wrong compiler
```
ujohnny, line 44: let's create a GH issue and put a link here

PR #7100, opened 2024-12-03
Adds support for the run button on main functions in nova 

LeFrosch:
Builds up on the work of #7062 and adds support for binary run configurations in CLion. Double run button in nova will be fixed in 251.

### DIFF
--- a/clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
+++ b/clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2024 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+  val symbolsHost = RadSymbolsHost.getInstance(element.project)
+  return symbolsHost.isEntryPointOffset(element.containingFile.viewProvider.virtualFile, element.startOffset)
+}
+
+private fun findTargets(context: ConfigurationContext): Collection<TargetInfo> {
+  val virtualFile = context.location?.virtualFile ?: return emptyList()
+
+  val targets = SourceToTargetFinder.findTargetsForSourceFile(
+    context.project,
+    File(virtualFile.path),
```
ujohnny, line 45: let's update this occurrence here and in other files to `virtualFile.toNioPath().toFile()`

ujohnny (approved):

PR #7108, opened 2024-12-04
chore: Remove examples from bazelignore

tpasternak:
It seems to affect Renovate

# Checklist

details.

# Description of this change

### DIFF
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,1 +1,0 @@
-examples


ujohnny:
Shouldn't we even add more to it (ignore integration testdata) rather that remove it completely? cc @LeFrosch

tpasternak:
I'm not sure, but Renovate seems to ignore these directories

LeFrosch:
@ujohnny Integration test data cannot be ignored since they need to be accessed by the tests at runtime as far as I understand. They are “deleted packages”, see .bazelrc.

tpasternak:
So can I remove that, I don't even really see the point of it, while it seems that Revovate fails to upgrade projects there

sellophane (approved):

PR #7113, opened 2024-12-04
CLion failing integration test run (bazel 8.0.0rc6) 

LeFrosch:


hvadehra:
Looks like this was caused by https://github.com/bazelbuild/bazel/commit/489d08bf472f4568b3f7f478a8747a30cb956111

Adding test --zip_undeclared_test_outputs to your .bazelrc should fix things.

Issue #7114, opened 2024-12-04
Kotlin language version is not read correctly in newer rules_kotlin versions


jdai8:
### Description of the bug:

https://github.com/bazelbuild/rules_kotlin/pull/1157 removes legacy struct providers from rules_kotlin. https://github.com/bazelbuild/intellij/pull/6394 changes the IntelliJ plugin to read the new toolchain provider, however this doesn't seem to be working for us.

I needed to make some changes to read these correctly and can put up a PR.

However, I saw that the change to use the `_kt_toolchain` attribute was reverted here. https://github.com/bazelbuild/intellij/commit/a88c6db967cd173fa73aeb76d8c8c7bbd0f8be3d#diff-7e834e0a725805ad4dfd127ba905e50964c78307ddb756da99d15a0a1ea28a46L1117

Was this change intentional? I see some Google commits with mentions of upcoming toolchain changes as well, so curious what the current status of this is.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Use rules_kotlin 2.0.0, load it in IntelliJ and observe that the Kotlin language version is not set.

### Which Intellij IDE are you using? Please provide the specific version.

IntelliJ 2024.3

### What programming languages and tools are you using? Please provide specific versions.

Kotlin

### What Bazel plugin version are you using?

Built from source at 353dac83cee869af8f4a987873044beb63c29233

### Have you found anything relevant by searching the web?

N/A

### Any other information, logs, or outputs that you want to share?

N/A

PR #7115, opened 2024-12-04
Fix reading of _kt_toolchain attribute

jdai8:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/7114

# Description of this change

Fixes https://github.com/bazelbuild/intellij/issues/7114.

rules_kotlin has moved to using a special _kt_toolchain attribute, but it looks like the plugin is unable to properly read them. This commit fixes the output group `intellij-info-kt` (instead of `intellij-info-kotlin`) and adds rule kinds so that the plugin recognizes the toolchain.

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -1127,7 +1127,7 @@ def collect_kotlin_toolchain_info(target, ide_info, ide_info_file, output_groups
     ide_info["kt_toolchain_ide_info"] = struct(
         language_version = kt.language_version,
     )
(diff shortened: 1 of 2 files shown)


tpasternak:
ok, I understand the issue, but I don't understand how this PR fixes that 😅 Don't we have to read `target[ToolchainInfo].lanugage_version` then?

tpasternak:
The previous usage of _kt_toolchain was removed accidentally, we have to bring it back

tpasternak:
Ok, I'll create a follow up PR

tpasternak (approved):

PR #7127, opened 2024-12-06
Sync select files

odisseus:
# Checklist

details.

# Description of this change


base/src/com/google/idea/blaze/base/qsync/QuerySyncAsyncFileListener.java
```diff
@@ -183,7 +193,7 @@ public void afterQuerySync(Project project, BlazeContext context) {
                     return;
                   }
                   if (requester.changePending.get()) {
-                    requester.requestSyncInternal();
+                    requester.requestSyncInternal(ImmutableList.of());
```
tpasternak, line 196: This won't work. Consider a sequence
1. File 1 is added
2. Sync Starts
3. File 2 is added
4. Sync Finishes
5. `afterQuerySync` is called (this line)

File 2 won't be synced. The `afterQuerySync` is supposed to be executed here nad catch File 2

tpasternak (changes requested):

tpasternak:
how about something lile that? https://github.com/bazelbuild/intellij/pull/7168

tpasternak:
Unfortunately I just noticed that the refresh logic is deeply tight with vcs ( Bazel -> Expand Sync To Working Set) and it is not enough to just use `QuerySyncActionStatsScope.createForFiles`

The problem is here, the `requiresFullUpdate` and `calculateAffectedPackages` ignore what is set in `createForFiles` and they fall back to git-based delta sync or full sync. 

https://github.com/bazelbuild/intellij/blob/93295b0e7c1f3b4e7f400fc5c58659b112d1a1a4/querysync/java/com/google/idea/blaze/qsync/ProjectRefresher.java#L80-L87

tpasternak:
FYI https://cs.android.com/android-studio/platform/tools/adt/idea/+/39d8b044d86236f31598799a368b787a7ad8f64f

PR #7130, opened 2024-12-06
[AOSP-pick] Do not wait for the write lock

LeFrosch:
Cherry pick AOSP commit [cff9f98e002cc7f7d7d379090f9dd4655221a278](https://cs.android.com/android-studio/platform/tools/adt/idea/+/cff9f98e002cc7f7d7d379090f9dd4655221a278).

in `BuildArtifactCacheDirectory` as it may cause a deadlock if the
wait queue looks like "W", "R".. when the read lock is already held.

The first read lock stamp is released only after all reads are processed
and they won't until the write is processed but it can't proceed.

Bug: 373957467
Test: n/a
Change-Id: Ice28bae0288296fa8ed17c31187b578affb8d46c

AOSP: cff9f98e002cc7f7d7d379090f9dd4655221a278


sellophane (approved):

LeFrosch:
Blocked by failing Mac CI, probably caused by us not using the JBR :(

mai93:
@LeFrosch passing after retry

Issue #7133, opened 2024-12-06
Attempt to convert a Git commit hash to Long


odisseus:
### Description of the bug:

I'm getting this stack trace:
```
2024-12-06 14:17:11,921 [ 552040]   WARN - #com.google.idea.blaze.base.qsync.QuerySyncProjectData - For input string: "d6371d1e598f96616c4dcb885a7fd122aede818c"
java.lang.NumberFormatException: For input string: "d6371d1e598f96616c4dcb885a7fd122aede818c"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Long.parseLong(Long.java:709)
	at java.base/java.lang.Long.parseLong(Long.java:832)
	at com.google.idea.blaze.base.qsync.QuerySyncProjectData.lambda$getBlazeVersionData$5(QuerySyncProjectData.java:156)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at com.google.idea.blaze.base.qsync.QuerySyncProjectData.getBlazeVersionData(QuerySyncProjectData.java:153)
	at com.google.idea.blaze.base.sync.aspects.strategy.SyncAspectTemplateProvider.lambda$getLanguageStringMap$4(SyncAspectTemplateProvider.java:149)
	at java.base/java.util.Optional.map(Optional.java:260)
	at com.google.idea.blaze.base.sync.aspects.strategy.SyncAspectTemplateProvider.getLanguageStringMap(SyncAspectTemplateProvider.java:149)
	at com.google.idea.blaze.base.sync.aspects.strategy.SyncAspectTemplateProvider.writeLanguageInfos(SyncAspectTemplateProvider.java:124)
	at com.google.idea.blaze.base.sync.aspects.strategy.SyncAspectTemplateProvider.prepareProjectAspect(SyncAspectTemplateProvider.java:91)
	at com.google.idea.blaze.base.sync.aspects.strategy.SyncAspectTemplateProvider.onQuerySyncStart(SyncAspectTemplateProvider.java:62)
	at com.google.idea.blaze.base.qsync.QuerySyncProject.sync(QuerySyncProject.java:263)
	at com.google.idea.blaze.base.qsync.QuerySyncProject.syncWithCurrentSnapshot(QuerySyncProject.java:254)
	at com.google.idea.blaze.base.qsync.QuerySyncProject.deltaSync(QuerySyncProject.java:250)
[...]

PR #7140, opened 2024-12-09
Support for Git commit hashes when building BlazeVersionData

odisseus:
# Checklist

details.

Issue number: #7133

# Description of this change


tpasternak:
Well, it's a blaze client version, I would prefer to keep it null.

tpasternak:
This way is just putting sort-of-a-random-number there which can backlash in future. Particularly because the order seems to matter (there is a method `blazeClientIsAtLeastCl`)

odisseus:
You are right, the commit prefixes have no particular order, so my code violates that expectation. We could try to find the latest tag, or the commit date, but there is no guarantee that they will be ordered either.

Issue #7200, opened 2024-12-19
Autoimport does not work for bazel project with .idea in the project root


ujohnny:
### Description of the bug:

When some project has both `.idea` and `MODULE.bazel/WORKSPACE` in the project root, then project is opened as "intellij" project and bazel integration is not involved.

### What Bazel plugin version are you using?

2024.12.03.0.1-api-version-243

ujohnny:
caused by #7092

ujohnny:
In case if `org.jetbrains.bazel` plugin is installed override with `bazel.project.prefer.google.plugin` registry key.

PR #7201, opened 2024-12-19
Allow to boost `.ijwb` over BSP and .idea folder.

dkashyn-sfdc:
# Checklist

details.

Issue number: [`7200`](https://github.com/bazelbuild/intellij/issues/7200)

# Description of this change

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -412,7 +412,7 @@
     <registryKey defaultValue="true"
                  description="By default Bazel plugin takes control and opens project if .ijwb folder is present. This property can be used to disable this behavior to allow to open .idea or other project models if they are exists."
                  key="bazel.project.auto.open.if.present"/>
-    <registryKey defaultValue="true"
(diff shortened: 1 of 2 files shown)


base/src/META-INF/blaze-base.xml
```diff
@@ -412,6 +412,9 @@
     <registryKey defaultValue="true"
                  description="By default Bazel plugin takes control and opens project if .ijwb folder is present. This property can be used to disable this behavior to allow to open .idea or other project models if they are exists."
                  key="bazel.project.auto.open.if.present"/>
```
jastice, line 414: can't we just reuse this one and set it false by default?

dkashyn-sfdc: present is mutually exclusive of auto-import, so no

ujohnny (approved):
The reason for this particular implementation is the goal to avoid "select project" dialog in headless mode if BSP plugin is installed. The additional check for BSP plugin existence is just for backward compatibility.

PR #7216, opened 2025-01-10
Fix mac agent native errors

LeFrosch:
Recently I started to notice some native errors on our Mac CI agents. This is probably because we are running the tests with java 17. Ideally we would use the right JBR version for the selected ij_product, but this would require a more significant change. Bumping java to 21 is a temporary workaround.


ujohnny:
IMO it should be enough to run on 21 and migration of JDK to 21 is not mandatory to address the issue. Otherwise LGTM

ujohnny (approved):
IMO it should be enough to run on 21 and migration of JDK to 21 is not mandatory to address the issue. Otherwise LGTM

PR #7229, opened 2025-01-17
[AOSP-pick] Remove default helper methods in BuildResultHelper

LeFrosch:
Cherry pick AOSP commit [ccf72a5c77927d76c9e7cf55378ea1fc6aec6334](https://cs.android.com/android-studio/platform/tools/adt/idea/+/ccf72a5c77927d76c9e7cf55378ea1fc6aec6334).

Utility methods are supposed to be static in preferably in utility
classes or as Kotlin extension methods.

This is to enable safe and bigger refactoring of this area of code.

Bug: n/a
Test: n/a
Change-Id: I9dd7c5515c3289b05042299c6aea3d5c70c70180

AOSP: ccf72a5c77927d76c9e7cf55378ea1fc6aec6334

### DIFF
--- a/clwb/BUILD
+++ b/clwb/BUILD
@@ -83,6 +83,7 @@ java_library(
         "//intellij_platform_sdk:plugin_api",
         "//sdkcompat",
         "//clwb/sdkcompat",
+        "//shared",
     ],
 )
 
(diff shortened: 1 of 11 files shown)


clwb/BUILD
```diff
@@ -83,6 +83,7 @@ java_library(
         "//intellij_platform_sdk:plugin_api",
         "//sdkcompat",
         "//clwb/sdkcompat",
+        "//shared",
```
ujohnny, line 86: is the order correct here, ie why `clwb` is inbetween?

LeFrosch: No it's not. I'll format the entire file in a new PR?

ujohnny (approved):

PR #7254, opened 2025-01-23
WIP - Unrecognized attribute regression with bazel 8 

LeFrosch:
Drop unrecognized attribute inspection. ~~I have never seen a real issue found by this inspection, let's just drop it?~~ Seems like a regression, maybe we can fix the regression, I would like to investigate a bit further.

Screenshot 2025-01-23 at 11 14 21 AM


LeFrosch:
Maybe there is also an underlying issue when using query sync. Taking a look again.

sellophane (approved):

LeFrosch:
Seems to be caused by missing load statements, so actually the inspection is correct. Maybe a bit too aggressive.

PR #7257, opened 2025-01-27
make label parsing more resilient

LeFrosch:
In some cases query sync might return some labels that cannot be validated later on. The actual fix would be to fix query sync to exclude these invalid labels; this just addresses the symptom.

### DIFF
--- a/base/src/com/google/idea/blaze/base/qsync/QuerySyncProjectData.java
+++ b/base/src/com/google/idea/blaze/base/qsync/QuerySyncProjectData.java
@@ -37,6 +37,7 @@ import com.google.idea.blaze.qsync.project.QuerySyncLanguage;
 import com.intellij.openapi.diagnostic.Logger;
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


base/src/com/google/idea/blaze/base/qsync/QuerySyncProjectData.java
```diff
@@ -102,7 +102,8 @@ public Collection<ProjectTarget> getReverseDeps(Path sourcePath) {
   public ImmutableList<TargetInfo> targets() {
     if (blazeProject.isPresent()) {
       return blazeProject.get().getTargetMap().values().stream()
-          .map(TargetInfo::fromBuildTarget)
+          .map(TargetInfo::tryFromBuildTarget)
+          .flatMap(Optional::stream)
```
ujohnny, line 106: maybe do this without `Optional` and just `filter(Objects::nonNull)`?

LeFrosch: I like Optionals, but yes probably unnecessary heap pressure.

ujohnny (approved):

PR #7259, opened 2025-01-28
Replace VFS with a manifest file for copying aspects

LeFrosch:
Since vfs is not requires some hacks with to refresh files and might
break in future it should be avoided. Further more, vfs is not available
in some tests and this way logic for copying aspects can be reused.


aspect/build_defs.bzl
```diff
@@ -11,22 +19,63 @@ _java_8_transition = transition(
...
+        transitive =
+            [it[DefaultInfo].files for it in ctx.attr.files] +
+            [it[DefaultInfo].files for it in ctx.attr.jars],
+    )
 
-    for jar in ctx.attr.jars:
-        files.extend(jar[DefaultInfo].files.to_list())
+    return [
+        DefaultInfo(files = files),
+        AspectFilesInfo(files = files, namespace = ctx.attr.namespace),
```
ujohnny, line 31: do we need to store `files` in this provider? is there a reason we can't use `DefaultInfo`?

ujohnny (approved):

PR #7193, opened 2024-12-18
xcode: query xcode parameters from provider based on attributes

ujohnny:
# Checklist

- [ ] I have filed an issue about this change and discussed potential changes with the maintainers.
- [ ] I have received the approval from the maintainers to make this change.
- [ ] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: `<please reference the issue number or url here>`

# Description of this change

Bazel 8 introduced new provider located in apple_common module, which is not available in cquery context, so the implementation was updated to just find a provider with required attributes to make implementation also work for bazel 7 and older

fixes #6604

PR #7266, opened 2025-01-31
Followup: query xcode parameters from provider based on attributes

LeFrosch:
Followup for #7193 after #7202 was merged.

Bazel 8 introduced new provider located in apple_common module,
which is not available in cquery context, so the implementation
was updated to just find a provider with required attributes to
make implementation also work for bazel 7 and older.

### DIFF
--- a/clwb/tests/integrationtests/com/google/idea/blaze/clwb/SimpleTest.java
+++ b/clwb/tests/integrationtests/com/google/idea/blaze/clwb/SimpleTest.java
@@ -61,7 +61,7 @@ public class SimpleTest extends ClwbIntegrationTestCase {
 
     final var scriptLines = Files.readAllLines(compilerSettings.getCompilerExecutable().toPath());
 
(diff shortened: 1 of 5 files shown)


clwb/tests/integrationtests/com/google/idea/blaze/clwb/SimpleTest.java
```diff
@@ -44,4 +48,20 @@ private void checkTest() {
...
+    }
+
+    final var compilerSettings = findFileCompilerSettings("main/test.cc");
+
+    final var compilerExecutable = compilerSettings.getCompilerExecutable();
+    assertThat(compilerExecutable).isNotNull();
+
+    final var scriptLines = Files.readAllLines(compilerSettings.getCompilerExecutable().toPath());
+
+    assertContainsPattern("export DEVELOPER_DIR=/.*/Xcode.*.app/Contents/Developer", scriptLines);
```
ujohnny, line 64: I wonder why we do not have similar environment being injected for Apple Clang support in CLion.

ujohnny (approved):

PR #7269, opened 2025-02-03
Bump CLion to fixed version

LeFrosch:
I must have missed this during review. But the CLion version was pointing to the latest EAP, breaking builds due to sha mismatch.


ujohnny (approved):

LeFrosch:
New EAP breaks integration test. Will check this later :(

PR #7270, opened 2025-02-03
Remove dependency on rules_java in aspect BUILD file

LeFrosch:
The BUILD file in the aspects still used `java_binary` in some places, which can lead to issues if a dependency on rules_cc is added to the MODULE file. Example error: 

```
        File "/home/daniel_brauner/.cache/bazel_bazel_daniel_brauner/6e5a2eb6331f062f80d8d791ac8e71b6/external/rules_java~~toolchains~remote_java_tools/BUILD", line 7, column 60, in <toplevel>
                load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library", "cc_proto_library")
Error: file '@rules_cc//cc:defs.bzl' does not contain symbol 'cc_proto_library'
```

I don't think the `java_binary` targets are actually required and instead we could just run the deploy jars directly. This removes the dependency on rules_java.


base/src/com/google/idea/blaze/base/buildview/BazelExecService.kt
```diff
@@ -97,6 +97,7 @@ class BazelExecService(private val project: Project) : Disposable {
       handler.addProcessListener(object : ProcessListener {
         override fun onTextAvailable(event: ProcessEvent, outputType: Key<*>) {
           ctx.output(PrintOutput.process(event.text))
+          LOG.debug("BUILD OUTPUT: " + event.text)
```
ujohnny, line 100: clean this up?

LeFrosch: No I would suggest to keep this. Because this way it is pretty easy to get the build log in integration tests: `cat <log file> | grep "BUILD OUTPUT"`

ujohnny (approved):
Overall change addresses possible incompatibility between bundled rules_java and rules_cc for bazel 6 and 7. However it's more of a workaround as we're not yet ready to commit for aspect split per ruleset/language and now it relies on backward compatibility of java_common and expects that `JavaRuntimeInfo` is available from `java_runtime`.

Issue #7258, opened 2025-01-27
Windows Query Sync: Invallid package names


LeFrosch:
### Description of the bug:

When importing the project with query sync, it is not possible to create a `Bazel IntelliJ Plugin` run configuration. The following exception is reported instead. I guess there is a `/` and `\` mismatch. 

```
2025-01-14 13:12:48,263 [  52724] SEVERE - #c.i.i.p.PluginManager - Invalid package name 'clwb\tests\projects\virtual_includes\bazel-virtual_includes\external\platforms\os': package names may contain only A-Z, a-z, 0-9, '/', '-', '.', ' ', '$', '(', ')' and '_'

java.lang.IllegalArgumentException: Invalid package name 'clwb\tests\projects\virtual_includes\bazel-virtual_includes\external\platforms\os': package names may contain only A-Z, a-z, 0-9, '/', '-', '.', ' ', '$', '(', ')' and '_'

    at com.google.idea.blaze.base.model.primitives.Label.create(Label.java:44)

    at com.google.idea.blaze.base.dependencies.TargetInfo.fromBuildTarget(TargetInfo.java:170)

    at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)

    at com.google.common.collect.CollectSpliterators$1WithCharacteristics.lambda$forEachRemaining$1(CollectSpliterators.java:72)

    at java.base/java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Streams.java:104)

    at com.google.common.collect.CollectSpliterators$1WithCharacteristics.forEachRemaining(CollectSpliterators.java:72)

    at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)

    at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)

    at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)

    at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)

    at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)

    at com.google.idea.blaze.base.qsync.QuerySyncProjectData.targets(QuerySyncProjectData.java:106)

[...]

PR #7271, opened 2025-02-03
Store the package and name of a Label as strings rather than paths, to avoid occurrence of backslashes on Windows

odisseus:
# Checklist

details.

Issue number: #7258 

# Description of this change

### DIFF
--- a/querysync/java/com/google/idea/blaze/qsync/query/QuerySummary.java
+++ b/querysync/java/com/google/idea/blaze/qsync/query/QuerySummary.java
@@ -202,8 +202,8 @@ public abstract class QuerySummary {
     public Query.StoredLabel indexLabel(Label l) {
       return Query.StoredLabel.newBuilder()
         .setWorkspace(index(l.getWorkspaceName()))
-        .setBuildPackage(index(l.getPackage().toString()))
-        .setName(index(l.getName().toString()))
+        .setBuildPackage(index(l.buildPackage()))
+        .setName(index(l.name()))
         .build();
     }
 
--- a/shared/java/com/google/idea/blaze/common/Label.java
+++ b/shared/java/com/google/idea/blaze/common/Label.java
@@ -21,8 +21,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import java.nio.file.Path;
 import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 /**
  * Represents an absolute build target label.
@@ -42,6 +40,7 @@ public record Label(String workspace, String buildPackage, String name) {
 
   public static Label of(String label) {
     Preconditions.checkArgument(!label.isBlank(), "Empty label");
+    Preconditions.checkArgument(!label.contains("\\"), "Label contains backslashes: " + label);
     final var workspacePosition = label.startsWith("@") ? (label.startsWith("@@") ? 2 : 1) : 0;
     final var workspaceEnd = label.indexOf("//", workspacePosition);
     final var buildPackagePosition =  workspaceEnd + 2;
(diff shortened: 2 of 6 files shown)


shared/java/com/google/idea/blaze/common/Label.java
```diff
@@ -42,6 +40,7 @@ public record Label(String workspace, String buildPackage, String name) {
 
   public static Label of(String label) {
     Preconditions.checkArgument(!label.isBlank(), "Empty label");
+    Preconditions.checkArgument(!label.contains("\\"), "Label contains backslashes: " + label);
```
ujohnny, line 43: add a test to `LabelTest`?

shared/java/com/google/idea/blaze/common/Label.java
```diff
@@ -57,17 +56,22 @@ public static Label of(String label) {
...
     if (workspace.isEmpty()) {
-      return Label.of(String.format("//%s:%s", systemIndependentPackagePath, name));
+      return Label.of(String.format("//%s:%s", packageWithForwardSlashes, nameWithForwardSlashes));
     } else {
-      return Label.of(String.format("@@%s//%s:%s", workspace, systemIndependentPackagePath, name));
+      return Label.of(String.format("@@%s//%s:%s", workspace, packageWithForwardSlashes, nameWithForwardSlashes));
     }
   }
 
+  private static String withForwardSlashes(Path p) {
```
ujohnny, line 68: `com.intellij.openapi.util.io.FileUtilRt#toSystemIndependentName`

shared/java/com/google/idea/blaze/common/Label.java
```diff
@@ -57,17 +56,22 @@ public static Label of(String label) {
   }
 
   public static Label fromWorkspacePackageAndName(String workspace, Path packagePath, Path name) {
```
ujohnny, line 58: add test too, to see that it works with `WindowsPath`?

querysync/java/com/google/idea/blaze/qsync/query/QuerySummary.java
```diff
@@ -202,8 +202,8 @@ public int index(String s) {
     public Query.StoredLabel indexLabel(Label l) {
       return Query.StoredLabel.newBuilder()
         .setWorkspace(index(l.getWorkspaceName()))
-        .setBuildPackage(index(l.getPackage().toString()))
-        .setName(index(l.getName().toString()))
+        .setBuildPackage(index(l.buildPackage()))
```
ujohnny, line 205: I searched for `getPackage()` and subsequent call to `toString` that introduces backslashes and found https://github.com/bazelbuild/intellij/blob/a88c6db967cd173fa73aeb76d8c8c7bbd0f8be3d/shared/java/com/google/idea/blaze/common/TargetTree.java#L239

ujohnny (dismissed):

shared/java/com/google/idea/blaze/common/Label.java
odisseus: It turns out not so easy to refactor. This package is pretty low in the dependency hierarchy, and it doesn't depend on the IntelliJ API.

shared/java/com/google/idea/blaze/common/Label.java
odisseus: You mean the subclass of `java.nio.file.Path`? Unfortunately, it only exists on Windows JVMs. However, I have verified manually that my changes fix the issue on Windows.

querysync/java/com/google/idea/blaze/qsync/query/QuerySummary.java
odisseus: I'm not sure this has any adverse effects (FWIW `` passes on Windows), but I will fix this just in case.

shared/java/com/google/idea/blaze/common/Label.java
odisseus: Done

LeFrosch (changes requested):
Just tried to import our repository on windows with your changes and got this exception. Without your changes it was possible to import our repository (after removing excluded directories from the project view file).

image
[idea.log](https://github.com/user-attachments/files/18707439/idea.log)

shared/java/com/google/idea/blaze/common/Label.java
```diff
@@ -57,17 +56,22 @@ public static Label of(String label) {
...
-      return Label.of(String.format("@@%s//%s:%s", workspace, systemIndependentPackagePath, name));
+      return Label.of(String.format("@@%s//%s:%s", workspace, packageWithForwardSlashes, nameWithForwardSlashes));
     }
   }
 
+  private static String withForwardSlashes(Path p) {
+    StringBuilder res = new StringBuilder();
+    p.iterator().forEachRemaining(part -> res.append(part).append("/"));
+    // Remove the trailing slash
+    return res.substring(0, res.length() - 1);
```
LeFrosch, line 72: Blows up for empty path.

LeFrosch (changes requested):

shared/java/com/google/idea/blaze/common/Label.java
odisseus: I don't think this can ever happen (even `Path.of("")` returns an iterator that is not empty), but I've added a check just in case.

LeFrosch: Oh, yes looks like you are right.

LeFrosch (approved):

Issue #7255, opened 2025-01-23
stdlib is not recognizable when using CLion and toolchains_llvm with sysroots


auzhva:
### Description of the bug:

CLion doesn't see standard includes such as `<map>`, `<list>` and others if hermetic toolchain is used. All of them are marked red, though code is perfectly buildable.

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Reproducible on configurations like here https://github.com/bazel-contrib/toolchains_llvm/blob/master/tests/MODULE.bazel#L162-L176

### Which Intellij IDE are you using? Please provide the specific version.

2024.3.2

### What programming languages and tools are you using? Please provide specific versions.

C/C++

### What Bazel plugin version are you using?

Bazel 7.4.0, and Bazel 8

### Have you found anything relevant by searching the web?

Might be related to this — https://github.com/bazelbuild/intellij/issues/3817 as Android also uses sysroots. But reproducible without Android.

LeFrosch:
Should be fixed by #7274. If you still experience this issue, feel free to reopen the ticket.

PR #7274, opened 2025-02-04
CLion support for custom sysroot

LeFrosch:

Issue number: #7255

# Description of this change

Add builtin include directories from the toolchain as system includes to the switch builder. If the toolchain does not use custom sysroot CLion is able to find all include directories during compiler info collection. However, if a custom sysroot is provided this might not allways be the case.


LeFrosch:
Also provides a workaround for: https://github.com/bazelbuild/bazel/issues/25145

aspect/intellij_info_impl.bzl
```diff
@@ -1374,6 +1374,7 @@ def make_intellij_info_aspect(aspect_impl, semantics, **kwargs):
         ),
         "_java_runtime": attr.label(
             default = "@bazel_tools//tools/jdk:current_java_runtime",
+            cfg = "exec",
```
ujohnny, line 1377: just make sure this does not exist in the master branch

ujohnny: ie avoid duplicate commits, this just looks suspicious

ujohnny (approved):

PR #7213, opened 2025-01-07
fix(qsync): External modules navigation in starlark files

sellophane:
# Checklist

- [ ] I have filed an issue about this change and discussed potential changes with the maintainers.
- [ ] I have received the approval from the maintainers to make this change.
- [ ] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: `<please reference the issue number or url here>`

# Description of this change

PR #7294, opened 2025-02-10
Followup: External modules navigation in starlark files

LeFrosch:
followup for https://github.com/bazelbuild/intellij/pull/7213

By populate the blazeInfo and externalWorkspaceData fields in
QuerySyncProjectData similar to how AspectProjectData does, the same
code that handles external module navigation for async also works for
qsync.


sellophane (approved):

LeFrosch:
rebased master

PR #7300, opened 2025-02-11
Use toolchains_aspects to propagate aspects to toolchains

mai93:
with bazel @ head `_cc_toolchain` attribute is removed and we should modify the aspect to propagate to the toolchain via the toolchain type not the attribute name.

The feature for aspect to toolchains propagation is only available at Bazel 8, so we would need to version the aspect based on the user Bazel version if we want to merge this.

### DIFF
--- a/aspect/intellij_info.bzl
+++ b/aspect/intellij_info.bzl
@@ -4,12 +4,17 @@ load(
     ":intellij_info_impl.bzl",
     "intellij_info_aspect_impl",
     "make_intellij_info_aspect",
+    "is_valid_aspect_target",
 )
 
 EXTRA_DEPS = [
     "embed",  # From go rules (bazel only)
     "_cc_toolchain",  # From rules_cc (bazel only)
     "_kt_toolchain",  # From rules_kotlin (bazel only)
+]
+
+TOOLCHAIN_TYPE_DEPS = [
+    "@@bazel_tools//tools/cpp:toolchain_type",  # For rules_cc
 ]
 
 def tool_label(tool_name):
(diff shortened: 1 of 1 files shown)


odisseus (commented):
Looks good

odisseus:
Looks like these changes break the project import:
```
	File ".../testdata/.ijwb/aspect/intellij_info_bundled.bzl", line 89, column 49, in <toplevel>
		intellij_info_aspect = make_intellij_info_aspect(_aspect_impl, semantics, toolchains_aspects = TOOLCHAIN_TYPE_DEPS,)
	File ".../testdata/.ijwb/aspect/intellij_info_impl_bundled.bzl", line 1384, column 18, in make_intellij_info_aspect
		return aspect(
Error in aspect: aspect() got unexpected keyword argument 'toolchains_aspects'
ERROR: initialization of module '.ijwb/aspect/intellij_info_bundled.bzl' failed
ERROR: Analysis of aspects '[//.ijwb/aspect:intellij_info_bundled.bzl%intellij_info_aspect] with parameters {} on //:foo' failed; build aborted: initialization of module '.ijwb/aspect/intellij_info_bundled.bzl' failed
```

mai93:
> Looks like these changes break the project import:

yes, as I mentioned in the description. we still need to add some versioning to the aspect so that it only uses `toolchains_aspects` with bazel > 8. Maybe someone can help with that?

LeFrosch:
I can take a more extensive look on Friday. The aspect templating is sadly in no state that would make this easy. Is this urgent?

mai93:
we need to reenable integration tests with Bazel@head

LeFrosch:
After taking a more extensive look, I am not sure if we can support this (especially in testing) without major changes to the aspect handling.

PR #7304, opened 2025-02-11
Open build toolwindow when the sync ends with an error

LeFrosch:
To avoid silent errors, open the build toolwindow if the sync exists with an error or a warning. Otherwise, it could be confusing for the user if they don't notice that the sync has failed.


odisseus (commented):
Looks good

base/src/com/google/idea/blaze/base/buildview/BuildViewScope.kt
```diff
@@ -82,6 +84,15 @@ class BuildViewScope(project: Project, private val title: String) : BlazeScope {
       else -> progress.finish()
     }
 
+    if (ctx.hasErrors() || ctx.hasWarnings()) {
+      // show the toolwindow of there was an error or warning during the sync
+      invokeLater {
```
ujohnny, line 89:
```
pluginProjectScope(project).launch(Dispatcher.EDT) {
  BuildContentManager.getInstance(project).getOrCreateToolWindow().activate(null)
}

```

ujohnny (approved):

PR #7308, opened 2025-02-11
Fix CLion integration tests for EAP

LeFrosch:
Update BUILD.clion251 to reflect layout changes in the latest EAP.

### DIFF
--- a/intellij_platform_sdk/BUILD.clion251
+++ b/intellij_platform_sdk/BUILD.clion251
@@ -54,6 +54,7 @@ java_import(
     jars = glob([
         "plugins/c-plugin/lib/*.jar",
         "plugins/cidr-base-plugin/lib/*.jar",
+        "plugins/cidr-base-plugin/lib/modules/*.jar",
         "plugins/cidr-debugger-plugin/lib/*.jar",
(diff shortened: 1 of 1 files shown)


LeFrosch:
For the record, to debug plugin loading issues for integration tests, set a break point in: 
```
community/platform/core-impl/src/com/intellij/ide/plugins/PluginManagerCore.kt:628(registerLoadingError)
```

intellij_platform_sdk/BUILD.clion251
```diff
@@ -54,6 +54,7 @@ java_import(
     jars = glob([
         "plugins/c-plugin/lib/*.jar",
         "plugins/cidr-base-plugin/lib/*.jar",
+        "plugins/cidr-base-plugin/lib/modules/*.jar",
```
ujohnny, line 57: @DanielBelow fyi, probably after your recent changes we'll have to adjust something here

ujohnny (approved):

intellij_platform_sdk/BUILD.clion251
DanielBelow: Was it maybe caused by the plugin-appender issue?

PR #7198, opened 2024-12-18
handling for debugging python test targets with transitions

andponlin-canva:
# Checklist

- [x] I have filed an issue about this change and discussed potential changes with the maintainers.
- [ ] I have received the approval from the maintainers to make this change.
- [x] This is not a stylistic, refactoring, or cleanup change.

Please note that the maintainers will not be reviewing this change until all checkboxes are ticked. See 
the [Contributions](https://github.com/bazelbuild/intellij#contributions) section in the README for more 
details.

# Discussion thread for this change

Issue number: [7196](https://github.com/bazelbuild/intellij/issues/7196)

# Description of this change

- The change will recognize Kinds with a `_transition_` prefix as being the Kind with with `_transition_` prefix stripped.
- The selection of the Target to map to from a source file will avoid, where possible, any with a `_` prefix.
- The Python templated aspect logic will allow the `rules_python` `PyInfo` provider to be considered with Bazel < 8. I am unsure of all the wider ramifications of this so it may require some discussion. I think internal Bazel rules were dropped on Bazel 5 so perhaps it is a reasonable assumption that the user would be using `rules_python` if they are using Python at all?

PR #7314, opened 2025-02-12
Return sorted sets (follow-up to #7198)

odisseus:
# Checklist

details.

Issue number: #7198 

# Description of this change

Immutability of collections that we return is an important part of the contract.

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/testmap/ProjectSourceToTargetFinder.java
+++ b/base/src/com/google/idea/blaze/base/run/testmap/ProjectSourceToTargetFinder.java
@@ -15,11 +15,11 @@
  */
 package com.google.idea.blaze.base.run.testmap;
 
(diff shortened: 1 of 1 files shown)


LeFrosch:
If I remember correctly, it was a set before. Do we introduce duplicates by moving from a set to a list? Maybe keep it a set if we don't want duplicates.

But overall, I think it is a good idea going back to immutable here.

odisseus:
The chance from Set to List was introduced in #7198. I think the idea was that the collection of targets should be ordered, and that well-known kinds should have precedence.

LeFrosch (approved):
Makes sense. Let's keep it as a list for now.

LeFrosch:
Not sure, in this case we might be able to drop the sort call too?

LeFrosch (approved):

PR #7317, opened 2025-02-13
[AOSP-pick] Move BuildResult out of legacy sync

sellophane:
Cherry pick AOSP commit [175db03df86fe2eeec561943d5ea57935427fca5](https://cs.android.com/android-studio/platform/tools/adt/idea/+/175db03df86fe2eeec561943d5ea57935427fca5).

Bug: n/a
Test: n/a
Change-Id: Ib03643073f9c71c1af68d024f3899c2acea8ce5c

AOSP: 175db03df86fe2eeec561943d5ea57935427fca5


LeFrosch (approved):
There are some inconsistencies caused by formatting. In this case it is not that big of deal because here they are only imports, but overall I would suggest applying the patch as is without formatting.

PR #7342, opened 2025-02-20
[AOSP-pick] Replace output group name filter

sellophane:
Cherry pick AOSP commit [c47d961501bca50da2e2c5bfd4129f214cc69a46](https://cs.android.com/android-studio/platform/tools/adt/idea/+/c47d961501bca50da2e2c5bfd4129f214cc69a46).

with an explicit group name where possible. Query sync does not require
a more expensive output group name filtering applied to each artifact.

Separate legacy and query sync use cases and simplify support for the
query sync one.

Bug: 327638725
Test: n/a (existing)
Change-Id: If2b465a63ea3f7c65d425ad7b1ebca645f3e0556

AOSP: c47d961501bca50da2e2c5bfd4129f214cc69a46

### DIFF
--- a/base/src/com/google/idea/blaze/base/qsync/BazelAppInspectorBuilder.java
+++ b/base/src/com/google/idea/blaze/base/qsync/BazelAppInspectorBuilder.java
@@ -18,6 +18,7 @@ package com.google.idea.blaze.base.qsync;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableMap;
 import com.google.idea.blaze.base.bazel.BazelExitCodeException;
 import com.google.idea.blaze.base.bazel.BuildSystem;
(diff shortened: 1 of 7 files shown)


base/src/com/google/idea/blaze/base/qsync/BazelAppInspectorBuilder.java
```diff
@@ -15,8 +15,6 @@
  */
 package com.google.idea.blaze.base.qsync;
 
-import static com.google.common.collect.ImmutableList.toImmutableList;
-
```
LeFrosch, line 19: To reduce merge conflicts I would advocate for prioritizing AOSP compatibility rather than cleaning up unused imports for files which are not actively maintained by us.

LeFrosch (changes requested):
The changes to the import from [here](https://cs.android.com/android-studio/platform/tools/adt/idea/+/c47d961501bca50da2e2c5bfd4129f214cc69a46:aswb/base/src/com/google/idea/blaze/base/qsync/BazelRenderJarBuilder.java;dlc=f56c98a1f76117418af4c0539d731ba0c99ebe60) were also not picked.

LeFrosch (approved):

PR #7351, opened 2025-02-21
IJWB headless tests

LeFrosch:
Followup for #6909

Move CLWB headless test (former integration test) infrastructure to testing and reuse it to create IJWB headless tests.


sellophane (approved):

BUILD
```diff
@@ -72,29 +73,11 @@ test_suite(
...
+        "//clwb:headless_tests",
         "//cpp:unit_tests",
         "//dart:unit_tests",
         "//python:unit_tests",
         "//skylark:unit_tests",
     ],
 )
 # END-EXTERNAL
-
-# ASwB tests, run with an Android Studio plugin SDK
```
odisseus, line 84: Why?

odisseus (commented):
Let's add the new tests now and fix the Windows issues later.

BUILD
LeFrosch: Because we can't even build aswb anymore. No use to keep tests around in my opinion.

odisseus (approved):

susinmotion:
It looks like this PR removes the aswb tests. Can you explain why that is, and what we should do instead? context: trying to release the Bazel plugin.

odisseus:
ASWB was the Bazel plugin for Android Studio. IIRC it had been discontinued a few years ago, long before this PR was merged. Unless you are trying to release that particular plugin, these tests and their removal should be of no relevance to you.

susinmotion:
I am trying to release that particular plugin!

ujohnny:
@susinmotion there could be more sense to make it from `google` branch rather than `master`, I think last time it was build from master somewhere in 2024 if not earlier.

susinmotion:
I suspect we are building from a branch, but maybe updated something on our end?... I'm filling in for the owner of the plugin so I'm lacking some context. Thanks for the breadcrumb, though, I'll look more tomorrow.

ujohnny:
@susinmotion but on the Google side plugin was moved to AOSP and has to be built there https://cs.android.com/android-studio/platform/tools/adt/idea/+/mirror-goog-studio-main:aswb/

PR #7357, opened 2025-02-25
CLion qsync improvements for Mac

LeFrosch:
1. Make cc compiler path exec root relative. The path was workspace relative before which is not correct.
2. Also make external include paths exec root relative.

Also added test for external include paths. To run qsync tests on mac we are still missing xcode integration.

### DIFF
--- a/clwb/tests/integrationtests/com/google/idea/blaze/clwb/base/ClwbIntegrationTestCase.java
+++ b/clwb/tests/integrationtests/com/google/idea/blaze/clwb/base/ClwbIntegrationTestCase.java
@@ -273,27 +273,42 @@ public abstract class ClwbIntegrationTestCase extends HeavyPlatformTestCase {
     return false;
   }
 
-  protected boolean enableAnalysisFor(VirtualFile file) throws ExecutionException {
+  protected SyncOutput enableAnalysisFor(VirtualFile file) throws ExecutionException {
+    final var context = BlazeContext.create();
+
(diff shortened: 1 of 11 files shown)


sellophane (approved):

clwb/tests/integrationtests/com/google/idea/blaze/clwb/base/ClwbIntegrationTestCase.java
```diff
@@ -273,27 +273,42 @@ protected boolean runQuerySync() {
...
+    LOG.info(String.format("PROJECT BUILD LOG:%n%s", output.collectLog()));
+
     try {
-      return future.get();
+      future.get();
+    } catch (ExecutionException e) {
+      LOG.error("enable analysis failed", e);
     } catch (InterruptedException e) {
-      fail("enable analysis was interrupted");
+      LOG.error("enable analysis was interrupted", e);
```
odisseus, line 308: Why are the exceptions swallowed here?

odisseus (changes requested):

clwb/tests/integrationtests/com/google/idea/blaze/clwb/base/ClwbIntegrationTestCase.java
LeFrosch: They are not, log error throws in tests and provides more details then `fail`.

odisseus (approved):

PR #7376, opened 2025-02-28
Followup: Use toolchains_aspects to propagate aspects to toolchains

LeFrosch:
Follow up for: #7300

### DIFF
--- a/aspect/testing/tests/src/com/google/idea/blaze/aspect/integration/BUILD
+++ b/aspect/testing/tests/src/com/google/idea/blaze/aspect/integration/BUILD
@@ -1,4 +1,6 @@
-load("//testing:test_defs.bzl", "bazel_integration_tests")
+load("@bazel_binaries//:defs.bzl", "bazel_binaries")
+load("@rules_bazel_integration_test//bazel_integration_test:defs.bzl", "bazel_integration_test")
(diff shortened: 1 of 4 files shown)


aspect/testing/tests/src/com/google/idea/blaze/aspect/integration/BUILD
```diff
@@ -15,8 +17,11 @@ java_binary(
     ],
 )
 
-bazel_integration_tests(
+bazel_integration_test(
```
mai93, line 20: did this disable the tests against older versions of bazel? I do not see it in the test result https://buildkite.com/bazel/intellij-plugin-aspect/builds/24766#01954c26-b10e-4a63-8d88-38d1b7f04258

LeFrosch: Yes it does. Because the integration tests do not support templating and therefore it is not possible to target multiple bazel versions with these changes.

mai93 (approved):

ujohnny:
Looks good and probably we'll remove the test since we have new integration tests that do the job during bazel 9 migration, because the test is still using workspace and probably does not make any sense to update it to bzlmod.

ujohnny (approved):

PR #7382, opened 2025-02-28
[AOSP-pick] Parameterize aspect at instantiation

sellophane:
Cherry pick AOSP commit [2e116b56c18f3c50f832e759537321ed0ba6fb34](https://cs.android.com/android-studio/platform/tools/adt/idea/+/2e116b56c18f3c50f832e759537321ed0ba6fb34).

This is to unify the test and production flows and make manual testing
from the command line tolerable.

It should also make the aspect configuration visible in CitC snapshots
and aspect builds easily reproducible in users' workspaces.

Bug: 327638725
Test: BazelDependencyBuilderTest
Change-Id: I4cf4b408097cecc7d324310d98c18ce36900468e

AOSP: 2e116b56c18f3c50f832e759537321ed0ba6fb34

### DIFF
--- a/aspect/build_dependencies.bzl
+++ b/aspect/build_dependencies.bzl
@@ -355,15 +355,11 @@ def declares_aar_import(ctx):
     """
     return ctx.rule.kind == "aar_import" and hasattr(ctx.rule.attr, "aar")
 
-def _collect_dependencies_impl(target, ctx):
+def _collect_dependencies_impl(target, ctx, params):
     return _collect_dependencies_core_impl(
         target,
         ctx,
-        ctx.attr.include,
-        ctx.attr.exclude,
-        ctx.attr.always_build_rules,
-        ctx.attr.generate_aidl_classes,
-        ctx.attr.use_generated_srcjars,
+        params,
         test_mode = False,
     )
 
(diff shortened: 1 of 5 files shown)


aspect/build_dependencies.bzl
```diff
@@ -409,19 +409,21 @@ def _target_within_project_scope(label, include, exclude):
...
-        for inc in include.split(","):
-            # This is not a valid label, but can be passed to aspect when `directories: .` is set in the projectview
-            if (inc == "//"):
-                result = True
-                break
-
-            inc = Label(inc)
-            if _get_repo_name(inc) == repo and _package_prefix_match(package, inc.package):
-                result = True
-                break
```
LeFrosch, line 421: I personally prefer this implementation. But of course I'm biased here. I think we should either keep the original implementation or use the AOSP implementation.

LeFrosch (changes requested):

aspect/build_dependencies.bzl
odisseus: I think the AOSP implementation is better, because it doesn't try to use a comma-delimited string as a makeshift collection.

base/src/com/google/idea/blaze/base/qsync/BazelDependencyBuilder.java
```diff
@@ -162,91 +173,134 @@ public OutputInfo build(
     Optional<BuildDepsStats.Builder> buildDepsStatsBuilder =
         BuildDepsStatsScope.fromContext(context);
     buildDepsStatsBuilder.ifPresent(stats -> stats.setBlazeBinaryType(invoker.getType()));
-
-    String includes =
-        projectDefinition.projectIncludes().stream()
-            .map(path -> "//" + path)
-            .collect(joining(","));
```
LeFrosch, line 169: @odisseus you are right, they are dropping the comma separated collection. In this case I would suggest to drop all my changes and use the AOSP implementation.

LeFrosch (approved):

PR #7399, opened 2025-03-04
Fix make variable expansion

LeFrosch:
The custom implementation form make variable expansion breaks in some edge cases. For example empty make variables. Therefore, this PR replaces the custom implementation with the [one](https://github.com/bazelbuild/bazel/blob/6f7faa659e5eb3e56c8a6274ebcb86884703d603/src/main/starlark/builtins_bzl/common/cc/cc_helper.bzl#L910) used by bazel.

### DIFF
--- a/aspect/make_variables.bzl
+++ b/aspect/make_variables.bzl
@@ -1,98 +1,194 @@
-"""Utility functions to expand make variables."""
+"""Utility functions to expand make variables. Implementation taken from cc_helper. """
 
-def _is_valid_make_var(varname):
-    """Check if the make variable name seems valid."""
-    if len(varname) == 0:
-        return False
-
-    # According to gnu make, any chars not whitespace, ':', '#', '=' are valid.
-    invalid_chars = ":#= \t\n\r"
-    for n in range(0, len(invalid_chars)):
-        if invalid_chars[n] in varname:
-            return False
-    return True
-
-def expand_make_variables(attr_name, expression, ctx, additional_subs = {}):
-    """Substitutes make variables defined in $() syntax.
-
-    Because ctx.expand_make_variables is deprecated, we need to be able to do the
-    substitution without relying on it.
-    Before the aspect is processed, the build system already detects most/all of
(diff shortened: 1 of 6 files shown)


aspect/make_variables.bzl
```diff
@@ -1,98 +1,194 @@
...
-    # make variables from ctx.var can be overridden
-    substitutions.update(additional_subs)
-
-    # skylark does not support while. This is the maximum iteration count this
-    # loop will need, but it will exit early if possible.
-    for _n in range(0, len(expression)):
-        if current_offset >= len(expression):
+"""Utility functions to expand make variables. Implementation taken from cc_helper. """
+
+# Source: https://github.com/bazelbuild/bazel/blob/6f7faa659e5eb3e56c8a6274ebcb86884703d603/src/main/starlark/builtins_bzl/common/cc/cc_helper.bzl#L910
```
ujohnny, line 3: license?

LeFrosch: The bazel implementation is Apache License, Version 2.0. Do we need to mention this here?

ujohnny: yep the license should be preserved

ujohnny (approved):

aspect/make_variables.bzl
```diff
@@ -1,98 +1,209 @@
...
+        unbounded_recursion = False
+        break
+
+    if unbounded_recursion:
+        fail("potentially unbounded recursion during expansion of {}".format(exp))
+    return exp
+
+def _lookup_var(ctx, additional_vars, var):
+    expanded_make_var_ctx = ctx.var.get(var)
+    expanded_make_var_additional = additional_vars.get(var)
```
LeFrosch, line 91: So we'd need to check ctx.rule.var here too?

LeFrosch: Thanks for reporting this, I'll look into this.

LeFrosch: Addressed by #7455

PR #7435, opened 2025-03-07
update python example to more recent version of rules_python

andponlin-canva:
# Checklist

details.

Issue number: N/A

# Description of this change

This change will update the Python `simple_code_generator` so that it works with more recent versions of `rules_python`.

### DIFF
--- a/examples/python/simple_code_generator/MODULE.bazel
+++ b/examples/python/simple_code_generator/MODULE.bazel
@@ -3,7 +3,7 @@ module(
     version = "1.0.0",
 )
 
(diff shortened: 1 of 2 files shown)


andponlin-canva:
The core Bazel system has an internal `PyInfo` and `rules_python` has its own `PyInfo`; the two being different. In Bazel 7 it was possible to provide both providers but it seems in Bazel 8 this is disallowed so in this example code I am making the change to use only the one from `rules_python`.

LeFrosch (approved):
Thanks for the clarification. Looks good to me.

PR #7440, opened 2025-03-07
Make ProjectOpenProcessors no strong info holder by default

LeFrosch:
In order to not automatically import every project as a Bazel project even if there are other project models available, the ProjectOpenProcessors cannot be strong info holders by default. 

The old behavior can be restored by setting the `bazel.project.auto.open` registry key to true.

Removes the following obsolete registry keys: 
- `bazel.project.prefer.google.plugin`
- `bazel.project.auto.open.if.present`

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -413,7 +413,7 @@
                  description="Determines whether AutoImportProjectOpenProcessor is a strong info holders."
                  key="bazel.project.auto.open"/>
     <registryKey defaultValue="true"
-                 description="Determines whether BlazeProjectOpenProcessor is a strong info holders."
+                 description="Determines whether BlazeProjectOpenProcessor is a strong info holder."
                  key="bazel.project.auto.open.if.present"/>
     <registryKey defaultValue="false"
                  description="The Bazel Plugin sometimes uses the --query_file option to pass the query via a file. By default, the file is deleted immediately after the query execution is finished. Set this flag to true to avoid file deletion"
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -409,12 +409,9 @@
     <registryKey defaultValue="false"
                  description="Scan for non-root source folders in source jars. (requires re-sync)"
                  key="bazel.sync.detect.source.roots"/>
-    <registryKey defaultValue="true"
-                 description="By default Bazel plugin takes control and opens project if .ijwb folder is present. This property can be used to disable this behavior to allow to open .idea or other project models if they are exists."
-                 key="bazel.project.auto.open.if.present"/>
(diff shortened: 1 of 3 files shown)


base/src/com/google/idea/blaze/base/project/BlazeProjectOpenProcessor.java
```diff
@@ -73,7 +73,7 @@ public boolean canOpenProject(VirtualFile file) {
 
   @Override
   public boolean isStrongProjectInfoHolder() {
-    return Registry.is("bazel.project.auto.open.if.present", false);
+    return Registry.is("bazel.project.auto.open");
```
ujohnny, line 76: I'm unsure if the change here from true to false is expected

LeFrosch: Are you suggesting to keep the extra registry key for this ProjectOpenProcessor?

ujohnny: yes I believe we should keep the original key here as is, but just drop 2nd arg

ujohnny (changes requested):

base/src/META-INF/blaze-base.xml
```diff
@@ -409,12 +409,12 @@
     <registryKey defaultValue="false"
                  description="Scan for non-root source folders in source jars. (requires re-sync)"
                  key="bazel.sync.detect.source.roots"/>
+    <registryKey defaultValue="false"
+                 description="Determines whether AutoImportProjectOpenProcessor is a strong info holders."
```
odisseus, line 413:
```suggestion
                 description="Determines whether AutoImportProjectOpenProcessor is a strong info holder."
```

base/src/META-INF/blaze-base.xml
```diff
@@ -409,12 +409,12 @@
...
+                 key="bazel.project.auto.open"/>
     <registryKey defaultValue="true"
-                 description="By default Bazel plugin takes control and opens project if .ijwb folder is present. This property can be used to disable this behavior to allow to open .idea or other project models if they are exists."
+                 description="Determines whether BlazeProjectOpenProcessor is a strong info holders."
```
odisseus, line 416:
```suggestion
                 description="Determines whether BlazeProjectOpenProcessor is a strong info holder."
```

ujohnny (approved):

ujohnny:
@dkashyn-sfdc we're going to change the behavior of the key you've introduced, please be aware of this and adjust VM options on your side.

dkashyn-sfdc:
@LeFrosch @ujohnny this will regress for project where `.idea` folder present which will be "strong" by default. Those settings were added to not end up in the situation "Bazel project import not happening and not suggested if `.idea` is present.

PR #7448, opened 2025-03-10
[AOSP-pick] Clean up usages of @com_google_protobuf//:protobuf_java

LeFrosch:
Cherry pick AOSP commit [1f25b080153fa75504b01f8bfa9f19389f705590](https://cs.android.com/android-studio/platform/tools/adt/idea/+/1f25b080153fa75504b01f8bfa9f19389f705590).

This version of protobuf should only be used for the native implementation
and for common proto definitions.

Test: existing
Bug: none
Change-Id: Id94e6a5b74ef52d5c389e448506e13b2e1bdb002

AOSP: 1f25b080153fa75504b01f8bfa9f19389f705590

### DIFF
--- a/querysync/javatests/com/google/idea/blaze/qsync/deps/BUILD
+++ b/querysync/javatests/com/google/idea/blaze/qsync/deps/BUILD
@@ -55,7 +55,7 @@ java_test(
         "//shared",
         "//third_party/java/junit",
         "//third_party/java/truth",
-        "@protobuf//:protobuf_java",
         "@com_google_guava_guava//jar",
+        "@protobuf//:protobuf_java",
     ],
 )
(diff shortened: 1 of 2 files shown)


sellophane (approved):

sellophane:
Are we here just moving deps one line lower?

LeFrosch:
In the AOSP commit they are switching from 
`@com_google_protobuf//:protobuf_java`
to
`@maven//:com.google.protobuf.protobuf-java`
but they both map to
`@protobuf//:protobuf_java`
see
https://github.com/LeFrosch/intellij-aosp-merge/blob/f1d36a536bd70fadfa0f0b65400ed4e1c4c309aa/aosp/_deaosp.py#L62-L64

sellophane:
> In the AOSP commit they are switching from `@com_google_protobuf//:protobuf_java` to `@maven//:com.google.protobuf.protobuf-java` but they both map to `@protobuf//:protobuf_java` see https://github.com/LeFrosch/intellij-aosp-merge/blob/f1d36a536bd70fadfa0f0b65400ed4e1c4c309aa/aosp/_deaosp.py#L62-L64

Yes, I've check the original commit already, I was just wondering if this change is a good idea, if it doesn't provide any value. But lets have it for consistency sake

PR #7455, opened 2025-03-10
CLion Bazel 9 compatability fixes

LeFrosch:
Fixes the following incompatibilities: 
- Not all make variables are stored in `ctx.var` and we also have to check `ctx.rule.var` in Bazel 9
- Only use the `toolchains_aspects` starting with Bazel 9 or else we get two toolchains per target.

Also adds the ability to run headless tests against `last_green` Bazel version. They are not part of the regular test suites but can be run manually or using the following test filter: `--test_tag_filters=bit_bazel_last_green` 

I propose to run the headless tests against the latest Bazel version as part of CI under soft failure to catch possible issues with future Bazel versions as early as possible.


aspect/intellij_info.bzl
```diff
@@ -95,7 +95,7 @@ def _aspect_impl(target, ctx):
 ##intellij_info_aspect = make_intellij_info_aspect(
 ##    _aspect_impl,
 ##    semantics,
-## #if( $bazel8OrAbove == "true" )
+## #if( $bazel9OrAbove == "true" )
 ##    toolchains_aspects = TOOLCHAIN_TYPE_DEPS,
 ## #end
```
LeFrosch, line 100: @mai93 is this correct?

aspect/make_variables.bzl
```diff
@@ -87,12 +87,19 @@ def _expand_nested_variable(ctx, additional_vars, exp, execpath = True, targets
     return exp
 
 def _lookup_var(ctx, additional_vars, var):
-    expanded_make_var_ctx = ctx.var.get(var)
-    expanded_make_var_additional = additional_vars.get(var)
-    if expanded_make_var_additional != None:
-        return expanded_make_var_additional
-    if expanded_make_var_ctx != None:
-        return expanded_make_var_ctx
+    expanded_make_var = additional_vars.get(var)
```
SergejSalnikov, line 90: What do you think about the following pattern:
```
expanded_make_var = a()
if expanded_make_var == None:
  expanded_make_var = b()
if expanded_make_var == None:
  expanded_make_var = c()

if expanded_make_var == None:
  fail()

return expanded_make_var
```

base/src/com/google/idea/blaze/base/sync/aspects/storage/AspectStorageService.kt
SergejSalnikov, line 1: I think you also need to call `AspectStorageService.of(project).prepare(context)` in `FastBuildServiceImpl.buildDeployJar`.

LeFrosch: Yes, thanks for pointing this out.

aspect/make_variables.bzl
LeFrosch: Works too, but at this point I feel like we are discussing personal preference.

SergejSalnikov: It is. just wanted to bring this up, as the code becomes shorter and having less exit points that are closely clustered.

sellophane (approved):

base/src/com/google/idea/blaze/base/sync/aspects/storage/AspectStorageService.kt
```diff
@@ -91,24 +95,46 @@ class AspectStorageService(private val project: Project, private val scope: Coro
...
 
-    val file = directory.resolve(file)
-    if (!Files.exists(file)) return Optional.empty()
+    val relativePath = directory.resolve(file)
+    if (!Files.exists(relativePath)) return Optional.empty()
 
-    val path = Path.of(settings.workspaceRoot).relativize(file)
-    return Optional.of(Label.fromWorkspacePackageAndName("", path.parent, path.fileName))
+    val absolutePath = Path.of(settings.workspaceRoot).relativize(relativePath)
+    return Optional.of(Label.fromWorkspacePackageAndName("", absolutPath.parent, absolutPath.fileName))
```
sellophane, line 137:
```suggestion
    return Optional.of(Label.fromWorkspacePackageAndName("", absolutePath.parent, absolutePath.fileName))
```

PR #7472, opened 2025-03-14
[AOSP-pick] Clean up and fix java-info file generation

sellophane:
Cherry pick AOSP commit [c760d03d456564b5f0dbaf48f74d121e792ff2e4](https://cs.android.com/android-studio/platform/tools/adt/idea/+/c760d03d456564b5f0dbaf48f74d121e792ff2e4).

1. Make `_collect_own_java_artifacts` return singleton depsets insets of
   lists of files or depsets later converted to depsets. This is for
   uniformity.

2. Make `_collect_own_and_dependency_java_artifacts` return a struct
   instead of a tuple so that it can return None.

3. Do not follow dependencies of targets that are not Java like (i.e.
   Java, Android or proto for Java/Android). This cuts off huge
   dependency chains of cc-like targets. Info files for cc targets are
   collected separately and non-recursively.

Bug: n/a
Test: n/a (existing ensures validity of the aspect in general)
Change-Id: If18bd8c9be8864845efca890c88081472d160ec6

AOSP: c760d03d456564b5f0dbaf48f74d121e792ff2e4


aspect/build_dependencies.bzl
```diff
@@ -695,34 +711,25 @@ def _collect_own_and_dependency_java_artifacts(
...
-            gen_srcs = gen_srcs,
-            srcs = own_files.srcs,
-            srcjars = own_files.srcjars,
-            android_resources_package = own_files.android_resources_package,
-        )
-    elif target_is_within_project_scope:
-        target_to_artifacts[str(target.label)] = _target_to_artifact_entry()
 
-    own_and_transitive_jar_depsets = list(own_files.jar_depsets)  # Copy to prevent changes to own_jar_depsets.
+    # Flattening is fine here as these are files from a single target (maybe some are re-exported from a few of its depende3ncies).
```
odisseus, line 719:
```suggestion
    # Flattening is fine here as these are files from a single target (maybe some are re-exported from a few of its dependencies).
```

odisseus (approved):

PR #7476, opened 2025-03-14
[AOSP-pick] Merge Local Blaze and Bazel Invokers

sellophane:
Cherry pick AOSP commit [f4b863f5a6e1b50b3e505cc7d52c521b16a34dec](https://cs.android.com/android-studio/platform/tools/adt/idea/+/f4b863f5a6e1b50b3e505cc7d52c521b16a34dec).

The Blaze and Bazel invokers are pretty much the same, except the binary type and the binary path. Merge them into a single `LocalInvoker` to avoid duplicate code as part of the cleanup.

Prior to cleanup, they were two different classes, but shared the same command runner and the build result helper.

Bug:374906681

Test: n/a
Change-Id: I7a9a570c48191ba6b05e1b5d227fc8a97ebe9398

AOSP: f4b863f5a6e1b50b3e505cc7d52c521b16a34dec

### DIFF
--- a/base/src/com/google/idea/blaze/base/bazel/LocalInvoker.java
+++ b/base/src/com/google/idea/blaze/base/bazel/LocalInvoker.java
@@ -1,0 +1,275 @@
+/*
+ * Copyright 2025 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.idea.blaze.base.bazel;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.io.Closer;
+import com.google.idea.blaze.base.async.process.ExternalTask;
+import com.google.idea.blaze.base.async.process.LineProcessingOutputStream;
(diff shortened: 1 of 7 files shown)


base/src/com/google/idea/blaze/base/bazel/LocalInvoker.java
```diff
@@ -0,0 +1,275 @@
...
+    return getBepStream(outputFile);
+  }
+
+  @Override
+  public InputStream invokeQuery(BlazeCommand.Builder blazeCommandBuilder) {
+    try {
+      performGuardCheck(project, blazeContext);
+    }
+    catch (ExecutionDeniedException e) {
+      logger.error(e);
```
odisseus, line 127: This looks like error swallowing to me.

base/src/com/google/idea/blaze/base/bazel/LocalInvoker.java
```diff
@@ -0,0 +1,275 @@
...
+          .stderr(
+            LineProcessingOutputStream.of(
+              line -> {
+                line = rootReplacement.apply(line);
+                // errors are expected, so limit logging to info level
+                if (isUnitTestMode) {
+                  // This is essential output in bazel-in-bazel tests if they fail.
+                  System.out.println(line.stripTrailing());
+                }
+                Logger.getInstance(this.getClass()).info(line.stripTrailing());
```
odisseus, line 155: The value of `line.stripTrailing()` can be stored in a variable to avoid repeated calls.

base/src/com/google/idea/blaze/base/bazel/LocalInvoker.java
LeFrosch: Yes, not perfect but `logger.error` should be fine. Also, not worth deviating from AOSP for this in my opinion.

base/src/com/google/idea/blaze/base/bazel/LocalInvoker.java
LeFrosch: Let's not deviating from AOSP for this either.

LeFrosch (approved):

Issue #7470, opened 2025-03-14
build fail for bazel 6.0


codernet:
### Description of the bug:

There was a change for generating intellij_info_bundled.bzl file.
https://github.com/bazelbuild/intellij/commit/4e6b7c0aff826f0ba52c063d9b69eaa4af55fda5#diff-f884456223ec33c5535ce1f43d915af337eaa3cba619ec8d56e7eb6506cba2acL101-L102

Using bazel version 6.0 will have the following error:

ERROR: Traceback (most recent call last):
	File "<path>/.ijwb/aspects/intellij_info_bundled.bzl", line 71, column 49, in <toplevel>
		intellij_info_aspect = make_intellij_info_aspect(_aspect_impl, semantics)
	File "<path>/.ijwb/aspects/intellij_info_impl_bundled.bzl", line 1361, column 33, in make_intellij_info_aspect
		default = tool_label("PackageParser_deploy.jar"),
	File "<path>/.ijwb/aspects/intellij_info_bundled.bzl", line 17, column 17, in tool_label
		return Label("tools/" + tool_name)
Error in Label: Illegal absolute label syntax: invalid label 'tools/PackageParser_deploy.jar': absolute label must begin with '@' or '//'
ERROR: initialization of module '.ijwb/aspects/intellij_info_bundled.bzl' failed

Is there a workaround except for upgrading bazel version?

Thanks

odisseus:
Try building the latest version of plugin with this patch:
```
diff --git a/aspect/BUILD b/aspect/BUILD
index 9a2b82d4b..10879643e 100644
--- a/aspect/BUILD
+++ b/aspect/BUILD
@@ -95,7 +95,7 @@ genrule(
     srcs = ["intellij_info.bzl"],
     outs = ["intellij_info_bundled.bzl"],
     cmd = "cat $(SRCS) >$@ && " +
-          "sed -i -e 's,//%s/tools:\" + tool_name,tools/\" + tool_name,g' $@ && " % _dev_aspect_path +
+          "sed -i -e 's,//%s/tools:\" + tool_name,//.ijwb/aspects:tools/\" + tool_name,g' $@ && " % _dev_aspect_path +
           "sed -i -e 's,//%s:flag_hack,:flag_hack,g' $@  && " % _dev_aspect_path +
           "sed -i -e 's,:intellij_info_impl.bzl,:intellij_info_impl_bundled.bzl,g' $@",
 )
```

odisseus:
[...]

PR #7483, opened 2025-03-18
Use absolute labels for tools in the IntelliJ aspect

odisseus:
This is required for compatibility with Bazel 6.0.0. See https://github.com/bazelbuild/intellij/issues/7470#issuecomment-2743086423

Tested manually on `examples/java/greetings_project` (commit c7a2a83ff681d22fa5b904bd973c91644fcc7aec).


LeFrosch:
This could not be merged as is because of clwb and I'd prefer to just stick with relative labels in the aspects.

PR #7487, opened 2025-03-21
Fixed a typo in the location of JARs which are used by the aspect sync.

odisseus:
# Checklist

details.

Issue number: #7470

# Description of this change

This fixes aspect sync for Bazel versions 6.1.0 and newer. See https://github.com/bazelbuild/intellij/issues/7470#issuecomment-2743086423

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/aspects/storage/AspectStorageService.kt
+++ b/base/src/com/google/idea/blaze/base/sync/aspects/storage/AspectStorageService.kt
@@ -39,7 +39,7 @@ import java.nio.file.Path
 import java.util.*
 
 private const val ASPECT_TASK_TITLE = "Write Aspects"
-private const val ASPECT_DIRECTORY = "aspects"
+private const val ASPECT_DIRECTORY = "aspect"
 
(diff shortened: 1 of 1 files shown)


LeFrosch:
If there are no hard-coded references to `aspect` over `aspects` this change is unnecessary, since all label references should be relative. In the case for Bazel 6.0, I'd argue that it might be fair to request users to upgrade to Bazel 6.1+ and Bazel 6 will be deprecated in June anyway.

odisseus:
Newer versions are also affected by this typo. In fact, there are quite a few hardcoded [references](https://github.com/search?q=repo%3Abazelbuild%2Fintellij%20%22aspect%2Ftools%22&type=code) to `aspect/tools`.

LeFrosch:
They should not matter because they are either only used for building or are stripped by the sed scripts. What do you mean affect? I am under the impression that when using Bazel 6.1+ everything is fine?

odisseus:
Hm. I'm pretty sure I have seen sync failures caused by the wrong directory path, but now I cannot reproduce them.

jastice:
are we going to follow up on this one?

jastice:
since we don't have current reproduction, closing

PR #7493, opened 2025-03-24
 Make ProjectOpenProcessors no strong info holder by default in CLion

LeFrosch:
Follow up for #7440.

In order to not automatically import every project as a Bazel project even if there are other project models available, the ProjectOpenProcessors cannot be strong info holders by default.

The old behavior can be restored in clwb by setting the bazel.project.auto.open registry key to true and is kept for ijwb.


ujohnny (approved):

LeFrosch:
CC @dkashyn-sfdc

PR #7521, opened 2025-03-28
[AOSP-pick] Isolate kotlin coroutine debugging use case

sellophane:
Cherry pick AOSP commit [89f52f6b0e05938e4ea778653bc91c0c8dd69f10](https://cs.android.com/android-studio/platform/tools/adt/idea/+/89f52f6b0e05938e4ea778653bc91c0c8dd69f10).

Note `externalDependenciesFor()` does what it says. It returns external
dependencies of a label but it does not return transitive dependencies.

The replacement methods added in this change do the same, however their
name reflect the intention of these methods, which is different. These
methods are supposed to tell whether a label depends directly or
indirectly on one of the well-known targets. The existing implementation
and its replacement, which is supposed to do the same as the existing
implementation, do not do what is needed, hence the new name.

Bug: 397354312
Test: n/a
Change-Id: Ice510c4199c377011d9aefde406094bb03c265a2

AOSP: 89f52f6b0e05938e4ea778653bc91c0c8dd69f10


LeFrosch:
@odisseus Either in aswb or in the non-public part of the AOSP.

LeFrosch (approved):

PR #7527, opened 2025-03-28
[AOSP-pick] Build workflows cleanup in tools/adt/idea (Part 2)

sellophane:
Cherry pick AOSP commit [1fd9e1e590e1ff78e2e4566fefbdbb3f4d6d8c01](https://cs.android.com/android-studio/platform/tools/adt/idea/+/1fd9e1e590e1ff78e2e4566fefbdbb3f4d6d8c01).

STAT (diff to AOSP): 208 insertions(+), 88 deletion(-)

This cleanup change is purely a refactor and the IDE behavior or the user experience is not expected to change in any way.

1. Migrate all callers of the blaze/bazel builds to using the new build invokers. The callers will now use the `LocalInvoker`, `RabbitBlazeInvoker`, and the `BuildApiBlazeInvoker` as needed.

2. To invoke blaze/bazel, they now use `invoke`, `invokeInfo`, and `invokeQuery` methods instead of the old `run`, `runTest`, `runQuery` etc. These methods take in the BlazeContext as a parameter, so the context used while creating the invoker is no longer used (it'll be removed in the next CLs). The new methods directly return the `BuildEventStreamProvider`. As a result, the `CommandRunner` and the `BuildResultHelper` abstractions are no longer used.

3. Most of the unit tests and integration tests are updated as necessary to be able to test the new invokers, except some integration tests. Since the invoker methods now return a stream that cannot be consumed more than once, we parse it to construct the build outputs within the apk build steps. This is the main reason some tests break and it needs more changes to make them work. They are currently marked as ignored with a TODO. See b/396309705 .

4. In the subsequent CL, the unused code will be deleted. This includes several abstract classes and interfaces, local test finder strategy, and a few tests.

Bug:374906681
Test: n/a
Change-Id: I3fa6311a269bbc18af9118de81d29d1eb73078f7

AOSP: 1fd9e1e590e1ff78e2e4566fefbdbb3f4d6d8c01

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
+++ b/base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
@@ -29,7 +29,8 @@ import com.google.idea.blaze.base.command.BlazeCommand;
 import com.google.idea.blaze.base.command.BlazeCommandName;
 import com.google.idea.blaze.base.command.BlazeFlags;
 import com.google.idea.blaze.base.command.BlazeInvocationContext;
-import com.google.idea.blaze.base.command.buildresult.BuildResultHelper;
+import com.google.idea.blaze.base.command.buildresult.BuildResultParser;
+import com.google.idea.blaze.base.command.buildresult.bepparser.BuildEventStreamProvider;
 import com.google.idea.blaze.base.console.BlazeConsoleLineProcessorProvider;
 import com.google.idea.blaze.base.issueparser.ToolWindowTaskIssueOutputFilter;
 import com.google.idea.blaze.base.model.primitives.WorkspaceRoot;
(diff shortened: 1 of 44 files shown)


LeFrosch (approved):

base/src/com/google/idea/blaze/base/run/confighandler/BlazeCommandGenericRunConfigurationRunner.java
```diff
@@ -252,10 +211,13 @@ protected ConsoleView createConsole() {
...
-                      invoker
-                          .getCommandRunner()
-                          .run(project, blazeCommandBuilder, buildResultHelper, context, envVars));
+                  () -> {
+                    try (BuildEventStreamProvider streamProvider =
+                        invoker.invoke(blazeCommandBuilder, context)) {
+                      return BlazeBuildOutputs.fromParsedBepOutput(
+                          BuildResultParser.getBuildOutput(streamProvider, Interners.STRING));
+                    }
+                  });
```
odisseus, line 220: The variable `envVars` is unused now. This is the cause of #7648.

PR #7545, opened 2025-04-02
[AOSP-pick] Move ASwB test dependencies to bzlmod

sellophane:
Cherry pick AOSP commit [4f69b14d55509e510aaeea943578110aba4acbf6](https://cs.android.com/android-studio/platform/tools/adt/idea/+/4f69b14d55509e510aaeea943578110aba4acbf6).

STAT (diff to AOSP): 0 insertions(+), 78 deletion(-)

To make test_deps bzlmod-compatible, all the artifacts, Bazel central
registry, and Bazel have been consolidated into a single repo rule that
can be called in vendor.MODULE.bazel.

Bug: 340640065
Test: N/A
Change-Id: I471041f39727181b0b7dacfa61f2a83968057be7

AOSP: 4f69b14d55509e510aaeea943578110aba4acbf6


LeFrosch:
What about aswb/testing/test_deps/deps.bzl ?

testing/src/com/google/idea/testing/runfiles/Runfiles.java
```diff
@@ -32,6 +33,12 @@ public static Path runfilesPath() {
     return runfilesPath("");
   }
 
+  /** Returns the runtime location of a data dependency in an external workspace. */
+  public static Path runfilesPath(String workspace, String path) throws IOException {
+    String combinedPath = workspace + "/" + path;
```
odisseus, line 38: I wonder if this would work on Windows.

LeFrosch: Should work, as far as I know you can use both (`/` and `\`) on Windows as path separators. But if this wasn't a pick I'd also prefer `Path.of` here.

LeFrosch (approved):

PR #7553, opened 2025-04-04
Bump intellij SDK versions

LeFrosch:
Bump all versions and move the maintenance script to tools. Rules kotlin seem to be broken for running the binary target :c

### DIFF
--- a/clwb/src/com/google/idea/blaze/clwb/run/BlazeDebuggerKind.java
+++ b/clwb/src/com/google/idea/blaze/clwb/run/BlazeDebuggerKind.java
@@ -53,7 +53,8 @@ public enum BlazeDebuggerKind {
 
     return switch (kind) {
       case BUNDLED_GDB, CUSTOM_GDB -> gdbBundledOrServer();
-      case BUNDLED_LLDB -> BUNDLED_LLDB;
+      // case BUNDLED_LLDB, CUSTOM_LLDB, use default for SDK compat reasons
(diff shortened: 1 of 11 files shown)


sellophane (approved):

clwb/src/com/google/idea/blaze/clwb/run/BlazeDebuggerKind.java
```diff
@@ -53,7 +53,8 @@ public static BlazeDebuggerKind byDefaultToolchain() {
 
     return switch (kind) {
       case BUNDLED_GDB, CUSTOM_GDB -> gdbBundledOrServer();
-      case BUNDLED_LLDB -> BUNDLED_LLDB;
+      // case BUNDLED_LLDB, CUSTOM_LLDB, use default for SDK compat reasons
+      default -> BUNDLED_LLDB;
```
ujohnny, line 57: I was going to mention here that your `ROSETTA_GDB` will fall into the wrong branch logically (though it can't happen in runtime here)

ujohnny: thus in my implementation it was like 
https://github.com/ujohnny/intellij/blob/a89338e9e868b64e53178aaa7d116bcbcaf53ce4/clwb/src/com/google/idea/blaze/clwb/run/BlazeDebuggerKind.java#L54

PR #7597, opened 2025-04-11
update java example

ujohnny:
The example was pretty outdated, also removed all the traces of WORKSPACE from it


odisseus:
We still support older version of Bazel ([down to 6.1.0](https://github.com/bazelbuild/intellij/pull/7487#issuecomment-2747275285)) which allow the old workspace definitions. Maybe it is worth to keep the old and new example as separate projects?

ujohnny:
> which allow the old workspace definitions. 

To be honest I'd say there's no point to have WORKSPACE files in 2025 at least in examples. The approach is dead and in half a year it will not work anymore. I'd really like to have future proof examples.

LeFrosch (approved):

Issue #7586, opened 2025-04-10
IllegalArgumentException: Not a valid label, no target name found


odisseus:
### Description of the bug:

```
java.lang.IllegalArgumentException: Not a valid label, no target name found: 
	at com.google.idea.blaze.base.model.primitives.Label.create(Label.java:44)
	at com.google.idea.blaze.base.command.buildresult.BuildEventProtocolOutputReader.parseTestResult(BuildEventProtocolOutputReader.java:122)
	at com.google.idea.blaze.base.command.buildresult.BuildEventProtocolOutputReader.parseTestResults(BuildEventProtocolOutputReader.java:88)
	at com.google.idea.blaze.base.command.buildresult.BuildResultParser.getTestResults(BuildResultParser.java:76)
	at com.google.idea.blaze.java.run.BlazeJavaRunProfileState.lambda$getCommandRunnerProcessHandler$0(BlazeJavaRunProfileState.java:370)
```

### Which category does this issue belong to?

Intellij

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Import `examples/kotlin/simple_project`.
2. Place a breakpoint inside `main/ProjectRunner.kt` and start debugging.
3. Keep stepping through until the program finishes.

### Which Intellij IDE are you using? Please provide the specific version.

IC-243.23654.153

### What programming languages and tools are you using? Please provide specific versions.

Kotlin

### What Bazel plugin version are you using?

e1fbe97c8e618e4e68905701fc1590a6ea17fbc7

### Any other information, logs, or outputs that you want to share?

Apparently this is a side effect of #7562. The code assumes that the `label` field is always present in `ActionCompletedId`. In fact, [it is optional](https://github.com/bazelbuild/bazel/blame/68277b3009fa5c748118672d2b2397f9a7596540/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto#L178), even though it isn't [marked as such](https://github.com/protocolbuffers/protobuf/issues/10463). The Protobuf serialiser interprets the absence of value as the default vale (empty string in this case).

#### The BEP message that causes this error looks like this:
```
id {
  action_completed {
    primary_output: "bazel-out/stable-status.txt"
    configuration {
      id: "system"
    }
  }
}
action {
  success: true
  primary_output {
[...]

PR #7598, opened 2025-04-11
Do not attempt to parse test results from an ACTION_COMPLETED build event if the event isn't associated with any label.

odisseus:
This is a quick fix for #7586. There might be a better way to handle these events, but this seems to be good enough.

# Checklist

details.

Issue number: #7586
# Description of this change


ujohnny:
I'd just suggest to revert the original pick

odisseus:
I assume the original commit was solving a different problem? Wouldn't that problem just come back?

PR #7606, opened 2025-04-14
fix: replace string with modern providers

comius:
This is cherrypick of: 15bdabea953ae75a4b38acb0df5dd15981bb0576
PiperOrigin-RevId: 745510931

Strings used to refer to legacy struct providers, which were removed from Bazel.

Legacy struct providers have been deprecated by Bazel. Replacing them with modern providers, will make it possible to simplify and remove legacy handling from Blaze.

The change is a no-op.

More information: https://github.com/bazelbuild/bazel/issues/25836

### DIFF
--- a/aspect/fast_build_info.bzl
+++ b/aspect/fast_build_info.bzl
@@ -113,11 +113,10 @@ def _fast_build_info_impl(target, ctx):
     if write_output:
         output_file = ctx.actions.declare_file(target.label.name + ".ide-fast-build-info.txt")
         ctx.actions.write(output_file, proto.encode_text(struct_omit_none(**info)))
-        output_files += [output_file]
+        output_files.append(output_file)
 
     output_groups = depset(output_files, transitive = dep_outputs)
-
-    return struct(output_groups = {"ide-fast-build": output_groups})
+    return [OutputGroupInfo(**{"ide-fast-build": output_groups})]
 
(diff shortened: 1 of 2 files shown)


LeFrosch (approved):
Looks good to me, CC @sellophane

aspect/fast_build_info.bzl
```diff
@@ -113,11 +113,10 @@ def _fast_build_info_impl(target, ctx):
     if write_output:
         output_file = ctx.actions.declare_file(target.label.name + ".ide-fast-build-info.txt")
         ctx.actions.write(output_file, proto.encode_text(struct_omit_none(**info)))
-        output_files += [output_file]
+        output_files.append(output_file)
 
     output_groups = depset(output_files, transitive = dep_outputs)
-
-    return struct(output_groups = {"ide-fast-build": output_groups})
+    return [OutputGroupInfo(**{"ide-fast-build": output_groups})]
```
ujohnny, line 119: I guess we can later fix this unconventional group name.

ujohnny (approved):
@comius thanks for the contribution!

aspect/fast_build_info.bzl
comius: Yes, the `**{}` is used, because there are minuses in the name.

PR #7611, opened 2025-04-14
Run buildifier to add missing loads

comius:
Run `buildifier --lint=fix  -r -v`. This add `loads` for the rules that were removed from Bazel 8. By Bazel 9 automatic loads will be disabled and all loads will need to be explicitly present.

Some loads that cause test failures are omitted and should be done in a separate PRs.

See: https://github.com/bazelbuild/bazel/issues/25755


aspect/intellij_info_impl.bzl
```diff
@@ -4,6 +4,9 @@ load(
     "@bazel_tools//tools/build_defs/cc:action_names.bzl",
     "ACTION_NAMES",
 )
+load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
+load("@rules_cc//cc/common:cc_info.bzl", "CcInfo")
+load("@rules_java//java/common:java_common.bzl", "java_common")
```
LeFrosch, line 9: This will not work in the aspects. If we really need these load statements here we need to template this for comparability reasons.

MODULE.bazel
```diff
@@ -17,6 +17,7 @@ bazel_dep(
     name = "rules_python",
     version = "1.2.0",
 )
+bazel_dep(name = "rules_cc", version = "0.0.17")
```
LeFrosch: I wonder why we need rules_cc to build the plugin?

LeFrosch (changes requested):
Overall, I think this is a good idea. But maybe in smaller batches and I think we need to decide if we want to keep compatibility with AOSP, because this will make picks harder in future. 

CC @ujohnny @sellophane

aspect/intellij_info_impl.bzl
comius: Ack. I reverted this part of the change. So the remaining change is still generated by buildifier, but it shouldn't break any functionality. Let's take this part to a second smaller PR.

MODULE.bazel
comius: Because they were referred to from the aspect. I didn't know that you used templating for that. I'll do a second PR to address JavaInfo and java_common use in the aspects. Since I'm not familiar with the setup, it would be useful if somebody else can fix @rules_cc references in the aspect to use templates.

comius: I had to add it back. There are references from the tests that are located in main repository. I added it as a dev_dependency, though.

aspect/intellij_info_impl.bzl
LeFrosch: I agree, let's upgrade the aspects separately since they might require more work.

querysync/javatests/com/google/idea/blaze/qsync/testdata/deppathkinds/BUILD
```diff
@@ -1,6 +1,6 @@
-load("@rules_java//java:defs.bzl", "java_library")
-load("@protobuf//bazel:proto_library.bzl", "proto_library")
 load("@protobuf//bazel:java_proto_library.bzl", "java_proto_library")
+load("@protobuf//bazel:proto_library.bzl", "proto_library")
+load("@rules_java//java:defs.bzl", "java_library")
```
LeFrosch, line 3: Just a question, why are there two different kinds of loads: 
1. `load("@rules_java//java:defs.bzl", "java_library")`
2. `load("@rules_java//java:java_library.bzl", "java_library")`

skylark/BUILD
```diff
@@ -12,14 +12,13 @@ load(
     "stamped_plugin_xml",
 )
 
-load("@rules_java//java:defs.bzl", "java_library")
+#
+# Description: A Starlark debugging client for IntelliJ.
+#
```
LeFrosch, line 17: I think this comment was supposed to stay on top of the file.

LeFrosch (approved):
Thanks a lot for this. Do we want to merge this? @ujohnny

skylark/BUILD
comius: Fixed. Yes, buidifier does this is there's no empty line to the first load. In that case it considers comment to belong to the load when reordering them.

querysync/javatests/com/google/idea/blaze/qsync/testdata/deppathkinds/BUILD
comius: `defs.bzl` is older (and legacy). At the moment we're advising everybody to use `<symbol>.bzl`. This results in minimal loads (defs.bzl needs to load everything, and by accident there java_proto_library in rules_java, that pulls down protobuf).

If there's a preexisting use of `defs.bzl`, buildifier won't touch it.

PR #7614, opened 2025-04-15
Remove reference to JavaInfo from fast_build_info

comius:
By Bazel 9 automatic loads will be disabled and all loads will need to be explicitly present.

This is a followup of https://github.com/bazelbuild/intellij/pull/7611
See: https://github.com/bazelbuild/bazel/issues/25755


LeFrosch:
I think in the fast build stuff we could add rules_java loads since they are java only CC @sellophane

odisseus (approved):

PR #7621, opened 2025-04-18
Return NO_STATUS test result if there was a build failure

odisseus:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/472#issuecomment-2801451073

# Description of this change

This prevents the spurious "All tests passed" message.
Screenshot 2025-04-18 at 15 59 04


odisseus:
Well, the real cause is the [new model](https://github.com/bazelbuild/intellij/pull/7527) of communication with the build process, which doesn't take into account its exit code. Either we try to factor the exit code into the new model (which I feel goes against its spirit), or we try to detect the build failure by analysing the BEP stream (which AFAIK is always possible).

ujohnny:
> I still have a feeling this rather addresses the symptom than the cause.

Indeed the cause is a misconfigured process handler.

odisseus:
I have confirmed that #7622 also fixes this issue, although the output looks a bit different.
Screenshot 2025-05-14 at 06 21 33

Screenshot 2025-05-14 at 06 19 06

PR #7628, opened 2025-04-22
Created a collection of pre-built WorkspaceLanguageSettings which are frequently used in tests.

odisseus:
This is a small refactor of unit and integration tests which doesn't affect their execution. The sole purpose of this change is to reduce boilerplate and make the tests more readable.


jastice:
not critical but keeping it open since it just needs approval

base/tests/utils/unit/com/google/idea/blaze/base/sync/projectview/WorkspaceLanguageSettingsExamples.java
```diff
@@ -0,0 +1,19 @@
+package com.google.idea.blaze.base.sync.projectview;
```
LeFrosch, line 16: Copyright missing.

odisseus: Fixed

PR #7636, opened 2025-05-02
headless tests: improve error logging

ujohnny:
Truth just prints `toString()` version of asserted elements that does not have required information.

* Let truth print the issue itself
* Also assert with an extra message for debugging purposes


odisseus (approved):

LeFrosch:
Actually, I don't think this change was necessary since the log should be printed (but not sure). However, I want to improve the overall error reporting anyway.

PR #7643, opened 2025-05-05
Fix annotation_processing in fast_build aspect

comius:
Previously java_info was shadowed with a new structure, so annotation_processing was certainly missing.


LeFrosch (approved):
Thanks, looks like an obvious bug.

Issue #7619, opened 2025-04-17
Generated "test" Run Configuration in context menu now requires manual user setup


mitchhentgesspotify:
### Description of the bug:

It used to be possible (with plugin version `2025.02.04.0.1-api-version-243`) to right-click a Bazel Package's folder and click on a generated Run Configuration that would execute tests:

![Image](https://github.com/user-attachments/assets/749076f5-b87e-4642-bd70-44647c7c6a05)

With plugin version `2025.02.18.0.1-api-version-243` (and later), the context menu now shows `'Unnamed'` instead, and choosing it then prompts the user to fill out the Run Configuration (and it's not immediately obvious which fields are still properly generated, and which need to be filled in):

![Image](https://github.com/user-attachments/assets/6e27e347-5da4-4d25-af8a-66da33b59c90)

![Image](https://github.com/user-attachments/assets/dcf1e1ea-c933-47a9-9b0c-f9e05f143a51)

### Which category does this issue belong to?

Intellij

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

[See this test repo](https://github.com/mitchhentgesspotify/intellij-bazel-right-click-test-repro)

### Which Intellij IDE are you using? Please provide the specific version.

2024.3.5

### What programming languages and tools are you using? Please provide specific versions.

Java

### What Bazel plugin version are you using?

2025.04.01.0.1-api-version-243

### Have you found anything relevant by searching the web?

Perhaps da8d881e96753fc30ef6f29a5a305f91804c04fa is the commit that caused the issue

jhemarcos:
This issue really hurts productivity.

jastice:
Thanks for the report! We'll get the issue fixed ASAP.
Meanwhile, if you're working with JVM projects, please have a look at the [new Bazel plugin](https://plugins.jetbrains.com/plugin/22977-bazel-eap-/); we're generally able to address issues more quickly there.

odisseus:
This is an unintended consequence of #7275. PR #7637 may fix the UX inconvenience.

ujohnny:
fixed in #7659

PR #7659, opened 2025-05-09
Revert "Make run configurations configurable (#7275)"

ujohnny:
This reverts commit da8d881e96753fc30ef6f29a5a305f91804c04fa.

# Checklist

details.

# Description of this change

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/BlazeCommandRunConfiguration.java
+++ b/base/src/com/google/idea/blaze/base/run/BlazeCommandRunConfiguration.java
@@ -44,7 +44,6 @@ import com.google.idea.blaze.base.settings.Blaze;
 import com.google.idea.blaze.base.settings.BlazeImportSettings.ProjectType;
 import com.google.idea.blaze.base.sync.data.BlazeProjectDataManager;
 import com.google.idea.blaze.base.ui.UiUtil;
-import com.google.idea.common.experiments.BoolExperiment;
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.Executor;
 import com.intellij.execution.RunManagerEx;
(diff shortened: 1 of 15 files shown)


odisseus (approved):

LeFrosch (approved):

SergejSalnikov:
What was the reason to revert?

odisseus:
IIRC there was a scenario (perhaps running a unit test), in which clicking the green triangle should just launch the run configuration at once. After this change, the "Modify run configuration" dialog started to pop up instead, even though all the necessary data were already present or could be inferred.

See #7619 for more context.

base/src/com/google/idea/blaze/base/run/BlazeCommandRunConfiguration.java
```diff
@@ -770,85 +689,14 @@ protected void applyEditorTo(BlazeCommandRunConfiguration config) {
...
-        updateHandlerEditor(config, config.handlerProvider);
+      if (config.handlerProvider != handlerProvider) {
+        updateHandlerEditor(config);
         handlerStateEditor.resetEditorFrom(config.handler.getState());
       } else {
         handlerStateEditor.applyEditorTo(config.handler.getState());
       }
     }
-
-    private void updateHandlerListAndAutoSelect(BlazeCommandRunConfiguration config, boolean autoSelect) {
```
LeFrosch, line 798: Also IIRC some changes here caused a bunch of race conditions with updating the run config. I think the race conditions are still there / were there before, but after this change they became unusable with the RCs.

SergejSalnikov: Thanks for the context. I think I know where the problem is. `applyEditorTo` is triggered every 500ms by an alarm, which in turn causes repeated calls to `updateTargetKind`. 

This feature is very useful for dual language targets. For example python/cpython, go/cgo and etc.

LeFrosch: If you find a fix for the RC and end up picking this PR, it would be amazing if you could share the fix, since this RC has been a source of friction for a long time ^^

SergejSalnikov: The CL mostly works as is.

I had to update all DebugggerRunners to be active only when corresponding RunConfigState is present and ignore the rule kinds:

For java it would be and so on:
```
BlazeJavaRunConfigState handlerState =
        config.getHandlerStateIfType(BlazeJavaRunConfigState.class);
```

image

SergejSalnikov: And for `BlazeCommandRunConfiguration` I do not initially assign  `TargetState.PENDING` handler, but rather use the current detection mechanism and let user to override it later if they want.

PR #7661, opened 2025-05-09
Remove aswb

LeFrosch:
The aswb part cannot be build, is unmaintained and we no longer actively pick aosp changes. No need to keep it around anymore.


odisseus:
Woah woah woah. Are we completely giving up on the AOSP picks?

LeFrosch:
We can a just the tool to exclude aswb changes if we decide to later on continue with the picking.

odisseus:
A contributor to Android Studio has been told to submit his PR (#7664) to this repository, so that the Google developers may review his code and pull it into their repository. For the sake of enabling such PRs, it may be worth to keep the aswb code in a separate branch.

ujohnny (approved):

PR #7665, opened 2025-05-12
Update CHANGELOG for v2025.05.13 release

satyanandak:


LeFrosch:
I guess only these two are interesting:
- #7643
- #7659 should fix #7619

mai93 (approved):

PR #7667, opened 2025-05-12
CLion only forward builtin includes from a custom sysroot

LeFrosch:

Issue number: #7489

# Description of this change

Builtin includes reported by the CcToolchainInfo provider sometimes contain include paths for both libc++ and libstdc++ regardless of what compiler options are specified. CLion is able to find the right include path during compiler info collection automatically. But if both include paths are manually added to the switch builder CLion might select the wrong one after all.

Therefore, only forward builtin includes which point into a custom sysroot and let CLion detect the correct builtin includes during compiler info collection.


LeFrosch:
This change requires further investigation, why does Bazel report both include directories? What if toolchain actually does define include paths that are not in the sysroot and cannot be discovered by CLion?

LeFrosch:
As far as I understand the issue here is caused by a bug in rules_cc. Maybe we should not forward all include paths reported by rules_cc since they have proven to be unreliable and instead relay on CLions compiler info collection.

But not sure yet how to filter the include paths reported by rules_cc.

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -313,13 +313,18 @@ private WorkspaceModel calculateConfigurations(
...
-            .flatMap(resolver)
-            .map(File::getAbsolutePath)
-            .forEach(compilerSwitchesBuilder::withSystemIncludePath);
+        // add includes from a custom sysroot as system includes
+        // Note: Only add includes from a custom sysroot manually, CLion can derive the other
+        // bulletin includes during the compiler info collection. Manually adding all builtin
+        // includes can lead to headers being resolved into the wrong include directory.
+        final var sysroot = compilerSettings.getSysroot();
+        if (sysroot != null) {
+          compilerSettings.getBuiltInIncludes().stream()
```
ujohnny: let's revise this PR and consider doing this with registry key. I'm unsure if this change as is would lead to some unresolved headers etc, this has to be validated with custom toolchain with sysroot.

cpp/src/com/google/idea/blaze/cpp/BlazeCWorkspace.java
```diff
@@ -182,6 +186,28 @@ private static CidrCompilerSwitches buildSwitchBuilder(
     return combinedBuilder.build();
   }
 
+  private static void logResolveConfiguration(BlazeResolveConfiguration config) {
+    final var builder = new StringBuilder();
```
ujohnny, line 194: You do some computations here which are ignored by default (because log level is not set), so it's better to wrap them with `if(logger.isTraceEnabled())`

ujohnny (changes requested):

PR #7673, opened 2025-05-15
Do not egerly refresh the execroot on import

LeFrosch:
A forced refresh, like:
```java
VfsUtil.markDirtyAndRefresh(
  /* async= */ false, 
  /* recursive= */ true, 
  /* reloadChildren= */ true, 
  root
);
```

Can be very expensive and lead to a modal refresh dialog. Furthermore, the manual refresh should not be necessary, since the VFS should pull changes as necessary.

### DIFF
--- a/base/src/com/google/idea/blaze/base/sync/ProjectUpdateSyncTask.java
+++ b/base/src/com/google/idea/blaze/base/sync/ProjectUpdateSyncTask.java
@@ -73,6 +73,7 @@ import com.intellij.openapi.roots.ContentEntry;
 import com.intellij.openapi.roots.ModifiableRootModel;
 import com.intellij.openapi.roots.ex.ProjectRootManagerEx;
 import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.util.registry.Registry;
 import com.intellij.openapi.vfs.VfsUtil;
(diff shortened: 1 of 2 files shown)


base/src/com/google/idea/blaze/base/sync/ProjectUpdateSyncTask.java
```diff
@@ -285,6 +286,10 @@ private void run(BlazeContext context) throws SyncCanceledException, SyncFailedE
 
   private static void refreshVirtualFileSystem(
       BlazeContext context, Project project, BlazeProjectData blazeProjectData) {
```
odisseus, line 288: This is a private method which is called only once. I would rather perform this check before this method, not inside it.

LeFrosch:
This change requires further investigation. Maybe the refresh is indeed necessary but the mark dirty might be not.

ujohnny (changes requested):
Based on previous changes in [72110e6bbbaab7645a664d47241cc4edfda8066d](https://github.com/bazelbuild/intellij/commit/72110e6bbbaab7645a664d47241cc4edfda8066d) this as is might cause some issues, maybe just explicit `mark dirty` should be removed

LeFrosch:
Superseded by #7715

Issue #7677, opened 2025-05-16
Support thrift_library and go_thrift_library indexing


H5-O5:
### Description of the feature request:

Hi I'm implementing intellij's indexing thrift-generated go file the same way as it handles proto-generated go files (go_proto_library). I have proof-of-concept working, basically just add thrift_library wherever proto_library exists, and add go_thrift_library wherever go_proto_library exists. I wonder what I can do to help make this PR merged. This commit I want to get merged is [here](https://github.com/H5-O5/intellij/commit/3c90d942c7eb25787fbfd6449206ccf51a72a79d). I'll create a PR if the maintainer says it is OK to do so.

### Which category does this issue belong to?

GoLand, Intellij

### What underlying problem are you trying to solve with this feature?

Make the plugin work with go_thrift_library rules.

H5-O5:
@jastice Is it OK if I create a PR for this issue?

LeFrosch:
Closing this. Refer to the PR for the reasoning.

PR #7687, opened 2025-05-22
feat: support indexing thrift_library

H5-O5:
# Checklist

details.

Issue number: #7677 

# Description of this change

Basically I just added THRIFT_LIBRARY alongside PROTO_LIBRARY, and `go_thrift_library` alongside `go_proto_library`. Since they both use `library[IntelliJInfo].output_groups["intellij-sources-go-outputs"]` I don't have to change how the index work. I just tell the code to recognize go_thrift_library and re-use `BlazeQueryParser::visitProtoRule` for indexing.

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -253,7 +253,7 @@ def get_aspect_ids(ctx):
 
 def _is_language_specific_proto_library(ctx, target, semantics):
     """Returns True if the target is a proto library with attached language-specific aspect."""
-    if ctx.rule.kind != "proto_library":
+    if ctx.rule.kind != "proto_library" and ctx.rule.kind != "thrift_library":
         return False
     if java_info_in_target(target):
         return True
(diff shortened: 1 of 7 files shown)


LeFrosch:
Can you provide a link to the rule set for thrift? I cannot find it in the Bazel Central Registry.

H5-O5:
> Can you provide a link to the rule set for thrift? I cannot find it in the Bazel Central Registry.

The go_thrift_library rule is from a private rule set that isn’t really ready for publishing to the Bazel Central Registry. But we can help depending on what you’d like to do.

If you just want to make sure the change works—yes, it does. It’s a simple and straightforward change, and I’ve tested it.

If you’d prefer to have some tests to make sure it won’t break in the future, we can put together a demo GitHub repo with the go_thrift_library rule included, so tests can be added there.

Setting up a proper BCR entry would be a lot more effort and probably not worth it right now. Let me know what you’d prefer, and we’ll work from there.

LeFrosch:
Unfortunately, we don’t want to merge support for private rule sets into the main plugin, as we aim to keep it focused on broadly applicable features. Maintaining support for internal or organization-specific rules is out of scope.

You're welcome to maintain a fork for your needs — sorry for the inconvenience. Ideally, we’d have an extension point to support these scenarios, but we don’t have anything like that at the moment.

Thanks again for understanding!

Issue #7703, opened 2025-05-27
"Code insight is unavailable because the sbt project is not loaded" for hybrid sbt / bazel project


Friendseeker:
### Issue

When a project is hybrid sbt / bazel (has `project/build.properties`), IntelliJ would think this project is a sbt project, and disable code insight since the sbt project is not loaded.

![Image](https://github.com/user-attachments/assets/f0e47929-3ad8-48a0-87d4-6c8f0941c19a)

### Proposed solution

Similar to https://github.com/JetBrains/intellij-scala/commit/e7a77382a095087d29c1c3bb82b9e7b183c4d921

Use `BuildToolModuleHandler` API to register that the project is already imported via Bazel.

### Which category does this issue belong to?

Intellij

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Create a Hello World Scala Bazel Project

Add a build.sbt and `project/build.properties` for the Hello World Project.

### Which Intellij IDE are you using? Please provide the specific version.

2025.1.1.1

### What programming languages and tools are you using? Please provide specific versions.

Scala

### What Bazel plugin version are you using?

2025.04.29.0.1-api-version-251

PR #7704, opened 2025-05-27
Implement `BuildToolModuleHandler` for Scala Projects to register Bazel Imported Scala Projects

Friendseeker:
# Checklist

details.

Issue number: https://github.com/bazelbuild/intellij/issues/7703

# Description of this change

Use `BuildToolModuleHandler` API to register that the project is already imported via Bazel.

## Before the change
image

## After the change
image

### DIFF
--- a/scala/src/META-INF/blaze-scala.xml
+++ b/scala/src/META-INF/blaze-scala.xml
@@ -18,8 +18,4 @@
     <SyncPlugin implementation="com.google.idea.blaze.scala.sync.AlwaysPresentScalaSyncPlugin"/>
     <TargetKindProvider implementation="com.google.idea.blaze.scala.ScalaBlazeRules"/>
   </extensions>
-
-  <extensions defaultExtensionNs="org.jetbrains.sbt">
-    <buildToolModuleHandler implementation="com.google.idea.blaze.scala.BlazeScalaBuildToolModuleHandler"/>
-  </extensions>
 </idea-plugin>
--- a/scala/src/META-INF/blaze-scala.xml
+++ b/scala/src/META-INF/blaze-scala.xml
@@ -18,4 +18,8 @@
     <SyncPlugin implementation="com.google.idea.blaze.scala.sync.AlwaysPresentScalaSyncPlugin"/>
     <TargetKindProvider implementation="com.google.idea.blaze.scala.ScalaBlazeRules"/>
   </extensions>
+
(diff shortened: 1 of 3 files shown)


odisseus (approved):

Jerry-Visier:
> Is `buildToolModuleHandler` endpoint already present in `243`? Because if not, then I'm quite surprised that all tests are green...

`BuildToolModuleHandler` is added to address the ticket https://youtrack.jetbrains.com/issue/SCL-23350

The ticket has fix version `2025.1.516`

The exact commit https://github.com/JetBrains/intellij-scala/commit/e7a77382a095087d29c1c3bb82b9e7b183c4d921 is also only in `idea251.x` branch.

Therefore, it should not be present in `243`.

sdkcompat/v252/BUILD
```diff
@@ -34,6 +34,10 @@ java_library(
         ]),
         intellij_ue = glob([
             "com/google/idea/sdkcompat/javascript/**",  #api223
+            "com/google/idea/sdkcompat/scala/**",  #api250
```
LeFrosch, line 37: The api tag seems wrong here, please use <version where the api changed> - 1. So that if we drop the previous version we can just grep for the API tag.

LeFrosch (changes requested):

sdkcompat/v252/BUILD
LeFrosch: If I understand this correctly the API was introduced in 251, please use 243 as the API tag.

LeFrosch:
Also regarding the concern from @sellophane. If the extension point `buildToolModuleHandler` is not available in 243 and we register it unconditionally in the xml, this might prevent the scala part from loading. Furthermore, it seems that most other extensions are registered in `scala/src/META-INF/scala-contents.xml` rather then `scala/src/META-INF/blaze-scala.xml`.

If this is the case and we cannot register this extension in 243, you would need to register it manually like we did it in CLion for some time. For example: https://github.com/bazelbuild/intellij/blob/9443d50765e3f7832b877bbdb14e73ec105d2894/clwb/sdkcompat/v243/com/google/idea/blaze/clwb/CLionNotificationProvider.kt#L102-L114

odisseus:
Since we don't support 2024.3 anymore (#7742), this PR can be reconsidered.

odisseus (approved):

PR #7715, opened 2025-06-06
Do not mark execution root as dirty

LeFrosch:
Do not mark the execution root as dirty before calling refresh. Foll up for after further investigation #7673.

Results of the investigation:
- Refresh after sync is required to pick up changes from the exec root
- Found an unnecessary refreshe, only refreshing the execution root should be sufficient
- Mark dirty is not required (however, might cause a race condition with the notifier)

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -439,6 +439,9 @@
     <registryKey defaultValue="true"
                  description="Enable basic syntax highlighting in Query Sync's 'Analysis Disabled' mode"
                  key="bazel.qsync.enable.basic.highlighting.in.non.analysis.mode"/>
+    <registryKey defaultValue="false"
+                 description="Mark execroot as dirty before refresh after sync"
+                 key="bazel.sync.mark.dirty"/>
     <editorNotificationProvider implementation="com.google.idea.blaze.base.wizard2.BazelNotificationProvider"/>
   </extensions>
 
--- a/clwb/tests/headlesstests/com/google/idea/blaze/clwb/base/ClwbHeadlessTestCase.java
+++ b/clwb/tests/headlesstests/com/google/idea/blaze/clwb/base/ClwbHeadlessTestCase.java
@@ -56,10 +56,7 @@ public abstract class ClwbHeadlessTestCase extends HeadlessTestCase {
     }
   }
 
-  protected void addAllowedVfsRoots(ArrayList<AllowedVfsRoot> roots) {
-    // required because of the mark dirty in com.google.idea.blaze.base.sync.ProjectUpdateSyncTask#refreshVirtualFileSystem
-    roots.add(AllowedVfsRoot.flat(""));
-  }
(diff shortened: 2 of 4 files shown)


clwb/tests/headlesstests/com/google/idea/blaze/clwb/base/ClwbHeadlessTestCase.java
```diff
@@ -56,10 +56,7 @@ private void setupSandboxBin() {
     }
   }
 
-  protected void addAllowedVfsRoots(ArrayList<AllowedVfsRoot> roots) {
-    // required because of the mark dirty in com.google.idea.blaze.base.sync.ProjectUpdateSyncTask#refreshVirtualFileSystem
-    roots.add(AllowedVfsRoot.flat(""));
-  }
+  protected void addAllowedVfsRoots(ArrayList<AllowedVfsRoot> roots) { }
```
ujohnny, line 59: I'm unsure about this particular change.

base/src/META-INF/blaze-base.xml
```diff
@@ -439,6 +439,9 @@
     <registryKey defaultValue="true"
                  description="Enable basic syntax highlighting in Query Sync's 'Analysis Disabled' mode"
                  key="bazel.qsync.enable.basic.highlighting.in.non.analysis.mode"/>
+    <registryKey defaultValue="false"
```
ujohnny, line 442: We know the consequences of markDirty removal and we're almost sure it's fine for C++. I'm a bit more worried about java support in this case. Shall we make an override for clwb and keep markDirty for ijwb?

ujohnny: FTR the consequences are the following
1. we know bazel has exited and all the files are changed and there are 0 pending changes
2. without markDirty we rely only on local fs notifications and there could be a race that not all the notifications are processed by the moment refresh is called. so new vfs snapshot after refresh might be not-up-to-date with the local filesystem contents. 
3. we know that eventually data will be up-to-date because most likely user will perform alt-tab that causes a refresh or perodic refresh will be performed. but there could be some code right after this in java or somewhere else that expects a correct snapshot.

clwb/tests/headlesstests/com/google/idea/blaze/clwb/base/ClwbHeadlessTestCase.java
LeFrosch: This change only updates the default allowed VFS roots in all headless tests. Previously we had to include the flat execroot because of the reload children flag.

base/src/META-INF/blaze-base.xml
LeFrosch: Yes, let's keep the default behavior for ijwb, since it is basically unmaintained now.

clwb/tests/headlesstests/com/google/idea/blaze/clwb/base/ClwbHeadlessTestCase.java
ujohnny: Got it, thanks

ujohnny (approved):

Issue #7723, opened 2025-06-09
IntelliJ Bazel plugin deleting the user's home directory


jadenPete:
### Description of the bug:

Hello! My name is Jaden; I'm an engineer at Lucid Software. We use the IntelliJ Bazel plugin for our Scala codebase, and we're seeing reports of the IntelliJ Bazel plugin deleting engineers' home directories. It's difficult to tell what circumstances cause this, because in doing so, the program erases the project and any information that might give us a clue, but we were able to collect some error messages from syslog:
```
May  8 16:42:24 MASKED intellij-idea-community_intellij-idea-community.desktop[5606]: 2025-05-08 16:42:24,407 [26674673]   WARN - com.google.idea.blaze.base.sync.projectstructure.ModuleEditorImpl - Could not delete file: /home/MASKED, will try to continue anyway.
```

We also found a stack trace:
```
java.nio.file.DirectoryNotEmptyException: /home/MASKED/.config (children: pulse, )
    at com.intellij.openapi.util.io.FileUtilRt.directoryNotEmptyExceptionWithMoreDiagnostic(FileUtilRt.java:931)
    at com.intellij.openapi.util.io.FileUtilRt.doDelete(FileUtilRt.java:879)
    at com.intellij.openapi.util.io.FileUtilRt.access$100(FileUtilRt.java:26)
    at com.intellij.openapi.util.io.FileUtilRt$1.postVisitDirectory(FileUtilRt.java:837)
    at com.intellij.openapi.util.io.FileUtilRt$1.postVisitDirectory(FileUtilRt.java:813)
    at java.base/java.nio.file.Files.walkFileTree(Files.java:2816)
    at java.base/java.nio.file.Files.walkFileTree(Files.java:2870)
    at com.intellij.openapi.util.io.FileUtilRt.deleteRecursively(FileUtilRt.java:813)
    at com.intellij.openapi.util.io.NioFiles.deleteRecursively(NioFiles.java:227)
    at com.intellij.openapi.vfs.impl.local.LocalFileSystemBase.deleteFile(LocalFileSystemBase.java:330)
    at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.deleteFile(PersistentFSImpl.java:784)
    at com.intellij.openapi.vfs.newvfs.impl.VirtualFileSystemEntry.delete(VirtualFileSystemEntry.java:302)
[...]

PR #7724, opened 2025-06-10
Abort deleting the .iml file if the supplied path doesn't look like one

odisseus:
This is a quick fix for #7723.


sellophane:
I would suggest going with [this](https://github.com/bazelbuild/intellij/pull/7726/)

odisseus:
Superseded by #7726.

PR #7728, opened 2025-06-12
Use relativize in BlazePackage.getPackageRelativePath

LeFrosch:
As pointed out here: https://github.com/bazelbuild/intellij/commit/d2eacd36e6e34f9d76e903e9f3eb3101b1825e7c#r159067641


base/src/com/google/idea/blaze/base/lang/buildfile/search/BlazePackage.java
```diff
@@ -105,15 +105,22 @@ public Label getPackageLabel() {
   }
 
   /**
-   * Returns the file path relative to this blaze package, or null if it does lie inside this
+   * Returns the file path relative to this blaze package, or null if it does not lie inside this
    * package
```
SergejSalnikov: full stop at the end of sentences.

PR #7730, opened 2025-06-12
Remove exclusive flag from headless tests

LeFrosch:
The exclusive flag was introduced so that Bazel does not run into any resource issues. Let's see if it can be removed, because this might reduce test runtime.


LeFrosch:
Tests do not run faster :c

PR #7734, opened 2025-06-13
Highlighting support for binary operators and augmented assignments in BUILD files.

### LINKED ISSUE #7733: `|=` operator is apparently not supported in Starlark
### Description of the bug:

Image

Plain `|` is supported, and `&=` and `^=` are also supported.

### Which category does this issue belong to?

_No response_

Just use `|=` somewhere in Starlark file. All uses seem to have red squiggles, regardless of the context.

IntelliJ IDEA 2025.1.1.1 (Ultimate Edition)

Just Starlark code is enough to repro the error

2025-04-30.0

Nothing, unfortunately :/

_No response_

odisseus:
Supported operators: `& ^ >> << &= ^= |= >>= <<=`

Fixes #7733.

### DIFF
--- a/base/src/com/google/idea/blaze/base/lang/buildfile/lexer/BuildLexerBase.java
+++ b/base/src/com/google/idea/blaze/base/lang/buildfile/lexer/BuildLexerBase.java
@@ -54,6 +54,9 @@ public class BuildLexerBase {
           .put('*', TokenKind.STAR_EQUALS)
           .put('/', TokenKind.SLASH_EQUALS)
           .put('%', TokenKind.PERCENT_EQUALS)
+          .put('^', TokenKind.CARET_EQUALS)
+          .put('&', TokenKind.AMPERSAND_EQUALS)
+          .put('|', TokenKind.PIPE_EQUALS)
           .build();
 
   private final LexerMode mode;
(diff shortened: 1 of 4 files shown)


base/src/com/google/idea/blaze/base/lang/buildfile/lexer/BuildLexerBase.java
```diff
@@ -769,10 +772,28 @@ private void tokenize() {
           popParen();
           break;
         case '>':
-          addToken(TokenKind.GREATER, pos - 1, pos);
+          if (lookaheadIs(0, '>') && lookaheadIs(1, '=')) {
```
LeFrosch, line 775: Maybe we can add a tokenize three chars and move the rest (<<, >>) also to two chars?

LeFrosch (changes requested):
This pre grammar kit stuff is fun lamo

base/src/com/google/idea/blaze/base/lang/buildfile/lexer/BuildLexerBase.java
odisseus: I just copied the code which handles `//=`. Sure, this whole class could use some refactoring (personally I would prefer using something like ANTLR to generate the parser), but I want to keep my changes as simple and straightforward as possible.

LeFrosch: I mean, I just wanted to suggest to stick with the structure that's already there. Like, to put the parsing for 2-token operators into the 2-token-method.

odisseus: I don't quite understand what you're objecting to. "Stick with the structure that's already there" is exactly what I did here. See line 830 for comparison.

PR #7735, opened 2025-06-13
Bump EAP SDKs

LeFrosch:


cpp/BUILD
```diff
@@ -53,7 +53,13 @@ stamped_plugin_xml(
 
 optional_plugin_xml(
     name = "optional_cidr",
-    module = ["com.intellij.modules.cidr.lang"],
```
ujohnny, line 56: `com.intellij.modules.cidr.lang` -> `com.intellij.cidr.lang` should be ok for old versions too

ujohnny (changes requested):

cpp/BUILD
ujohnny: because of 
```
    {
      "name": "com.intellij.cidr.lang",
      "kind": "plugin",
      "classPath": [
        "plugins/c-plugin/lib/c-plugin.jar"
      ]
    },
``` 

in `product-info.json`

LeFrosch: Yes, looks like tests are also passing using just `com.intellij.cidr.lang`. I think I would still like to keep the `select_for_version` since it might useful in future.

ujohnny (approved):

PR #7765, opened 2025-06-26
CLion: add suport for clang-cl

LeFrosch:
Adds simpler heuristic whether clang-cl is used by looking at the current compiler name provided by Bazel.

### DIFF
--- a/clwb/clwb.bazelproject
+++ b/clwb/clwb.bazelproject
@@ -16,7 +16,7 @@ targets:
 workspace_type: intellij_plugin
 
 build_flags:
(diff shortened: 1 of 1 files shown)


clwb/clwb.bazelproject
```diff
@@ -16,7 +16,7 @@ targets:
 workspace_type: intellij_plugin
 
 build_flags:
-  --define=ij_product=clion-oss-under-dev
+  --define=ij_product=clion-2025.1
```
ujohnny: this looks to be an accidental change

ujohnny (approved):

PR #7769, opened 2025-06-27
Disable expand sync to workings set by default

LeFrosch:
For me "Expand Sync to Working Set" always causes problems. Let's disable it by default?


ujohnny:
discussed that external users migration is not a part of this change. applies to new users only

PR #7772, opened 2025-06-30
.bazelfile import if python plugin

andponlin-canva:
# Checklist

details.

_Please see discussion below_

# Description of this change

Some projects may contain `.bazelfile` settings which are specific for Python. The Bazel plugin will fail to load these Python-specific sections (and consequently the whole project) unless the corresponding Bazel-Python extension (`BlazePythonSyncPlugin`) is registered and that will not be the case if the Python plugins are not installed and enabled.

This change provides a way around this problem by allowing an import of additional sections to occur only if the required Python plugin is installed and enabled.

As an example, consider the following files;

`.ijwb/.bazelproject` as;

```
if_python_plugin_import .ijwb/python.bazelproject
```

`python.bazelproject` as;

```
python_code_generator_rule_names:
  test_codegen_files_py
  test_codegen_directory_py
```

With this setup, somebody without the Python plugin installed is still able to open the project.

### DIFF
--- a/base/src/com/google/idea/blaze/base/projectview/section/sections/IfPythonPluginImportSection.java
+++ b/base/src/com/google/idea/blaze/base/projectview/section/sections/IfPythonPluginImportSection.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2025 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.idea.blaze.base.projectview.section.sections;
+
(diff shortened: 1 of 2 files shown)


odisseus:
I don't like this approach. Are we going to define a conditional section for every plugin out there?

IMO the right approach is checking whether a certain language or plugin is enabled, and ignoring the sections which aren't applicable.

andponlin-canva:
> the right approach is checking whether a certain language or plugin is enabled, ...

The sections are only "registered" for parsing if the associated concrete `BlazeSyncPlugin` is instantiated and registered so it's hard to know in advance if they would or would not be available. For example, the following sections are only available if the `BlazePythonSyncPlugin` is initialized;

- `python_debug_flags`
- `python_code_generator_rule_names`

> ...and ignoring the sections which aren't applicable.

A few options that, subject to further investigation, may be viable which are aligned with this;

- The `AlwaysPresentPythonSyncPlugin` plugin could register the Python-related sections so that they were always present but would only have effect if the `BlazePythonSyncPlugin` were also initialized. This is probably a straight-forward unobtrusive solution.
- The Bazel / IJ plugin could more generally warn rather than error if an unknown section were encountered

@odisseus ; Let me know if either of these options were preferable to you.

(may be slow to reply)

andponlin-canva:
> The AlwaysPresentPythonSyncPlugin plugin could register the Python-related sections so that they were always present but would only have effect if the BlazePythonSyncPlugin were also initialized. This is probably a straight-forward unobtrusive solution.

I have modified the PR to this solution instead.

(may be slow to reply)

LeFrosch (approved):
Looks good to me

PR #7783, opened 2025-07-05
Draft: Add support for rule executable_jar

joca-bt:
I'm trying to add support for the `executable_jar` [rule](https://github.com/joca-bt/extra_rules_java/blob/master/docs/executable_jar.md). While these changes add support for debugging from BUILD files, the gutter on source files doesn't recognize the target.

image

Would you be so kind to point me in the right direction to add this functionality?

# Checklist

details.

### DIFF
--- a/java/src/com/google/idea/blaze/java/JavaBlazeRules.java
+++ b/java/src/com/google/idea/blaze/java/JavaBlazeRules.java
@@ -31,6 +31,7 @@ public final class JavaBlazeRules implements Kind.Provider {
 
   /** Java-specific blaze rules. */
   public enum RuleTypes {
+    EXECUTABLE_JAR("executable_jar", LanguageClass.JAVA, RuleType.BINARY),
     JAVA_LIBRARY("java_library", LanguageClass.JAVA, RuleType.LIBRARY),
     JAVA_TEST("java_test", LanguageClass.JAVA, RuleType.TEST),
     JAVA_JUNIT5_TEST("java_junit5_test", LanguageClass.JAVA, RuleType.TEST),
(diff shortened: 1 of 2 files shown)


LeFrosch:
Hey, thanks for the PR. @jastice do you want to proceed with this PR? The rule set is listed on the BCR: https://registry.bazel.build/modules/extra_rules_java

jastice:
Hi @joca-bt, from my side it's good to merge, but I would recommend contributing to the new Bazel plugin by JetBrains at https://github.com/JetBrains/hirschgarten/ instead, as we're going to discontinue support for this one roughly by end of the year.

LeFrosch:
Judging by the code, this is not enough to fully support your rule set, or is it? @joca-bt

joca-bt:
Hello @LeFrosch, it's not. My changes make the rule available in BUILD files and I can debug from there, but in source files the gutter context is not working as I wrote in my initial message above. Could you point me in the right direction to fix that?

@jastice we are still using this plugin, so we'd like to make it available first in this one. Once the other exits EAP we will integrate with that one.

jastice:
> we are still using this plugin, so we'd like to make it available first in this one. Once the other exits EAP we will integrate with that one.

We are going to announce GA this month :)

joca-bt:
@jastice can you provide some pointers on where to start for hirschgarten for these changes? It shows such targets as runnable by default, however it doesn't show the debug option for such targets.

sellophane:
@joca-bt `org.jetbrains.bazel.ui.gutters.BazelRunLineMarkerContributor` and `org.jetbrains.bazel.ui.widgets.tool.window.utils.LoadedTargetsMouseListener`, specifically `fillWithEligibleActions`. These two are responsible for gutters and available actions.

Issue #7794, opened 2025-07-12
Bazel protobuf generated pb.h files are not indexed


baibinghere:
### Description of the bug:

CLion 2025.1.3 + Bazel 2025.06.24.0.1.
Remote SSH Development Environment.
Bazel installed on `Host`. Imported as bazel project.
Verified the pb.h files are generated in `./bazel-bin/`, `./bazel-out/` folders.
When including headers such as `configs/node_config.pb.h`, CLion IDE complains `'configs/node_config.pb.h' file not found`, while it should be referenced either to the file under `./bazel-bin/config/node_config.pb.h` or to the protobuf definition file.

baibinghere:
Is there any way to find the Bazel plugin logs?

baibinghere:
Hi team, I just tested an older version with CLion 2025.1.3. It is working:

<img width="1917" height="1906" alt="Image" src="https://github.com/user-attachments/assets/550f7af7-23c8-479e-8b14-86c2c044f41e" />

The `*.pb.h` files can be correctly referenced to `bazel-out/*.pb.h`.

grefab:
Same here. Last version that works for me in CLion is v2025.05.27. All after that (beta and stable) fail to index the auto-generated *.pb.h files.

If it helps: After downgrading the plugin, it needs a compile to have the files indexed.

ujohnny:
@grefab so the version 06.10 does not work for you, is it correct?

ujohnny:
I suspect this could be caused by #7701.

grefab:
> [@grefab](https://github.com/grefab) so the version 06.10 does not work for you, is it correct?

Correct. I tried 06.10, 06.24 and 07.08, all fail at indexing *.pb.h. 05.27 works fine.

grefab:
Here you can find a minimal example to reproduce: https://github.com/grefab/7794

ujohnny:
Brief update from our side:

1. This is indeed caused by #7701
2. Here's a 1 line change that "fixes" the issue
```
diff --git a/cpp/src/com/google/idea/blaze/cpp/HeaderRootTrimmerImpl.kt b/cpp/src/com/google/idea/blaze/cpp/HeaderRootTrimmerImpl.kt
--- a/cpp/src/com/google/idea/blaze/cpp/HeaderRootTrimmerImpl.kt	(revision a4292e3d9034fa65e5743095a3650f3b90dfc16c)
[...]

PR #7801, opened 2025-07-17
CLion: add registry key to allow bazel-bin as header search path

LeFrosch:

Issue number: #7794

# Description of this change
Add a registry flag to allow bazel-bin as a header search path again and enable it by default for now until we find a proper solution.

### DIFF
--- a/cpp/src/com/google/idea/blaze/cpp/HeaderRootTrimmerImpl.kt
+++ b/cpp/src/com/google/idea/blaze/cpp/HeaderRootTrimmerImpl.kt
@@ -114,8 +114,8 @@ private fun collectHeaderRoots(
   val result = mutableListOf<Path>()
   for (directory in possibleDirectories) {
     val add = when {
-      // never allow the bazel-bin as a header search path
-      path.isBazelBin(projectData.blazeInfo) && !allowBazelBin -> false
+      // only allow bazel-bin as a header search path when the registry key is set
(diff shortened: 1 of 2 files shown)


cpp/src/com/google/idea/blaze/cpp/HeaderRootTrimmerImpl.kt
```diff
@@ -108,11 +109,13 @@ private fun collectHeaderRoots(
...
   val possibleDirectories = executionRootPathResolver.resolveToIncludeDirectories(path).map(File::toPath)
 
+  val allowBazelBin = Registry.`is`("bazel.cpp.sync.allow.bazel.bin.header.search.path")
+
   val result = mutableListOf<Path>()
   for (directory in possibleDirectories) {
     val add = when {
       // never allow the bazel-bin as a header search path
-      path.isBazelBin(projectData.blazeInfo) -> false
+      path.isBazelBin(projectData.blazeInfo) && !allowBazelBin -> false
```
ujohnny: Flip the condition

ujohnny: Also I believe it should be `isBazelBin && allowBazelBin -> true`, shouldnt it be?

ujohnny (approved):

PR #7805, opened 2025-07-18
Handle user cancellation more gracefully

LeFrosch:
Do not log build aborted BEP messages and do not show JobCancelledExceptions in the UI.


base/src/com/google/idea/blaze/base/sync/SyncPhaseCoordinator.java
```diff
@@ -883,6 +884,9 @@ private static void logSyncError(BlazeContext context, Throwable e) {
       if (cause instanceof SyncCanceledException) {
         return;
       }
+      if (cause instanceof JobCancellationException) {
```
ujohnny, line 887: it's not a 100% correct implementation to skip these exceptions, they have to be rethrown with `Logger.shouldRethrow`, but given the existing code which is almost from the previous decade, I'd say let's meet the invariant here

PR #7810, opened 2025-07-24
Release 24.07.2025

mai93:


CHANGELOG
```diff
@@ -1,3 +1,8 @@
+v2025.07.24
+===========
+* This is the last release for this plugin from Google. Starting from the following release this plugin is not offered by nor affiliated with Google.
```
LeFrosch, line 3: Can we also add: * CLion: allow bazel-bin as header search path again #7801

since this is also part of the release.

LeFrosch (approved):

PR #7812, opened 2025-07-25
Plugin Transfer: add notification

LeFrosch:
This PR adds a notification that is shown once and informs the user about the plugin transfer.


LeFrosch:
@jastice this notification will be displayed in IJwB too. Are you fine with this or would you insist on a custom notification?

jastice:
> @jastice this notification will be displayed in IJwB too. Are you fine with this or would you insist on a custom notification?

Thanks for highlighting. It would be better if the ijwb notification includes a link to the new plugin, but maybe that's not so necessary since we have the PluginReplacement implementation

jastice (approved):

PR #7815, opened 2025-07-31
CLion: release CI

LeFrosch:
Adds a GitHub action that can be manually triggered to create a new release. The action does the following:
- Build clion-oss-latest-stable
- Build clion-oss-oldest-stable
- Create a new release on GitHub including the tag
- Upload the zip files to the market place

The zip files are uploaded with the hidden flag, so they have to be manually promoted on the marketplace after verification.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,96 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release Version'
+        required: true
+        type: string
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        product: [clion-oss-latest-stable, clion-oss-oldest-stable]
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Bazel
+        uses: bazel-contrib/setup-bazel@0.15.0
+        with:
+          bazelisk-cache: true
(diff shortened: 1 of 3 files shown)


LeFrosch:
image

.github/workflows/release.yml
```diff
@@ -0,0 +1,96 @@
...
+          name: clion-oss-latest-stable
+
+      - name: Download Oldest Plugin Zip
+        uses: actions/download-artifact@v4
+        with:
+          name: clion-oss-oldest-stable
+
+      - name: Upload Releases
+        run: |
+          find . -type f -name "*.zip" -exec curl -i --header "Authorization: Bearer ${{ secrets.MARKETPLACE_TOKEN }}" -F pluginId=9554 -F file=@{} -F isHidden=true -F channel=stable https://plugins.jetbrains.com/plugin/uploadPlugin \; 1> /dev/null
```
ujohnny: maybe passing errors to `/dev/null` is unnecessary here as otherwise how we know what is an actual error?

LeFrosch: Shouldn't 1 be stdout? Because the web request returns the entire plugin description and it was hard to see what's going on. That's why I only wanna keep stderr.

ujohnny: True, my mistake. But then can we use silent curl flag or -o /dev/null?

LeFrosch:
I'll go ahead and merge this, but not add the upload keys yet. Let's check who can invoke this and whether the permissions are correct.

Issue #6812, opened 2024-09-28
Corrupted arguments passed to target executable when debugging with GDB


czartur:
### Description of the bug:

When debugging with GDB on CLion 2024.2, the plugin corrupts arguments passed to the target executable. The root cause seems to be improper handling of shell-escaped arguments.

For example:
"--tag=a b" becomes "--tag=a\ b"
"--tag=a,b" becomes "--tag=a\\,b" 

Looking at the [`gdbserver`](https://github.com/bazelbuild/intellij/blob/master/clwb/gdbserver) script, we can observe that shell escaping is enforced on **line 96**: 

```bash
new_args+=("$(printf "%q" "${old_arg}")")
``` 

I am not sure why this is necessary as the script works as expected with:

```bash
new_args+=("${old_arg:q}")
```

### What's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

1. Create a bazel hello world project (e.g. [stage1-tutorial](https://github.com/bazelbuild/examples/tree/main/cpp-tutorial/stage1))
2. Add run configuration and set executable flags to "a,b"
3. Debug and check argv[1] == "a\,b"

### Which Intellij IDE are you using? Please provide the specific version.

CLion 2024.2

### What programming languages and tools are you using? Please provide specific versions.

C++: gcc 11.4, gdb 12.1

### What Bazel plugin version are you using?

2024.09.10.0.1

kekxv:
me too.

kekxv:
@ujohnny Is there a plan to fix it?

clion debug run `args` no work.

<img width="462" height="314" alt="Image" src="https://github.com/user-attachments/assets/6e1b22e1-7722-444e-b749-28a267d60463" />

<img width="399" height="370" alt="Image" src="https://github.com/user-attachments/assets/25f1f052-55c6-4f7f-ae1c-a289fd23f005" />

ujohnny:
@kekxv let us check

kekxv:
@ujohnny @LeFrosch 
hi,
Is there any progress? Is there a way to reproduce this issue?

LeFrosch:
Hi, @kekxv I was not able to reproduce your issue. As far as I could tell args defined in the BUILD file do work. If you still experience this issue, a small project to reproduce this would be really help full.

kekxv:
@LeFrosch 

https://github.com/clangTools/kpipeline

[...]

PR #7816, opened 2025-08-01
CLion: fix gdbserver argument parsing

LeFrosch:

Issue number: #6812

# Description of this change

This PR addresses multiple issues with argument parsing and extends the execution headless test to also debug run configurations.

- removes manual escaping of inferior arguments in the gdbserver script, because [since 2020](https://sourceware.org/pipermail/gdb-patches/2020-May/168439.html?utm_source=chatgpt.com) gdbserver does this automatically
- collect and forward args defined in the BUILD file when using LLDB
- fixes index out of bound exception for empty user defined arguments

### DIFF
--- a/clwb/gdbserver
+++ b/clwb/gdbserver
@@ -35,6 +35,25 @@ if [[ -r ~/.gdbserver_wrapper_options ]]; then
   source ~/.gdbserver_wrapper_options
 fi
 
+gdbserver_wrapper::version_lt() {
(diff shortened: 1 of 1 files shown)


kekxv:
@LeFrosch When is the release planned?

ujohnny:
So if we understand the gdbserver changes right, this change could break gdbserver < 10.1, so probably we would like to put a version detection and use the previous logic for old gdbserver versions

ujohnny (approved):

PR #7865, opened 2025-09-05
Update .manged.bazelproject for use with the new Bazel plugin

LeFrosch:
This PR allows this project to be imported by the new Bazel plugin.

### DIFF
--- a/.managed.bazelproject
+++ b/.managed.bazelproject
@@ -1,28 +1,20 @@
 directories:
   .
-  -aswb/testdata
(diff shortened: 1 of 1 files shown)


ujohnny:
this would break the use of old plugin, I'd propose to apply this change while keeping existing version as old.managed.bazelproject

LeFrosch:
This file was only for query sync `clwb/.bazelproject` can still be used with the old plugin.

PR #7874, opened 2025-09-11
CLion: make sysroot execroot relative

LeFrosch:
Previously the sysroot was always resolved as workspace relative. Which is not correct in all cases. This PR resolves the sysroot from the execution root using a custom CoptsProcessor. 

Also removes the legacy BlazeCompilerFlagsProcessor.


cpp/src/com/google/idea/blaze/cpp/copts/CoptsSysrootProcessor.kt
```diff
@@ -0,0 +1,50 @@
...
+      else -> ImmutableList.of()
+    }
+  }
+
+  override fun apply(
+    value: String,
+    sink: CompilerSpecificSwitchBuilder,
+    resolver: ExecutionRootPathResolver
+  ) {
+    val file = File(value)
```
LeFrosch: Use NIO

cpp/src/com/google/idea/blaze/cpp/copts/CoptsSysrootProcessor.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.jetbrains.cidr.lang.workspace.compiler.GCCCompilerKind
+import com.jetbrains.cidr.lang.workspace.compiler.OCCompilerKind
+import java.io.File
+
+class CoptsSysrootProcessor : CoptsProcessor.Transform() {
+
+  override fun flags(kind: OCCompilerKind): ImmutableList<String> {
+    return when (kind) {
+      GCCCompilerKind, ClangCompilerKind -> ImmutableList.of("--sysroot")
+      // TODO: add support for MSVC
```
LeFrosch: Remove this.

PR #7880, opened 2025-09-15
[AOSP-pick] Inline helper and overloaded methods

LeFrosch:
Cherry pick AOSP commit [27726d3b009cf97835aff9247fc78c80a2effa68](https://cs.android.com/android-studio/platform/tools/adt/idea/+/27726d3b009cf97835aff9247fc78c80a2effa68).

They make the core code harder to understand.

Helper methods should be added to static-only helper classes or
implemented as Kotlin extension functions.

Bug: 409737053
Test: n/a
Change-Id: Icb8a4efd8a89f0f4fe168a89ece1c33e8bd39c56

AOSP: 27726d3b009cf97835aff9247fc78c80a2effa68


LeFrosch:
Qsync only, no need to pick this after all.

PR #7895, opened 2025-09-22
Fix project wrongfully opened as query sync

LeFrosch:
I don't think we'll ever support query sync properly for C/CPP and users are not complaining about aspect sync either, therefore let's just drop it.


ujohnny:
and maybe later remove all qsync items from the code if we want

PR #7897, opened 2025-09-22
Fix aspect expand_location bug

LeFrosch:
Adapts `expand_make_variables` to access the correct attributes when called from inside the aspect. The original implementation copied from the Bazel sources was called from a rule.

Also removes the duplicated `ctx.expand_location` call and only keeps the correctly setup call in `expand_make_variables`.


LeFrosch:
TODO: let's split this PR into the actual fix and the clean up :eyes:

ujohnny:
It would be nice to clarify the description here as it says that `Pass data dependencies to ctx.expand_location` while call to `ctx.expand_location` is removed (though the real call is performed later).

LeFrosch:
Clean up moved to: #7905

PR #7834, opened 2025-08-11
CLion: include cache

LeFrosch:
The PR adds a cache for generated headers and `_virtual_include` directories of a target to avoid placing header search paths into the execution root. 

Header search paths in the execution root can cause unnecessary files to be loaded into the VFS, since indexing code might recursively iterate over all files in these directories.

PR #7902, opened 2025-09-26
Add support for rules_foreign_cc

LeFrosch:
adds dedicated support for rules_foreign_cc based on #7834


LeFrosch:
merged into: #7834

PR #7925, opened 2025-10-06
Include Cache: update cquery scripts

LeFrosch:
This PR updates the cquery scripts to handle some edge cases.

Extracted from: #7834

### DIFF
--- a/tools/cquery/cc_info.bzl
+++ b/tools/cquery/cc_info.bzl
@@ -3,12 +3,16 @@
 
 CC_INFO = "CcInfo"
 
+def file_list_to_paths(files):
+    return [getattr(f, "path", f) for f in files]
+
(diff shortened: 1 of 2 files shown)


tools/cquery/cc_info.bzl
```diff
@@ -40,7 +44,7 @@ def format_cc_info(info):
 def format(target):
     buffer = "%s: " % target.label
 
-    if CC_INFO in providers(target):
+    if CC_INFO in (providers(target) or []):
```
ujohnny, line 47: in what case we can have `providers(target)` as None?

LeFrosch: sorry idk, but in some cases it is and this keeps the cquery from failing

PR #7926, opened 2025-10-06
Transition Support: add configuration id to TargetKey

LeFrosch:
In order to support transitions we need to be able to differentiate targets with different configurations. This PR adds the configuration id to the TargetKey class which is different after every transition.


LeFrosch:
Configuration id will only be exposed to aspects starting with Bazel 8.5: https://github.com/bazelbuild/bazel/pull/27128

LeFrosch:
resolved merge conflicts

PR #7930, opened 2025-10-10
Add BazelShowTargetInfoAction

LeFrosch:
This PR adds another debug action (inspired by show compiler info) to dump proto information we have for a target.

### DIFF
--- a/base/src/com/google/idea/blaze/base/actions/BazelShowTargetInfoAction.kt
+++ b/base/src/com/google/idea/blaze/base/actions/BazelShowTargetInfoAction.kt
@@ -16,6 +16,7 @@
 package com.google.idea.blaze.base.actions
 
 import com.google.idea.blaze.base.ideinfo.TargetIdeInfo
+import com.google.idea.blaze.base.settings.Blaze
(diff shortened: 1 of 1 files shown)


base/src/com/google/idea/blaze/base/actions/BazelShowTargetInfoAction.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.intellij.openapi.project.DumbAwareAction
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.ui.popup.JBPopupFactory
+import com.intellij.openapi.ui.popup.PopupStep
+import com.intellij.openapi.ui.popup.util.BaseListPopupStep
+import com.intellij.testFramework.LightVirtualFile
+
+private val LOG = Logger.getInstance(BazelShowTargetInfoAction::class.java)
+
+class BazelShowTargetInfoAction : DumbAwareAction() {
```
ujohnny, line 36: Let’s make the action disabled for non-bazel projects in the overriden update

PR #7951, opened 2025-10-21
Add switch project view action

LeFrosch:
This PR adds an action to load a different project view by right clicking it in the file explorer, similar to what the new plugin offers.

### DIFF
--- a/base/src/META-INF/blaze-base.xml
+++ b/base/src/META-INF/blaze-base.xml
@@ -168,9 +168,9 @@
     <action id="Bazel.ShowTargetInfo"
       class="com.google.idea.blaze.base.actions.BazelShowTargetInfoAction"
       text="Bazel show target info"/>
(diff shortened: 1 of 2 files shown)


LeFrosch:
Depends on: #7952

LeFrosch:
Question for review: Do we need a read action for Action.update since we access a VirtualFile?

ujohnny:
lgtm

ujohnny:
> Question for review: Do we need a read action for Action.update since we access a VirtualFile?

Per documentation and considering as an example `VirtualFileSystemEntry#getName` is not thread-safe, the answer is yes. Though global lock introduces many performance implications and in the ideal world any operation you perform in update should be given an immutable snapshot of data which should not require any locks. Most likely code will work fine and the potential issues might arise on shutdown (ie update & ijplatform shutdown race condition) and so one.

PR #7952, opened 2025-10-21
chore(clean up): Remove support for query sync specific actions

LeFrosch:
Since we deprecated query sync, explicit handling for query sync specific actions is no longer required.


ujohnny:
could you please drop [docs/querysync.md](https://github.com/bazelbuild/intellij/blob/master/docs/querysync.md) in a separate PR as well?

PR #7953, opened 2025-10-21
chore(clean up): Remove support for different project view files

LeFrosch:
Since we no longer support ASwB nor Blaze this logic is no longer required. Also removes the deprecated `FileTypeFactory` for project view files.


ujohnny:
Should be fine and hope there are no external users with `.blazeproject` for some historical reasons.

ujohnny (approved):

PR #7956, opened 2025-10-22
Update headless tests for Bazel 9

LeFrosch:
Fixes test issues with Bazel 9.

### DIFF
--- a/clwb/tests/headlesstests/com/google/idea/blaze/clwb/LibCppTest.java
+++ b/clwb/tests/headlesstests/com/google/idea/blaze/clwb/LibCppTest.java
@@ -40,11 +40,22 @@ public class LibCppTest extends ClwbHeadlessTestCase {
 
   @Override
   protected ProjectViewBuilder projectViewText(BazelVersion version) {
-    return super.projectViewText(version).addBuildFlag(
-        // set the compiler to clang, only required for linux
-        "--action_env=CC=/usr/bin/clang",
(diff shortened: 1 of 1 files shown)


tools/cquery/cc_toolchain.bzl
```diff
@@ -1,7 +1,18 @@
 # cquery script to dump the CcToolchainInfo of a target
 # usage: bazel cquery '<target>' --output=starlark --starlark:file=tools/cquery/cc_toolchain.bzl
 
-CC_TOOLCHAIN_INFO = "@@_builtins//:common/cc/cc_toolchain_info.bzl%CcToolchainInfo"
+BUILTIN_CC_TOOLCHAIN_INFO = "@@_builtins//:common/cc/cc_toolchain_info.bzl%CcToolchainInfo"
+RULESET_CC_TOOLCHAIN_INFO = "@@rules_cc+//cc/private/rules_impl:cc_toolchain_info.bzl%CcToolchainInfo"
```
ujohnny, line 5: I wonder if it's intentionally private there. Shall we check with https://github.com/bazelbuild/rules_cc that it would be nice to have providers in non-private area for stable targets introspection.

LeFrosch: I guess normally you would access the provider with `cc_common.CcToolchainInfo`. But since we cannot load anything in cquery files the hardcoded string is required here.

PR #7967, opened 2025-10-28
Include Cache: fix cache removes external prefix 

LeFrosch:
This PR adds support for external targets inside the header cache.


cpp/src/com/google/idea/blaze/cpp/sync/HeaderCacheService.kt
```diff
@@ -169,6 +169,19 @@ class HeaderCacheService(private val project: Project) {
   private fun isInBazelBin(location: ArtifactLocation): Boolean {
     return location.rootPath().isNotBlank() && isInBazelBin(Path.of(location.rootPath()))
   }
+
+  private fun resolveCachePath(cacheDirectory: Path, location: ArtifactLocation): Path {
+    val root = Path.of(location.rootPath())
+
+    // if the path has an external prefix, this needs to map into the cache as well
+    val cacheRoot = if (root.nameCount > 3) {
```
ujohnny, line 178: let's add a comment how 3 is calculated here

PR #8000, opened 2025-11-21
Prepartation for Bazel 9: add load statements to intellij_project_sdk BUILD files

LeFrosch:
Since autoload is disabled in Bazel 9 we need load statements here.


LeFrosch:
Preparation for: #8004

ujohnny (approved):

PR #8001, opened 2025-11-21
Prepartation for Bazel 9: update cc_info cquery script

LeFrosch:
In Bazel 9 the Provider name changes since it is no longer builtin.


LeFrosch:
Preparation for: #8004

ujohnny (approved):

PR #8002, opened 2025-11-21
Prepartation for Bazel 9: add templating for CcInfo and cc_common

LeFrosch:
Since Bazel 9 disables autoloads we can no longer use the builtin symbols and have to add loads. However, to preserver backwards compatibility this is done vie templating if rules_cc is found in the module graph.

### DIFF
--- a/aspect/cc_info.bzl
+++ b/aspect/cc_info.bzl
@@ -1,24 +1,16 @@
 # TEMPLATE-INCLUDE-BEGIN
 ## #if( $isCcEnabled == "true" )
-##load("@rules_cc//cc:defs.bzl", _CcInfo = "CcInfo", _cc_common = "cc_common")
+##load("@rules_cc//cc:defs.bzl", "CcInfo", "cc_common")
 ## #end
 ##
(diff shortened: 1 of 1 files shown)


LeFrosch:
Preparation for: #8004

aspect/cc_info.bzl
```diff
@@ -1,7 +1,24 @@
...
+##CcInfoCompat = _CcInfo
+##cc_common_compat = _cc_common
+## #else
+##CcInfoCompat = CcInfo
+##cc_common_compat = cc_common
+## #end
 # TEMPLATE-INCLUDE-END
+
+# TEMPLATE-IGNORE-BEGIN
+load("@rules_cc//cc:defs.bzl", _CcInfo = "CcInfo", _cc_common = "cc_common")
```
ujohnny: what is the purpose of this remapping?

LeFrosch: Actually no reason anymore, looks like an artifact from a previous iteration.

ujohnny (approved):

PR #8003, opened 2025-11-21
Prepartation for Bazel 9: get xcode version from @bazel_tools//tools/cpp:host_xcodes

LeFrosch:
The previous target (@bazel_tools//tools/osx:current_xcode_config) which we used in the cquery no longer exists in Bazel 9. However, the actual target which provides the version (@bazel_tools//tools/cpp:host_xcodes) exists in Bazel 9 and all currently supported versions (6-7).

### DIFF
--- a/cpp/tests/utils/com/google/idea/blaze/cpp/MockXcodeSettingsProvider.java
+++ b/cpp/tests/utils/com/google/idea/blaze/cpp/MockXcodeSettingsProvider.java
@@ -16,6 +16,7 @@
 package com.google.idea.blaze.cpp;
 
 import com.google.idea.blaze.base.scope.BlazeContext;
+import com.google.idea.blaze.base.model.BlazeProjectData;
 import com.intellij.openapi.project.Project;
 import java.util.Optional;
 
@@ -28,7 +29,7 @@ public class MockXcodeSettingsProvider implements XcodeCompilerSettingsProvider 
   }
 
   @Override
-  public Optional<XcodeCompilerSettings> fromContext(BlazeContext context, Project project)
+  public Optional<XcodeCompilerSettings> fromContext(BlazeContext context, Project project, BlazeProjectData projectData)
       throws XcodeCompilerSettingsException {
     if (this.settings != null) {
       return Optional.of(this.settings);


LeFrosch:
Preparation for: #8004

cpp/src/com/google/idea/blaze/cpp/XcodeCompilerSettingsProviderImpl.java
```diff
@@ -52,7 +52,7 @@ public class XcodeCompilerSettingsProviderImpl implements XcodeCompilerSettingsP
   private static final String[] QUERY_XCODE_VERSION_SCRIPT_LINES = new String[]{
       "#!/bin/bash",
       "__BAZEL_BIN__ cquery \\",
-      " 'deps(\"@bazel_tools//tools/osx:current_xcode_config\")' \\",
+      " 'deps(\"@bazel_tools//tools/cpp:host_xcodes\")' \\",
```
ujohnny: not sure, but I have a feeling that item on the left has current config, while rhs lists all
https://bazel.build/reference/command-line-reference#flag--xcode_version_config

LeFrosch: So apparently host_xcodes is the default value for `--xcode_version_config`: https://github.com/bazelbuild/bazel/blob/f57b39513a30d0ebd46e8888b2445313b3198de1/tools/cpp/BUILD.tools#L125-L128

Where as current_xcode_config was the current value of `--xcode_version_config`: https://github.com/bazelbuild/bazel/blob/3b1ed466d08107049dc1d232f43a7018997a2505/src/main/starlark/builtins_bzl/common/xcode/xcode_config_alias.bzl

But this rule has been removed in https://github.com/bazelbuild/bazel/commit/873b036afbf77ede4588aeaa95b119fb486accef and is no longer in the builtins. Idk where it moved to. 

As shown in the new aspect, it is possible to move the xcode version collection to the aspect finally. I could port this code and we could use this for Bazel 8+.

ujohnny: Don't we need some short term solution here as bazel 9 release is kinda close and full migration to the new aspect is a bit risky? Or is it ok for us to depend on aspect change here?

ujohnny: Though I'm a bit puzzled about the real need in this code. We have it to set `DEVELOPER_DIR` and `SDKROOT` for compiler script, but we never had such a need in CMake support. Is it so that we rely on default xcode selection, while bazel can override it?

ujohnny: Maybe this code is incorrectly placed, this has to be somewhat related closer to toolchain info rather than just "random" Xcode-related target query.

LeFrosch: Yes, I agree. By porting I meant only taking the code I wrote for xcode info collection and adding it to the old aspect. If this is possible.

LeFrosch:
As agreed, we'll proceed as follows. Only use the `@bazel_tools//tools/cpp:host_xcodes` for Bazel 9+ and keep the previous logic for older Bazel versions. While `host_xcodes` only provides the default it should be fine for most cases. 

With the new aspect we'll receive the correct xcode version from the aspect and can deprecate this entire cquery setup.

PR #8010, opened 2025-11-24
Add recheck action

LeFrosch:
Adds an action to rerun CI by commenting /recheck on a PR. The action will add an empty commit.


LeFrosch:
So apparently the default GH token does not have enough permissions to push a commit to the fork. We can either add a personal access token of mine to the bot or revert the this PR :(

PR #8013, opened 2025-11-25
Updated intellij_plugin_zip rule

LeFrosch:
Clean up of the `intellij_plugin_zip` rule (and `intellij_plugin_debug_target` rule). They are only used by CLwB, so no backwards comparability required. This PR removes a lot of unused Google features and manages data files as runfiles, this makes loading them in tests easier.

This is change is not just for the fun of deleting code and writing new rules, but also in preparation for bundling more tools (like the REPL or execlog parser).


ujohnny (approved):

LeFrosch:
@gemini-code-assist review

PR #8004, opened 2025-11-21
Updated project to Bazel 9

LeFrosch:
Updates the project build to Bazel 9 and adds Bazel 9 to the headless tests. Since the aspect tests are run by the project's Bazel version, they have to be updated for Bazel 9 as well.

PR #8014, opened 2025-11-26
Prepartation for Bazel 9: add headless tests

LeFrosch:
Adds headless tests for Bazel 9 (rc3). However, this requires removing some of the Bazel 6 headless tests since the test fixtures cannot be compatible across these many Bazel versions.


LeFrosch:
Extracted from: #8004

ujohnny (approved):

PR #8029, opened 2025-12-01
Prevent freezes in AspectSyncSourceToTargetMap.getSourceToTargetMap

ujohnny:
getSourceToTargetMap might be called under read action and it's
a quite heavy operation. we can bring cancellation checks into it,
but

a. it might be called under non-cancellable read action
b. it might take tens of seconds to compute, so no readaction
can survive this long

so if we're called under readaction, we just return null and schedule
background computation, and hope we will be called some time later


base/src/com/google/idea/blaze/base/sync/SyncCache.kt
```diff
@@ -0,0 +1,120 @@
...
+
+        if (result is Acquired<Pair<Boolean, Any?>> && result.value.first) {
+            @Suppress("UNCHECKED_CAST")
+            return result.value.second as T?
+        } else {
+            // in case there's a computation running for the same key, then this
+            // is no-op, otherwise this is
+            scope.launch {
+                get(key, computable)
+            }
```
LeFrosch: Will this swallow exceptions?

base/src/com/google/idea/blaze/base/sync/LockCriticalSection.kt
```diff
@@ -0,0 +1,55 @@
+package com.google.idea.blaze.base.sync
```
LeFrosch: Missing copy right

base/src/com/google/idea/blaze/base/sync/LockCriticalSection.kt
```diff
@@ -0,0 +1,55 @@
...
+ * Encapsulates the data into [hidden] and allows retriable access to it
+ * with cancellation checks
+ */
+class LockCriticalSection<Hidden : Any>(private val hidden: Hidden) {
+    private val lock = Mutex()
+
+    sealed class TryLockResult<out T> {
+        data class Acquired<T>(val value: T) : TryLockResult<T>()
+        data object NotAcquired : TryLockResult<Nothing>()
+    }
```
LeFrosch: Isn't this basically an Optional<T>?

base/src/com/google/idea/blaze/base/sync/SyncCache.kt
```diff
@@ -0,0 +1,120 @@
...
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.project.Project
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+
+/** Computes a cache on the project data.  */
+@Service(Service.Level.PROJECT)
+class SyncCache(private val project: Project, private val scope: CoroutineScope) {
+    /** Computes a value based on the sync project data.  */
```
LeFrosch: Wrong indentation of 4 spaces instead of 2

base/src/com/google/idea/blaze/base/sync/SyncCache.kt
```diff
@@ -0,0 +1,120 @@
...
+    interface SyncCacheComputable<T> {
+        fun compute(project: Project, projectData: BlazeProjectData): T?
+    }
+
+    private data class Data(val cache: MutableMap<Any, Any?>)
```
LeFrosch: What's the benefit of wrapping the Map in the data class?

base/src/com/google/idea/blaze/base/sync/LockCriticalSection.kt
```diff
@@ -0,0 +1,55 @@
...
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.milliseconds
+
+/**
+ * Encapsulates the data into [hidden] and allows retriable access to it
+ * with cancellation checks
+ */
+class LockCriticalSection<Hidden : Any>(private val hidden: Hidden) {
```
LeFrosch: Since we only use this class for the SyncCache and it's only ever instantiated for a HashMap, we could drop some of the generic parameters in favor of specialization?

LeFrosch (changes requested):
Overall looks good just a bunch of nit picks. And I changed my mind, this is probably more reliable then my `Deferred` idea.

base/src/com/google/idea/blaze/base/sync/SyncCache.kt
ujohnny: interesting as I use https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml, but it does not have kotlin settings, moreover there's no kotlin styleguide in this repo at all

LeFrosch: Oh true, I just stuck with the two space indentation from the Java code.

base/src/com/google/idea/blaze/base/sync/SyncCache.kt
ujohnny: no this won't, they will be thrown up to the dispatcher and reported as IDE error

base/src/com/google/idea/blaze/base/sync/LockCriticalSection.kt
ujohnny: It, actually, is, but this is more typed thing to make code that uses it more readable

base/src/com/google/idea/blaze/base/sync/SyncCache.kt
ujohnny: done

base/src/com/google/idea/blaze/base/sync/LockCriticalSection.kt
ujohnny: done

base/src/com/google/idea/blaze/base/sync/SyncCache.kt
ujohnny: I'd like to hide the object beyond lock, so I won't fail with trylock and so on in the code that accesses it

base/src/com/google/idea/blaze/base/sync/LockCriticalSection.kt
ujohnny: I moved the class to `util` package, so it can be reused if there's a similar case somewhere else

base/src/com/google/idea/blaze/base/sync/SyncCache.kt
LeFrosch: Which I guess might even be an improvement over the current situation, since if I understand it correctly, null is handled more gracefully in the callee.

LeFrosch:
Would we like to add background progress for calculating this mapping. Because, now if one invokes an actin like `AddSourceToProjectAction` while the mapping is being calculated the action will fail (or rather do nothing). It might be nice to have some visual indication for this calculation?

ujohnny:
> Would we like to add background progress for calculating this mapping. Because, now if one invokes an actin like `AddSourceToProjectAction` while the mapping is being calculated the action will fail (or rather do nothing). It might be nice to have some visual indication for this calculation?

I guess the action will be disabled then because `sourceCoveredByProjectViewTargets` will return null and then in `updateForBlazeProject` it will be set to false since `description` will be null.

LeFrosch:
Okay, this should be fine ^^

ujohnny:
So I double checked the idea I had, since there's now `SourceToTargetMap.getInstance(project).init()` in `ProjectTargetManagerImpl.TargetSyncListener#afterSync`, then sync will not finish unless the map is available, so there will be a progress reported, though it could be not computing the data, but waiting for the data to be computed on the background thread without any progress being reported. I believe it's ok here to have less items exposed outside.

PR #8031, opened 2025-12-03
chore(clean up): remove query sync

LeFrosch:
Query sync is deprecated for some time now and removing it will make refactors easier in future. I did this some time ago to enable a refactor, but I forgot for which exactly.


LeFrosch:
Fixes build issues for: #8030

PR #8043, opened 2025-12-08
chore(clean up): add common base for debug actions

LeFrosch:
Adds a common base for debug actions to make it easier to add new debug actions in future.


LeFrosch:
@gemini-code-assist review

PR #8044, opened 2025-12-08
Add proxy support

LeFrosch:
This PR adds support for warding CLion's proxy settings to Bazel. However, this is disabled by default and can be enabled in the settings: Bazel Settings | Forward proxy settings


LeFrosch:
@gemini-code-assist review

PR #8046, opened 2025-12-08
drop fdebug-compilation-dir for macos debug

ujohnny:
Use of fdebug-compilation-dir in macos debug runs
leads to 100% remote cache miss because the path for
compilation dir is valid for the current user only.
The change reworks the original PR #4403 and provides a
more correct fix for #2101 by launching lldb in the
proper working directory and using relative paths to set
breakpoints

### DIFF
--- a/clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java
+++ b/clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java
@@ -48,7 +48,6 @@ import com.google.idea.blaze.base.settings.Blaze;
 import com.google.idea.blaze.base.settings.BlazeUserSettings;
 import com.google.idea.blaze.base.settings.BuildSystemName;
 import com.google.idea.blaze.base.sync.data.BlazeProjectDataManager;
-import com.google.idea.blaze.clwb.ToolchainUtils;
 import com.google.idea.blaze.cpp.CppBlazeRules;
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configuration.EnvironmentVariablesData;
@@ -78,6 +77,7 @@ import com.jetbrains.cidr.execution.debugger.remote.CidrRemoteDebugParameters;
 import com.jetbrains.cidr.execution.debugger.remote.CidrRemotePathMapping;
 import com.jetbrains.cidr.execution.testing.google.CidrGoogleTestConsoleProperties;
 import java.io.File;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
@@ -289,7 +289,17 @@ public final class BlazeCidrLauncher extends CidrLauncher {
 
       final DebuggerDriverConfiguration debuggerDriver;
       if (debuggerKind == BlazeDebuggerKind.BUNDLED_LLDB) {
-        debuggerDriver = new BlazeLLDBDriverConfiguration(project, workspaceRoot.directory().toPath());
+        final var projectData = BlazeProjectDataManager.getInstance(project).getBlazeProjectData();
+        final Path executionRoot;
+        if (projectData == null) {
+          executionRoot = workspaceRoot.directory().toPath();
+        } else {
+          executionRoot = projectData.getBlazeInfo().getExecutionRoot().toPath();
+        }
+
+        // the working directory must be the execroot to make breakpoints work
+        debuggerDriver = new BlazeLLDBDriverConfiguration(project, executionRoot,
+                workspaceRoot.directory().toPath());
       } else {
         final var startupCommands = getGdbStartupCommands(workspaceRootDirectory);
         debuggerDriver = new BlazeGDBDriverConfiguration(project, startupCommands, workspaceRoot);
--- a/clwb/src/com/google/idea/blaze/clwb/run/BlazeLLDBDriverConfiguration.java
+++ b/clwb/src/com/google/idea/blaze/clwb/run/BlazeLLDBDriverConfiguration.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2025 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.google.idea.blaze.clwb.run;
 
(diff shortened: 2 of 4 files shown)


clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java
```diff
@@ -290,7 +289,20 @@ public CidrDebugProcess createDebugProcess(CommandLineState state, XDebugSession
 
       final DebuggerDriverConfiguration debuggerDriver;
       if (debuggerKind == BlazeDebuggerKind.BUNDLED_LLDB) {
-        debuggerDriver = new BlazeLLDBDriverConfiguration(project, workspaceRoot.directory().toPath());
+        // it would be better to get execroot from bazel info
+
+        var execPath = runner.executableToDebug.toPath();
+        while (execPath.getParent() != null && !execPath.getFileName().toString().equals("_main")) {
+          execPath = execPath.getParent();
+        }
```
LeFrosch: Isn't this the execution root?
```
$ bazel info execution_root
/home/daniel_brauner/.cache/bazel/_bazel_daniel_brauner/bccb7c79f93a001e60be70bbf2beed3e/execroot/_main
```

So couldn't we use:
```java
final var projectData = BlazeProjectDataManager.getInstance(project).getBlazeProjectData();
projectData.getBlazeInfo().getExecutionRoot();
```

ujohnny: it is. we can use `BlazeProjectDataManager` but there's a little chance that when we do rebuild in before run task, the executable might be built in a different `execution_root`, though this is quite unlikely. I'll update the review

ujohnny:
@alexander-doroshko most likely you'll have to implement something similar to support LLDB debugging in your project

clwb/src/com/google/idea/blaze/clwb/run/BlazeLLDBDriverConfiguration.java
```diff
@@ -7,23 +23,41 @@
...
+        if (absolutePath == null) {
+            return null;
+        }
+
+        // we need to relativize the path to the project root for breakpoints.
+        // LLDB started from execroot can set only relative breakpoints, and for
+        // absolute it requires path mapping to be set, which is quite hard to
+        // configure properly because the source location is dynamic inside the build sandbox
+        return projectRoot.relativize(Path.of(absolutePath)).toString();
     }
```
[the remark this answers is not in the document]
ujohnny: absolute paths do not work anyway, so adding such if won't make program work, I guess we can omit this advice

clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java
ujohnny: done

clwb/src/com/google/idea/blaze/clwb/run/BlazeLLDBDriverConfiguration.java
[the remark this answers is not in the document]
LeFrosch: Why don't absolute paths work? Paths from non-hermetic toolchains might really be absolute, or does LLDB always need relative paths?

clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java
```diff
@@ -289,7 +289,17 @@ public CidrDebugProcess createDebugProcess(CommandLineState state, XDebugSession
 
       final DebuggerDriverConfiguration debuggerDriver;
       if (debuggerKind == BlazeDebuggerKind.BUNDLED_LLDB) {
-        debuggerDriver = new BlazeLLDBDriverConfiguration(project, workspaceRoot.directory().toPath());
+        final var projectData = BlazeProjectDataManager.getInstance(project).getBlazeProjectData();
+        final Path executionRoot;
+        if (projectData == null) {
+          executionRoot = workspaceRoot.directory().toPath();
```
LeFrosch, line 295: If this branch is take the project is already in a really weird state, so good enough heuristic ig 🙃

clwb/src/com/google/idea/blaze/clwb/run/BlazeLLDBDriverConfiguration.java
```diff
@@ -7,23 +23,41 @@
...
+    @Override
+    public String convertToProjectModelPath(@Nullable String absolutePath) {
+        if (absolutePath == null) {
+            return null;
+        }
+
+        // we need to relativize the path to the project root for breakpoints.
+        // LLDB started from execroot can set only relative breakpoints, and for
+        // absolute it requires path mapping to be set, which is quite hard to
+        // configure properly because the source location is dynamic inside the build sandbox
```
LeFrosch, line 51: So we really do need relative paths, always?

ujohnny: yes, LLDB does not accept absolute paths for breakpoints because it requires them as `/private/var/tmp/_bazel_enovozhilov/57684a2d51f77638ccc35abaf86393a0/sandbox/darwin-sandbox/10/execroot/_main/src/hello_world.cc` for example, but mapping from source directory to sandbox is pretty challenging, which would require lldb/python script that calculated this based on `image lookup -vn main` output

clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java
ujohnny: Probably it would make sense even to throw here, but I'm not sure (because no breakpoints will work)

ujohnny:
no tests since our CI does not allow debugging on macOS

clwb/src/com/google/idea/blaze/clwb/run/BlazeLLDBDriverConfiguration.java
LeFrosch: I was more thinking for paths out side of the sandbox, like from non-hermetic toolchains. They could just be something like /usr/include/stdlib.h?

clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrLauncher.java
LeFrosch: Agree, throwing is probably a bad idea here.

clwb/src/com/google/idea/blaze/clwb/run/BlazeLLDBDriverConfiguration.java
ujohnny: Good point. You're most likely correct here, same as gemini wrote earlier that we might be relativizing too much. Probably this is worth doing https://github.com/bazelbuild/intellij/pull/8046#discussion_r2604391551

ujohnny:
@LeFrosch I also realized that the change makes breakpoints impossible in out-of-project files, so libraries/etc. But maybe this is kinda expected, though for those we will need to have a mapping from `external` to sandbox

LeFrosch:
> mapping from external to sandbox

@ujohnny I think this might be really difficult since sandbox paths are not predictable by nature. But I also have no idea what is required from that mapping anyway.

PR #8052, opened 2025-12-10
Fix unexpected cancellations while selecting test targets

ujohnny:
After #8029 run test gutters almost always display "Nothing here" Now SyncCache checks for cancellations and `ReverseDependencyMap.get` is considered as canceled, because `ProgressManager.checkCancelled` ensures that Job is active, but it's not because of the following:

`PendingAsyncTestContext.fromTargetFuture` used directExecutor, which might reuse / reuses future thread, but future is executed in coroutine scope converted to future, so scope's Job is completed when future is done. We should avoid using directExecutor as mentioned in Futures documentation


FINDMEAI:
nicee

PR #8055, opened 2025-12-10
Add RadBazelProjectFallbackConfigDisabler

LeFrosch:
We found that providing fallback configuration to backend leads to recursive watch roots being set up for include paths mentioned in the fallback config. We know this has very negative performance implications because of greediness in VfsWarmQueue. This extension point allows us to disable fallback configurations for Bazel projects. (commit message from @ujohnny)

However, since the extension point was introduced in 2025.2.5, it is not save to simply instantiate the implementation class and we need to do some manual registration magic.

### DIFF
--- a/clwb/src/com/google/idea/blaze/clwb/radler/RadBazelProjectFallbackConfigDisabler.kt
+++ b/clwb/src/com/google/idea/blaze/clwb/radler/RadBazelProjectFallbackConfigDisabler.kt
@@ -18,8 +18,11 @@ package com.google.idea.blaze.clwb.radler
 import com.google.idea.blaze.base.settings.Blaze
 import com.google.idea.sdkcompat.radler.RadProjectFallbackConfigDisabler
 import com.intellij.openapi.project.Project
+import com.intellij.openapi.util.registry.Registry
 
 class RadBazelProjectFallbackConfigDisabler : RadProjectFallbackConfigDisabler {
 
-  override suspend fun disableFallbackConfigForProject(project: Project): Boolean = Blaze.isBlazeProject(project)
+  override suspend fun disableFallbackConfigForProject(project: Project): Boolean {
+    return Registry.`is`("bazel.clwb.radler.fallback.configurations.disabled") && Blaze.isBlazeProject(project)
+  }
 }
--- a/clwb/src/com/google/idea/blaze/clwb/radler/optional-plugin.xml
+++ b/clwb/src/com/google/idea/blaze/clwb/radler/optional-plugin.xml
@@ -22,6 +22,10 @@
     <SyncStatusContributor implementation="com.google.idea.blaze.clwb.radler.RadSyncStatusContributor"/>
   </extensions>
 
+  <extensions defaultExtensionNs="com.intellij">
+    <registryKey defaultValue="true" description="Disable fallback resolve configurations for radler." key="bazel.clwb.radler.fallback.configurations.disabled"/>
+  </extensions>
+
   <applicationListeners>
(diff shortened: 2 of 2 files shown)


clwb/src/com/google/idea/blaze/clwb/radler/RadBazelProjectFallbackConfigRegister.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import com.google.idea.blaze.base.util.pluginApplicationDisposable
+import com.intellij.ide.AppLifecycleListener
+import com.intellij.openapi.extensions.Extensions
+
+private const val EXTENSION_POINT_NAME = "com.intellij.rider.cpp.core.projectModel.radProjectFallbackConfigDisabler"
+private const val IMPLEMENTATION_CLASS = "com.google.idea.blaze.clwb.radler.RadBazelProjectFallbackConfigDisabler"
+
+/**
+ * This is required for backwards compatibility with CLion 2025.2.4 and earlier. Since the extension
```
ujohnny, line 26: Sad, I was afraid that we have to do this and it turned out we have. Sorry for this

clwb/src/com/google/idea/blaze/clwb/radler/RadBazelProjectFallbackConfigRegister.kt
```diff
@@ -0,0 +1,41 @@
...
+ * point was introduced in 2025.2.5, it is not save to simply instantiate the implementation class.
+ */
+class RadBazelProjectFallbackConfigRegister : AppLifecycleListener {
```
ujohnny, line 31: let's leave #api252 here so once we're cleaning it up we will update this code

clwb/src/com/google/idea/blaze/clwb/radler/RadBazelProjectFallbackConfigRegister.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  @Suppress("UnstableApiUsage")
+  override fun appStarted() {
+    val extensionPoint = Extensions.getRootArea().getExtensionPointIfRegistered<Any>(EXTENSION_POINT_NAME) ?: return
+
+    // if the extension point is registered, we can safely instantiate the implementation
+    val implementation = Class.forName(IMPLEMENTATION_CLASS).getConstructor().newInstance()
+
+    extensionPoint.registerExtension(implementation, pluginApplicationDisposable())
```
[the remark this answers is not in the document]
LeFrosch: Like the comment says, if the extension point is there we can safely instantiate the class and exceptions here do not seem to affect overall plugin loading.

clwb/src/com/google/idea/blaze/clwb/radler/RadBazelProjectFallbackConfigRegister.kt
LeFrosch: done

clwb/src/com/google/idea/blaze/clwb/radler/RadBazelProjectFallbackConfigRegister.kt
LeFrosch: No worries, and it's not like it's the first time we had to this 🙃

ujohnny:
@LeFrosch did you verify the change in the behavior with reported watchroots?

LeFrosch:
> did you verify the change in the behavior with reported watchroots? 

@ujohnny no, not yet. Will do so tomorrow. Let's wait with merging until then.

PR #8090, opened 2026-01-07
Fix Scala source resolution for custom rules forwarding ScalaInfo

FrankPortman:
## Problem

The aspect uses `rule_kind.startswith("scala")` to determine whether to omit `interface_jar` for Scala targets in `library_artifact()`. This breaks source resolution (go-to-definition / Cmd+B) for custom wrapper rules that forward `ScalaInfo` and `JavaInfo` providers but don't have rule names starting with `scala_`.

For example, a custom rule like this doesn't get proper source resolution:

```python
my_scala_library = rule(
    implementation = _my_impl,
    attrs = {"scala_target": attr.label(providers = [JavaInfo])},
    provides = [JavaInfo],  # forwards JavaInfo from underlying scala_library
)
```

## Solution

Instead of checking the rule name prefix, detect Scala targets via the presence of the `ScalaInfo` provider:

1. Added `scala_info.bzl` with `scala_info_in_target()` function
2. Updated `library_artifact()` to accept `is_scala` boolean parameter
3. Updated `collect_java_info()` to use provider detection

This allows any rule that forwards `ScalaInfo` to be treated correctly, regardless of its name.

## Testing

Tested locally with a custom rule wrapping `scala_library` - source resolution now works correctly.

### DIFF
--- a/aspect/intellij_info_impl.bzl
+++ b/aspect/intellij_info_impl.bzl
@@ -747,9 +747,10 @@ def collect_java_info(target, ctx, semantics, ide_info, ide_info_file, output_gr
 
     ide_info_files = []
     sources = sources_from_target(ctx)
-    # Detect Scala targets via ScalaInfo provider rather than rule name prefix.
-    # This allows custom rules that forward ScalaInfo to be treated correctly.
-    is_scala = scala_info_in_target(target)
+    # Detect Scala targets via ScalaInfo provider, with fallback to rule name prefix
+    # for backwards compatibility. This allows custom rules that forward ScalaInfo
+    # to be treated correctly while still supporting standard scala_* rules.
+    is_scala = scala_info_in_target(target) or ctx.rule.kind.startswith("scala")
     jars = [library_artifact(output, is_scala) for output in java_outputs]
     class_jars = [output.class_jar for output in java_outputs if output and output.class_jar]
     output_jars = [jar for output in java_outputs for jar in jars_from_output(output)]
--- a/aspect/scala_info.bzl
+++ b/aspect/scala_info.bzl
@@ -7,18 +7,13 @@
 def scala_info_in_target(target):
     """Returns True if the target has ScalaInfo provider, indicating it's a Scala target."""
 # TEMPLATE-IGNORE-BEGIN
-    # Check if ScalaInfo provider exists (from rules_scala)
-    # We use a string-based check to avoid hard dependency on rules_scala
-    for provider in dir(target):
-        if "ScalaInfo" in str(provider):
-            return True
     return False
 # TEMPLATE-IGNORE-END
 
 # TEMPLATE-INCLUDE-BEGIN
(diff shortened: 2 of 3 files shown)


LeFrosch:
Looks good to me, do you object to merging this? @sellophane

aspect/intellij_info_impl.bzl
```diff
@@ -173,13 +174,19 @@ def get_source_jars(output):
...
+        java_output: The Java output to create a library artifact from.
+        is_scala: If True, omit interface_jar (Scala targets don't use ijar).
+                  This is detected via ScalaInfo provider rather than rule name.
+    """
     if java_output == None or java_output.class_jar == None:
         return None
     src_jars = get_source_jars(java_output)
 
-    if rule_kind != None and rule_kind.startswith("scala"):
+    if is_scala:
```
FrankPortman, line 189: We could have it satisfy based on both approaches, to make sure its maximally backwards compatible? I'm not 100% sure what is preferable because I don't actually understand why the ijar thing even matters for Scala source detection.

FrankPortman:
friendly bump @LeFrosch @sellophane

sellophane:
I tested this and this works alright, but I have some reservations about this pseudo-reflection done in absence of scala_rules...

FrankPortman:
> I tested this and this works alright, but I have some reservations about this pseudo-reflection done in absence of scala_rules...

@sellophane honestly I just misread how the other templates worked and thought they did the same string checking. I am happy to add a real dep on `rules_scala` if you prefer. And I can also make the check based on "starts with `scala` or `ScalaInfo` in target" if you want maximal backwards compatibility. Let me know.

FrankPortman:
Actually it already exists --> https://github.com/bazelbuild/intellij/blob/ijwb/MODULE.bazel#L16

I was just going kind of quickly because I wasn't sure if y'all were accepting any PRs, but happy to clean this up now. Just let me know what it will take to land this.

sellophane:
Falling back to string-based recognition seems optimal, however I would argue that we can go with simply returning hardcoded False if there is no Scala support

FrankPortman:
> Falling back to string-based recognition seems optimal, however I would argue that we can go with simply returning hardcoded False if there is no Scala support

Can you expand on this - do you mean when users don't set Scala as a language in `.bazelproject`?

sellophane:
> > Falling back to string-based recognition seems optimal, however I would argue that we can go with simply returning hardcoded False if there is no Scala support
> 
> Can you expand on this - do you mean when users don't set Scala as a language in `.bazelproject`?

Yes, exactly this.

FrankPortman:
@sellophane please take a look at the most recent version. It copies, as best as possible, the way the other language templating works. It uses a real `ScalaInfo` provider check. It does, however, fall back on the `scala` prefix match, for backwards compatability. I figure y'all would prefer to avoid "I am seeing this very subtle issue in the latest release" from releasing this.

PR #8093, opened 2026-01-09
add base:integration_tests to clwb_tests

ujohnny:
a long ago they were ran as a part of ijwb_common_tests that were removed later after ijwb removal

### DIFF
--- a/base/src/com/google/idea/blaze/base/lang/buildfile/formatting/BuildFileFoldingBuilder.java
+++ b/base/src/com/google/idea/blaze/base/lang/buildfile/formatting/BuildFileFoldingBuilder.java
@@ -107,8 +107,25 @@ public class BuildFileFoldingBuilder implements FoldingBuilder {
   private static void foldLongStrings(List<FoldingDescriptor> descriptors, ASTNode node) {
     boolean isMultiLine = node.textContains('\n');
     if (isMultiLine) {
-      descriptors.add(new FoldingDescriptor(node, node.getTextRange()));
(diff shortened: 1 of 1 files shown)


base/src/com/google/idea/blaze/base/lang/buildfile/formatting/BuildFileFoldingBuilder.java
```diff
@@ -107,10 +107,27 @@ private static void foldFunctionCall(List<FoldingDescriptor> descriptors, ASTNod
...
+      QuoteType type = str.getQuoteType();
+      String contents = str.getStringContents();
+      if (contents.contains("\n")) {
+        int newlineIx = contents.indexOf('\n');
+        String firstLine = newlineIx == -1 ? contents : contents.substring(0, newlineIx);
+        return type.wrap(firstLine + "...");
+      }
+      return type.wrap("...");
+    }
+    return "...";
```
LeFrosch, line 128: This seems a bit odd and I wonder why this was not required before 👀

LeFrosch (approved):

base/src/com/google/idea/blaze/base/lang/buildfile/formatting/BuildFileFoldingBuilder.java
ujohnny: Claude says the behavior got changed in 2025.2(?) but I don't trust it here and believe it's just a generated text. Also I'm unsure if this was even working as expected in production, I never saw a folded multiline block

PR #8099, opened 2026-01-12
Reenable base:integration_tests

LeFrosch:
Apparently, somewhere along the way we dropped the base:integration_tests from CI by accident. 

Follow up for: #8093

### DIFF
--- a/.managed.bazelproject
+++ b/.managed.bazelproject
@@ -9,6 +9,7 @@ directories:
   -testing/test_deps/projects
 
 targets:
+  //:clwb_tests
   //clwb:all
   //clwb:clwb_bazel_dev
   //aspect/testing/...


ujohnny (approved):

base/src/com/google/idea/blaze/base/lang/buildfile/formatting/BuildFileFoldingBuilder.java
```diff
@@ -107,10 +107,27 @@ private static void foldFunctionCall(List<FoldingDescriptor> descriptors, ASTNod
...
+      String contents = str.getStringContents();
+      if (contents.contains("\n")) {
+        int newlineIx = contents.indexOf('\n');
+        String firstLine = newlineIx == -1 ? contents : contents.substring(0, newlineIx);
+        return type.wrap(firstLine + "...");
+      }
+      return type.wrap("...");
+    }
+    return "...";
+  }
```
LeFrosch, line 129: NOTE: as discussed, we are not sure if the folding actually works. This needs to be checked.

PR #8120, opened 2026-01-24
Fix CcInfo being None when C/C++ is disabled

FrankPortman:
When $isCcEnabled is false, the load statement for CcInfo is skipped, but `CcInfoCompat = CcInfo` was outside the template conditional. This causes CcInfo to be undefined, resulting in:

```
Error in aspect: at index 0 of required_aspect_providers,
got element of type NoneType, want Provider
```

Move CcInfoCompat and cc_common_compat assignments inside the template conditional with proper fallback to None when CC is disabled.

### DIFF
--- a/aspect/cc_info.bzl
+++ b/aspect/cc_info.bzl
@@ -10,18 +10,17 @@
 load("@rules_cc//cc:defs.bzl", "CcInfo", "cc_common")
 
 CC_USE_GET_TOOL_FOR_ACTION = True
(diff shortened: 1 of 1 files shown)


LeFrosch:
Thanks for the PR, this should be handled by @odisseus

ujohnny:
> Thanks for the PR, this should be handled by @odisseus

I believe you’ve meant @sellophane

PR #8150, opened 2026-02-10
Prefer prebuild protoc

LeFrosch:
This avoids frequent rebuilds of the protoc compiler.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,3 +19,7 @@ common --repo_env=ANDROID_HOME=""
 common --repo_env=ANDROID_SDK_ROOT=""
 common --repo_env=ANDROID_NDK_HOME=""
 common --repo_env=ANDROID_NDK_ROOT=""
+
+# Avoids frequent protoc rebuilds 
+build --@protobuf//bazel/toolchains:prefer_prebuilt_protoc
+


.bazelrc
```diff
@@ -18,4 +18,8 @@ build --legacy_external_runfiles
 common --repo_env=ANDROID_HOME=""
 common --repo_env=ANDROID_SDK_ROOT=""
 common --repo_env=ANDROID_NDK_HOME=""
-common --repo_env=ANDROID_NDK_ROOT=""
\ No newline at end of file
+common --repo_env=ANDROID_NDK_ROOT=""
+
+# Avoids frequent protoc rebuilds 
+build --@protobuf//bazel/toolchains:prefer_prebuilt_protoc
+
```
agluszak, line 25: I think you also need
```suggestion
build --incompatible_enable_proto_toolchain_resolution
```

LeFrosch: Seems to be enabled by default on Bazel 9

```
$ bazel help build | grep incompatible_enable_proto_toolchain_resolution
--[no]incompatible_enable_proto_toolchain_resolution (a boolean; default: "true")
```

PR #8167, opened 2026-02-23
Transition Support: create resolve config for each configuration

LeFrosch:
Create a unique resolve configuration for each discovered configuration per target.

### DIFF
--- a/cpp/src/com/google/idea/blaze/cpp/BlazeResolveConfiguration.kt
+++ b/cpp/src/com/google/idea/blaze/cpp/BlazeResolveConfiguration.kt
@@ -84,14 +84,13 @@ data class BlazeResolveConfiguration(
 }
 
 private fun computeDisplayName(targets: Collection<TargetKey>): String {
-  LOG.assertTrue(targets.map { it.configurationId() }.distinct().size == 1)
-
   val minTargetKey = requireNotNull(targets.minOrNull())
 
   return buildString {
     append(minTargetKey.label())
 
-    if (!minTargetKey.configurationId().isBlank()) {
+    // on resolve configuration can cover multiple Bazel configurations with same compiler option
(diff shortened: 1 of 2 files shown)


base/src/com/google/idea/blaze/base/model/BlazeConfigurationData.java
```diff
@@ -55,7 +56,19 @@ public static BlazeConfigurationData fromProto(ProjectData.BlazeConfigurationDat
...
+
+    for (final var entry : configurations.entrySet()) {
+      // the short_id provided by the aspect is the prefix of the actual configuration hash
+      if (entry.getKey().startsWith(key.configurationId())) {
+        return entry.getValue();
+      }
+    }
+
+    return null;
   }
```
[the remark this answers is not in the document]
LeFrosch: since Bazel itself uses the prefixes I think it is safe to assume this is not gonna happen.

cpp/src/com/google/idea/blaze/cpp/BlazeResolveConfiguration.kt
```diff
@@ -84,14 +84,19 @@ data class BlazeResolveConfiguration(
 private fun computeDisplayName(targets: Collection<TargetKey>): String {
   val minTargetKey = requireNotNull(targets.minOrNull())
 
+  val name = String.format(
+    "%s (%s)",
+    minTargetKey.label().toString(),
+    minTargetKey.configurationId().ifBlank { "unknown" },
```
ujohnny: let's omit this if configurationId is null and we don't need to update the tests then too

LeFrosch: done

ujohnny (approved):

PR #8175, opened 2026-02-25
Execlog parser

LeFrosch:
This PR adds an action to parse Bazel's execlog files and display their content in the IDE. To avoid depending on the protobuf plugin the basic proto lexer was copied and adapted for highlighting.


common/util/src/com/google/idea/common/util/RunJarService.kt
```diff
@@ -52,6 +59,24 @@ class RunJarService {
     }
   }
 
+  @Throws(ExecutionException::class)
+  private suspend fun capture(jar: Path, vararg args: String): ProcessOutput {
+    val handler = run(jar, *args)
```
ujohnny: let's use `CapturingProcessRunner` here

PR #8176, opened 2026-02-26
Fix nested Bazel execution deadlock

LeFrosch:
The action to set up Bazel in GitHub CI provides a system or user RC file which overrides the output_user_root. This can cause the nested Bazel client to try to connect to the outer Bazel server, effectively creating a dead lock.

So apparently:
- the system rc is required in order for Build Kit to work correctly
- and the home rc needs to be ignored for GH to work correctly 🫠


LeFrosch:
GH CD works with these changes: https://github.com/LeFrosch/intellij-bazel/actions/runs/22439001473

PR #8181, opened 2026-03-02
Include Cache: generate symlinks

LeFrosch:
Currently the include cache is populated with copies of the actual files. To make it easier in future to resolve to the correct source file this PR creates symlinks inside the cache if applicable.


MODULE.bazel.out
```diff
@@ -0,0 +1,205 @@
+bazel_dep(name = "platforms", version = "1.0.0")
```
ujohnny: accidental leftover

LeFrosch: done

PR #8295, opened 2026-04-16
Fix compiler wrapper escapes for defines with spaces

LeFrosch:
-D flags containing spaces/quotes/parens are not properly escaped.

Reported on YT: https://youtrack.jetbrains.com/issue/BAZEL-3079/CLION-flags-with-spaces-are-not-properly-escaped.


cpp/src/com/google/idea/blaze/cpp/CompilerWrapperProviderImpl.java
```diff
@@ -65,13 +65,10 @@ public File createCompilerExecutableWrapper(
...
-            "      # the final line if it does not.",
-            "      echo >> ${arg#@}",
-            "      # Args file, the read will remove a layer of escaping",
-            "      while read; do",
-            "        parsedargs+=($REPLY)",
-            "      done < ${arg#@}",
+            "      # Parse response file with xargs (handles backslash-escaped spaces/quotes) into array",
+            "      while read -r -d ''; do",
+            "        parsedargs+=(\"$REPLY\")",
+            "      done < <(xargs printf '%s\\0' < \"${arg#@}\")",
```
ujohnny, line 71: if the input is `\n`-separated then setting IFS to '\n' should be enough without xargs involved here

cpp/src/com/google/idea/blaze/cpp/CompilerWrapperProviderImpl.java
```diff
@@ -65,13 +65,10 @@ public File createCompilerExecutableWrapper(
...
             "  case \"$arg\" in",
             "    @*)",
-            "      # Make sure the file ends with a newline - the read loop will not return",
-            "      # the final line if it does not.",
-            "      echo >> ${arg#@}",
-            "      # Args file, the read will remove a layer of escaping",
-            "      while read; do",
-            "        parsedargs+=($REPLY)",
-            "      done < ${arg#@}",
+            "      # Parse response file with xargs (handles backslash-escaped spaces/quotes) into array",
```
ujohnny, line 68: actually CLion is capable of detecting issues of response files processing `com.jetbrains.cidr.lang.workspace.compiler.GCCCompiler#tryRunGCC` and it makes a second attempt to run compiler without response file.

ujohnny (approved):

cpp/src/com/google/idea/blaze/cpp/CompilerWrapperProviderImpl.java
ujohnny: though CLion expects a very particular error message, which might not be reported by bazel wrapper

cpp/src/com/google/idea/blaze/cpp/CompilerWrapperProviderImpl.java
LeFrosch: compiler wrapper was removed

cpp/src/com/google/idea/blaze/cpp/CompilerWrapperProviderImpl.java
LeFrosch: I created a new OCCompilerKind for Bazel which provide a modified GCCCompiler instance that does not even try to pass response files.

cpp/src/com/google/idea/blaze/cpp/BazelCompilerKind.kt
```diff
@@ -0,0 +1,74 @@
...
+
+  override fun getCommandLineShortener(): OCCompilerCommandLineShortener = BasicCompilerCommandLineShortener()
+
+  override fun getCompilerInstance(
+    project: Project,
+    compilerExecutable: File,
+    compilerWorkingDirectory: File,
+    environment: CidrToolEnvironment,
+    tempFilesPool: TempFilesPool,
+  ): OCCompiler = object : GCCCompiler(compilerExecutable, compilerWorkingDirectory, environment, tempFilesPool) {
```
ujohnny: let's extract this to a named class

cpp/src/com/google/idea/blaze/cpp/BazelCompilerKind.kt
```diff
@@ -0,0 +1,74 @@
...
+ * An [OCCompilerKind] wrapper that disables response file usage.
+ *
+ * Bazel's compiler wrapper scripts do not support response files (`@file` arguments).
+ * This kind delegates all behavior to the underlying [delegate] kind, but overrides
+ * [getCompilerInstance] to return a [BazelGCCCompiler] (which disables response file
+ * shortening) and [getCommandLineShortener] to return a no-op shortener.
+ *
+ * Yes, there are two places to define the command line shortener :(
+ */
+open class BazelCompilerKind(val delegate: OCCompilerKind) : OCCompilerKind by delegate {
```
ujohnny, line 41: warning about non delegated methods is suspicious as they are default

cpp/src/com/google/idea/blaze/cpp/BazelCompilerKind.kt
```diff
@@ -0,0 +1,74 @@
...
+  override fun equals(other: Any?): Boolean =
+    delegate == other || (other is BazelCompilerKind && delegate == other.delegate)
+
+  override fun hashCode(): Int = delegate.hashCode()
+
+  override fun toString(): String = "Bazel($delegate)"
+}
+
+/** Bazel-specific GCC compiler kind that disables response files. */
+object BazelGCCCompilerKind : BazelCompilerKind(GCCCompilerKind)
```
ujohnny, line 79: not sure about readResolve warning here, nothing in CLion code does that
maybe we can omit it

ujohnny (approved):

cpp/src/com/google/idea/blaze/cpp/BazelCompilerKind.kt
LeFrosch: done

cpp/src/com/google/idea/blaze/cpp/BazelCompilerKind.kt
LeFrosch: I think they do need an override since we cannot be sure whether the delegate overrides them or not.

cpp/src/com/google/idea/blaze/cpp/BazelCompilerKind.kt
LeFrosch: this warning seems bogus.

PR #8297, opened 2026-04-16
Transition Support: convert BlazeCidrRunConfigurationRunner to kotlin

LeFrosch:
Required for later changes that enable automatic configuration selection from run configs.


clwb/src/com/google/idea/blaze/clwb/run/BlazeCidrRunConfigurationRunner.kt
```diff
@@ -0,0 +1,117 @@
...
+      executableToDebug = getExecutableToDebug(env).toFile()
+      return true
+    } catch (e: ExecutionException) {
+      ExecutionUtil.handleExecutionError(env.project, env.executor.toolWindowId, env.runProfile, e)
+    }
+    return false
+  }
+
+  /**
+   * Builds blaze C/C++ target in debug mode, and returns the output build artifact.
```
ujohnny, line 66: though this is not 100% correct

ujohnny (approved):

PR #8317, opened 2026-04-23
Execution: detect missing debug flags rather then injecting them

LeFrosch:
Previously the plugin always injected a fixed set of debug build flags (`--compilation_mode=dbg`, `--copt=-g2`, `--strip=never`, etc.) whenever a CC run configuration was launched in debug mode. This could silently invalidate the analysis cache and cause high rebuild times when the user switched between "Run" and "Debug" even when their project already produced a debuggable binary.                                                                                                                                                                                                                                                                                     
   
This PR flips the default: we no longer inject anything. Instead, `DebugInfoCheck` runs an `aquery` over the target's dependencies before the debug session starts and warns when the resolved compile actions lack debug info.

For anyone who needs the old behavior, `bazel.clwb.debug.extraflags.legacy` (default off) restores the injection uniformly across both the bundled-debugger and gdbserver paths.


base/src/com/google/idea/blaze/base/buildview/RunConfigBuild.kt
```diff
@@ -0,0 +1,97 @@
...
+      ?.blazeFlagsState
+      ?.flagsForExternalProcesses
+      ?: emptyList()
+
+    val cmd = BlazeCommand.builder(BlazeCommandName.BUILD)
+      .addTargets(target)
+      .addBlazeFlags(additionalFlags)
+      .addBlazeFlags(flags)
+      .addBlazeFlags(externalFlags)
+      .addBlazeFlags()
```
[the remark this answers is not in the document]
LeFrosch: done

clwb/src/com/google/idea/blaze/clwb/run/DebugInfoCheck.kt
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Finds the [OCCompilerKind] for a given Bazel configuration by matching the
+ * configuration checksum against OCResolveConfigurations in the workspace.
+ */
+private fun findCompilerKind(project: Project, bazelConfigHash: String): OCCompilerKind? {
+  for (config in OCWorkspace.getInstance(project).configurations) {
+    val id = BlazeResolveConfigurationID.fromOCResolveConfiguration(config) ?: continue
+
+    if (bazelConfigHash.startsWith(id.configurationId)) {
+      return config.getCompilerSettings(CLanguageKind.CPP).compilerKind
```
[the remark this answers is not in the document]
LeFrosch: wrong

common/aquery/ActionGraph.kt
```diff
@@ -0,0 +1,101 @@
+package com.google.idea.common.aquery;
```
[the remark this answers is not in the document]
LeFrosch: done

common/aquery/ActionGraph.kt
```diff
@@ -0,0 +1,101 @@
...
+      targetMap = input.targetsList.associateBy { it.id },
+      allActions = input.actionsList,
+    )
+  }
+
+  val targets: Sequence<Target> get() = targetMap.values.asSequence().map { Target(it) }
+
+  val defaultTarget: Target get() = Target(targetMap.values.minBy { it.id })
+
+  val defaultConfiguration: Configuration get() = Configuration(configurationMap.values.minBy { it.id })
```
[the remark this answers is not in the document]
LeFrosch: there should always at least be one target, tho it is possible to build an empty set of targets 🙃

clwb/src/com/google/idea/blaze/clwb/run/DiscoverTargetConfigurations.kt
```diff
@@ -0,0 +1,88 @@
...
+      .addBlazeFlags(additionalFlags)
+      .addBlazeFlags(flags)
+      .addBlazeFlags(externalFlags)
+      .addBlazeFlags("--output=streamed_proto")
+
+    BazelExecService.of(project).exec(ctx, cmd).use { result ->
+      result.throwOnFailure()
+
+      val graph = try {
+        ActionGraph.fromProto(result.stdout)
```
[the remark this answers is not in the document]
LeFrosch: I think this is wrong as well

clwb/src/com/google/idea/blaze/clwb/run/DebugInfoDismissalState.kt
```diff
@@ -0,0 +1,57 @@
...
+@State(name = "DebugInfoDismissalState")
+class DebugInfoDismissalState : PersistentStateComponent<DebugInfoDismissalState> {
+
+  var dismissedForProject: Boolean = false
+  var dismissedTargets: MutableList<String> = mutableListOf()
+
+  override fun getState(): DebugInfoDismissalState = this
+
+  override fun loadState(state: DebugInfoDismissalState) {
+    XmlSerializerUtil.copyBean(state, this)
```
ujohnny, line 35: ideally we should have kotlinx.serialization compiler plugin attached to this target, `DebugInfoDismissalState` ( or inner State class marked as `@Serializable` and have nothing like this

clwb/src/com/google/idea/blaze/clwb/run/DebugInfoCheck.kt
```diff
@@ -0,0 +1,134 @@
...
+import com.jetbrains.cidr.lang.workspace.compiler.ClangClCompilerKind
+import com.jetbrains.cidr.lang.workspace.compiler.ClangCompilerKind
+import com.jetbrains.cidr.lang.workspace.compiler.GCCCompilerKind
+import com.jetbrains.cidr.lang.workspace.compiler.MSVCCompilerKind
+import com.jetbrains.cidr.lang.workspace.compiler.OCCompilerKind
+
+private const val NOTIFICATION_TITLE = "Debug Info Warning"
+
+private val NOTIFICATION_MESSAGE = """
+  CLion detected that your target might not be debuggable.
```
ujohnny: maybe 

```
...
Read more in our GitHub docs <link> 
...
```

clwb/src/com/google/idea/blaze/clwb/run/DebugInfoCheck.kt
```diff
@@ -0,0 +1,134 @@
...
+
+  override val title: String = "Check CC Debug Info"
+
+  override fun run(ctx: BlazeContext) {
+    val target = configurations.mainTarget
+
+    val compilerKind = findCompilerKind(project, configurations.mainConfiguration.checksum)
+
+    val nonDebuggableTargets = configurations.compileActions.asSequence()
+      .filter { (_, action) -> !checkDebugInfoPresent(action.arguments, compilerKind) }
```
ujohnny: if target foo is just a combination of n libraries and 0 sources, will it have these flags set?

LeFrosch: interesting, don't know, lemme check

LeFrosch: a cc_library with no sources does not produce any artifacts, i.e. does not run any CppCompile action.

ujohnny:
We're probably safe to merge this once https://github.com/bazelbuild/rules_cc/pull/692 is available and just require recent version of rules_cc

LeFrosch:
Added a project view option (`inject_debug_flags`) that can be set to revert to the previous behavior. All registry keys have been dropped. The options is disabled by default. However, we don't have a link-options check yet for macos.

This is the current version of the popup:
image

clwb/src/com/google/idea/blaze/clwb/run/DebugInfoWarningDialog.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.project.Project
+import com.intellij.openapi.ui.DialogWrapper
+import com.intellij.ui.dsl.builder.panel
+import javax.swing.Action
+import javax.swing.JComponent
+import javax.swing.SwingConstants
+
+private const val NOTIFICATION_TITLE = "Debug Info Warning"
+private const val DOC_URL = "https://github.com/bazelbuild/intellij/blob/master/docs/cpp/debugging.md"
```
ujohnny: make it jb.gg

docs/cpp/debugging.md
```diff
@@ -0,0 +1,92 @@
...
+> the next build, since the build configuration changes.
+
+## I had working debug before — what should I do?
+
+1. Click **inject debug flags** in the dialog. The plugin writes
+   `inject_debug_flags: true` to your local `.bazelproject` and
+   re-runs the configuration with the injected flags. This is similar
+   to the plugin's previous behavior.
+2. Verify command-line debug still works:
+   `bazel run -c dbg //your:target` and attach a debugger, or build
```
ujohnny, line 78: >attach a debugger

not related to this review, but why do we even do run under gdbserver? why can't we just `gdb -p ...`

ujohnny (approved):
Let's adjust the link and LGTM

clwb/src/com/google/idea/blaze/clwb/run/DebugInfoWarningDialog.kt
LeFrosch: done

PR #8326, opened 2026-05-08
Workaround for breaking API change in 261.1

LeFrosch:
The signature of the `RadSymbolsHost.isEntryPointOffset` changed with 261.1. Since this is a breaking change within one major version we cannot use the sdkcompat layer here.

### DIFF
--- a/clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
+++ b/clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
@@ -23,6 +23,7 @@ import com.google.idea.blaze.base.run.producers.BinaryContextProvider.BinaryRunC
 import com.google.idea.blaze.cpp.CppBlazeRules.RuleTypes
 import com.intellij.clion.radler.core.protocol.RadSymbolsHost
 import com.intellij.execution.actions.ConfigurationContext
+import com.intellij.openapi.project.Project
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.psi.PsiElement
 import com.intellij.psi.PsiFile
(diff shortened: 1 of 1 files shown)


zhrkvl:
This looks better than trying to make a wrapper for the new API. As `PsiFile` is already provided, using virtual file and then searching for PsiFile again would be inefficient 
```
  fun isInEntryPoint(file: VirtualFile, offset: Int): Boolean {
    val psiFile = PsiManager.getInstance(project).findFile(file)
    return /* new API */  
  }
  
  fun isEntryPointOffset(file: VirtualFile, offset: Int): Boolean {
    val psiFile = PsiManager.getInstance(project).findFile(file)
    return  /* new API with */  
   }
```

clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
```diff
@@ -44,11 +47,26 @@ class RadBinaryContextProvider : BinaryContextProvider {
   }
 }
 
+// #api261: workaround for breaking API change inside major version
+private val isEntryPointOffsetMethod: Method by lazy {
```
ujohnny: technically this might live under api261 sdkcompat because this applies to 261 only, but maybe it would be just unnecessary complication.

clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
```diff
@@ -44,11 +47,26 @@ class RadBinaryContextProvider : BinaryContextProvider {
...
+    RadSymbolsHost::class.java.getDeclaredMethod("isEntryPointOffset", VirtualFile::class.java, Int::class.javaPrimitiveType)
+  } catch (_: NoSuchMethodException) {
+    RadSymbolsHost::class.java.getDeclaredMethod("isEntryPointOffset", PsiFile::class.java, Int::class.javaPrimitiveType)
+  }
+}
+
 private fun isMain(element: PsiElement?): Boolean {
   if (element !is LeafElement) return false
 
+  val file = if (isEntryPointOffsetMethod.parameterTypes[0] == VirtualFile::class.java) {
```
ujohnny: maybe indeed there's no need to call `parameterTypes[0]` here and we can persist this information while making `lazy` computation

LeFrosch: done

clwb/src/com/google/idea/blaze/clwb/radler/RadBinaryContextProvider.kt
LeFrosch: agree, we can do this with the bump to 262.

ujohnny (approved):
overall ok for such an issue

PR #8342, opened 2026-05-22
Remove generic Bazel flags field from run configs

LeFrosch:
The free-form flags field is removed because per-config Bazel flags participate in Bazel's analysis cache key and cause sync/cache-invalidation issues; a dedicated single-line "Bazel test filter" field replaces its only remaining legitimate use case. Existing run-config XML is migrated on load: `--test_filter=...` flows into the new field, `--test_arg=...` flows into the executable-flags field, and any remaining flags are surfaced to the user via a one-time modal dialog that offers to copy them into the project view's `build_flags:` section.

### DIFF
--- a/base/src/com/google/idea/blaze/base/run/state/LegacyBlazeFlagsMigrationNotifier.kt
+++ b/base/src/com/google/idea/blaze/base/run/state/LegacyBlazeFlagsMigrationNotifier.kt
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2026 The Bazel Authors. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.idea.blaze.base.run.state
+
+import com.google.idea.blaze.base.projectview.ProjectViewEdit
+import com.google.idea.blaze.base.projectview.section.ListSection
+import com.google.idea.blaze.base.projectview.section.sections.BuildFlagsSection
+import com.google.idea.blaze.base.run.BlazeCommandRunConfiguration
+import com.google.idea.blaze.base.settings.ui.OpenProjectViewAction
+import com.google.idea.blaze.base.sync.BlazeSyncManager
+import com.intellij.execution.RunManager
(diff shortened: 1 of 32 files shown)


base/src/com/google/idea/blaze/base/run/state/LegacyBlazeFlagsMigrationNotifier.kt
```diff
@@ -0,0 +1,119 @@
...
+
+private const val SHOWN_KEY = "legacy_blaze_flags_migration_shown"
+private const val SYNC_REASON = "LegacyBlazeFlagsMigrationNotifier"
+
+/**
+ * Once per project, scans for run configurations carrying legacy `<blaze-user-flag>` entries that
+ * no longer have a typed home and offers to move them into the project view's `build_flags:`
+ * section via a modal dialog.
+ */
+class LegacyBlazeFlagsMigrationNotifier : StartupActivity.DumbAware {
```
ujohnny: use `ProjectActivity`

base/src/com/google/idea/blaze/base/run/state/LegacyBlazeFlagsMigrationNotifier.kt
```diff
@@ -0,0 +1,119 @@
...
+
+  override fun runActivity(project: Project) {
+    if (ApplicationManager.getApplication().isUnitTestMode) return
+    if (isLegacyMigrationShown(project)) return
+
+    val affected = collectLegacyConfigurations(project)
+    if (affected.isEmpty()) return
+
+    ApplicationManager.getApplication().invokeLater {
```
ujohnny: and then this will be suspend

ujohnny (changes requested):

base/src/com/google/idea/blaze/base/run/state/LegacyBlazeFlagsMigrationNotifier.kt
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Once per project, scans for run configurations carrying legacy `<blaze-user-flag>` entries that
+ * no longer have a typed home and offers to move them into the project view's `build_flags:`
+ * section via a modal dialog.
+ */
+class LegacyBlazeFlagsMigrationNotifier : ProjectActivity {
+
+  override suspend fun execute(project: Project) {
+    if (ApplicationManager.getApplication().isUnitTestMode) return
```
ujohnny: let's not modify production flow in tests

base/src/com/google/idea/blaze/base/run/state/LegacyBlazeFlagsMigrationNotifier.kt
```diff
@@ -0,0 +1,122 @@
...
+    if (isLegacyMigrationShown(project)) return
```
ujohnny: we actually have `com.intellij.ide.util.RunOnceUtilKt#runOnceForProject` for such a purpose

ujohnny (approved):

PR #8348, opened 2026-06-02
chore(SDK): add 262 support

LeFrosch:


.bazelrc
```diff
@@ -1,5 +1,5 @@
-build --java_language_version=21 --java_runtime_version=21
-build --tool_java_language_version=21 --tool_java_runtime_version=21
+build --java_language_version=21 --java_runtime_version=25
```
ujohnny: though this is not 100% correct because 253 (which is still supported) requires 21 and 25 is not the same as 21

ujohnny (approved):

intellij_platform_sdk/BUILD.clion261
```diff
@@ -20,6 +20,26 @@ load("@rules_java//java:defs.bzl", "java_import")
...
 package(default_visibility = ["//visibility:public"])
 
+toolchain(
+    name = "runtime",
+    toolchain_type = "@bazel_tools//tools/jdk:runtime_toolchain_type",
+    target_settings = [
+        "@//intellij_platform_sdk:clion-2026.1-or-alias",
+        "@//intellij_platform_sdk:jbr_runtime",
+    ],
+    toolchain = "@jbr_21_0_11//:runtime",
```
ujohnny: isn't this wrong?

LeFrosch: updated to jbr 25.0.3

PR #8419, opened 2026-07-24
Provide a deterministic fallback for RadRunConfigurationChooser

LeFrosch:
When no run configuration was executed and auto is selected in the drop no `OCResolveContext` was selected. A deterministic in `RadRunConfigurationChooser` guarantees that if a Bazel specific resolve configuration is a available it will be selected.

### DIFF
--- a/clwb/src/com/google/idea/blaze/clwb/radler/RadRunConfigurationChooser.kt
+++ b/clwb/src/com/google/idea/blaze/clwb/radler/RadRunConfigurationChooser.kt
@@ -36,6 +36,11 @@ class RadRunConfigurationChooser : OCResolveConfigurationChooserAdapter() {
   override fun doSelectResolveConfiguration(
     project: Project,
     configurations: List<OCResolveConfiguration>,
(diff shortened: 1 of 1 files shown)


clwb/src/com/google/idea/blaze/clwb/radler/RadRunConfigurationChooser.kt
```diff
@@ -36,6 +36,11 @@ class RadRunConfigurationChooser : OCResolveConfigurationChooserAdapter() {
   override fun doSelectResolveConfiguration(
     project: Project,
     configurations: List<OCResolveConfiguration>,
+  ): OCResolveConfiguration? = byLastBuild(project, configurations) ?: byConfigurationID(configurations)
```
ujohnny: let's better use function with body

ujohnny (changes requested):

clwb/src/com/google/idea/blaze/clwb/radler/RadRunConfigurationChooser.kt
LeFrosch: done

ujohnny (approved):

PR #8426, opened 2026-07-30
Add support for Apple clang

LeFrosch:
Adds apple clang to our compiler kind detection heuristic.


ujohnny (approved):

cpp/src/com/google/idea/blaze/cpp/copts/CoptsIncludeProcessor.kt
```diff
@@ -55,7 +55,7 @@ abstract class CoptsIncludeProcessor : CoptsProcessor.Transform() {
 
     override fun flags(kind: OCCompilerKind?): ImmutableList<String> {
       return when (kind) {
-        GCCCompilerKind, ClangCompilerKind -> ImmutableList.of("-isystem")
+        GCCCompilerKind, ClangCompilerKind, AppleClangCompilerKind -> ImmutableList.of("-isystem")
```
ujohnny, line 58: the Kind API is so wrong.

PR #8442, opened 2026-08-12
Add Bazel binary verification

LeFrosch:
A sync fails with a not so clear "Could not run Bazel info" error when the specified binary cannot be executed. This PR adds a check that tries to verify if the binary can be executed or not. The check can be disabled to limit the impact of false positives.


base/src/com/google/idea/blaze/base/sync/ProjectStateSyncTask.java
```diff
@@ -200,23 +239,73 @@ private SyncProjectState getProjectState(BlazeContext context, BlazeSyncParams p
...
-                 .runBlazeInfo(
-                     project,
-                     context,
-                     importSettings.getBuildSystem(),
-                     syncFlags);
+          .runBlazeInfo(
+              project,
+              context,
+              importSettings.getBuildSystem(),
+              syncFlags);
```
LeFrosch, line 303: Sorry, accidentally formatted the file 🙃

ujohnny (approved):

PR #8473, opened 2026-09-21
Enable the IntelliJ aspect by default

LeFrosch:
To reduce the maintenance overhead, switch to the new IntelliJ aspect by default. If this is successful we can remove the old one and stop maintaining it.


ujohnny:
Please mention registry key in the changelog

ujohnny (approved):

